PR #2, opened 2016-09-06
Devicegw uses libc instead of cli

DunderRoffe:


breakreturn:
Looks OK to me.

libsoftwarecontainer/src/gateway/devicenodegateway.cpp
```diff
@@ -50,22 +52,25 @@ bool DeviceNodeGateway::activateGateway()
...
             }
         } else {
             // No major & minor numbers specified => simply map the device from the host into the container
             getContainer()->mountDevice(dev.name);
 
-            // TODO : check if it is fine to authorize write access systematically
-            std::string cmd = StringBuilder() << "chmod o+rwx " << dev.name;
-            getContainer()->executeInContainer(cmd);
+            if (chmod(dev.name.c_str(), (S_IRUSR + S_IWUSR + S_IXUSR) | mode) != 0) {
+                log_error() << "Could not 'chmod o+rwx' mounted dev" << dev.name;
```
frznlogic: This is not equivalent to o+rwx, this is equivalent to u+rwx. Which should it be?

PR #3, opened 2016-09-07
Further improvements to config check devicegw

DunderRoffe:


libsoftwarecontainer/src/gateway/devicenodegateway.cpp
```diff
@@ -38,18 +38,31 @@ ReturnCode DeviceNodeGateway::readConfigElement(const json_t *element)
...
+
+        if (!minorSpecified) {
+            log_error() << "Minor version not specified but major version is. " << notAllowed;
+            return ReturnCode::Failure;
+        }
+
+        if (!modeSpecified) {
+            log_error() << "Mode has to be specified when minor and major version is.";
+            return ReturnCode::Failure;
+        }
```
frznlogic, line 53: This is very chatty code, might be rewritten to simplify it? Possibly just a 

const bool xxxSpecified = checkSpecified(dev.xxx, "xxx");

?

libsoftwarecontainer/src/gateway/devicenodegateway.cpp
```diff
@@ -62,11 +75,11 @@ bool DeviceNodeGateway::activateGateway()
 {
     for (auto &dev : m_devList) {
         log_info() << "Mapping device " << dev.name;
-        const int mode = std::atoi(dev.mode.c_str());
 
         if (dev.major.length() != 0) {
             const int majorVersion = std::atoi(dev.major.c_str());
             const int minorVersion = std::atoi(dev.minor.c_str());
+            const int mode = std::atoi(dev.mode.c_str());
```
tobsan: Do we really want to use atoi or something more failsafe?

frznlogic: std::stoi might be better i guess? Is that enough?

DunderRoffe: Solved this by using parseInt that is already used in the agent so moved it to comon.

PR #5, opened 2016-09-07
CreateContainer config added

frznlogic:
This configuration can be used to setup certain configuration that needs
to be done as early as possible.

### DIFF
--- a/agent/src/agent.cpp
+++ b/agent/src/agent.cpp
@@ -37,14 +37,17 @@
 #include "libsoftwarecontainer.h"
 #include "softwarecontainer-common.h"
 
+#include <jsonparser.h>
+
 LOG_DEFINE_APP_IDS("SCAG", "SoftwareContainer agent");
 
 namespace softwarecontainer {
 
 LOG_DECLARE_DEFAULT_CONTEXT(PAM_DefaultLogContext, "MAIN", "Main context");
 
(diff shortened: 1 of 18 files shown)


DunderRoffe:
Looks ok to me

agent/src/agent.cpp
```diff
@@ -103,13 +106,75 @@ class SoftwareContainerAgent
         return valid;
     }
 
+    ReturnCode readConfigElement(const json_t *element)
+    {
+        std::string wo;
```
breakreturn, line 111: I personally think short variable names like this are more difficult to read. In this case it's not a problem as the code is short, but I don't see any benefit either. I mean, I want it to read e.g. "does write often equal one", not "does veee-ooh equal one". A matter of taste, just my two cents.

agent/src/agent.cpp
```diff
@@ -103,13 +106,75 @@ class SoftwareContainerAgent
...
+        return ReturnCode::SUCCESS;
+    }
+
+    /**
+     * Parse config needed for starting up the container in a correct manner.
+     */
+    bool parseConfig(const std::string &config)
+    {
+        if (config.size() == 0)
+        {
```
breakreturn, line 133: The bracket should be on the line above for consistency.

PR #6, opened 2016-09-08
Breaking out class method implementations

frznlogic:
A lot of class method implementations have been declared inline in the header files.
This commit breaks them out into separate cpp files.
Some classes have also been broken into different .h files.


agent/src/main.cpp
```diff
@@ -0,0 +1,119 @@
...
+    int preloadCount = 0;
+    int userID = 0;
+    bool shutdownContainers = true;
+    int timeout = 2;
+
+    int option_index = 0;
+    int c = 0;
+    while((c = getopt_long(argc, argv, "p:u:s:t:",long_options, &option_index)) != -1) {
+        switch(c)
+        {
```
breakreturn: Bracket should be on the line above.

agent/src/softwarecontaineragent.cpp
```diff
@@ -0,0 +1,231 @@
...
+    }
+
+    return valid;
+}
+
+ReturnCode SoftwareContainerAgent::readConfigElement(const json_t *element)
+{
+    std::string wo;
+    if(!read(element, "writeOften", wo))
+    {
```
breakreturn: Bracket should be on the line above.

agent/src/softwarecontaineragent.cpp
```diff
@@ -0,0 +1,231 @@
...
+            m_softwarecontainerWorkspace.m_writeOften = false;
+        }
+    }
+    return ReturnCode::SUCCESS;
+}
+
+bool SoftwareContainerAgent::parseConfig(const std::__cxx11::string &config)
+{
+    if (config.size() == 0)
+    {
```
breakreturn: bracket

agent/src/softwarecontaineragentadaptor.h
```diff
@@ -0,0 +1,56 @@
...
+
+    SoftwareContainerAgent &m_agent;
+
+};
+}
+
+
+// Utility class for DBus Adaptors
+class DBusCppAdaptor: public SoftwareContainerAgentAdaptor, public DBus::IntrospectableAdaptor, public DBus::ObjectAdaptor {
+    public:
```
breakreturn: It's a bit inconsistent w.r.t. indentation, compare how it's done in the code above with "public:" for example.

agent/src/main.cpp
```diff
@@ -0,0 +1,119 @@
...
+
+int signalHandler(void *data) {
+    log_debug() << "Caught signal, exiting!";
+    Glib::RefPtr<Glib::MainLoop> *ml = static_cast<Glib::RefPtr<Glib::MainLoop> *>(data);
+    (*ml)->quit();
+    return 0;
+}
+
+
+int main(int argc, char * *argv)
```
breakreturn: Space between \* and *? Funky. Is this intentional?

agent/src/softwarecontaineragent.cpp
```diff
@@ -0,0 +1,231 @@
...
+ReturnCode SoftwareContainerAgent::readConfigElement(const json_t *element)
+{
+    std::string wo;
+    if(!read(element, "writeOften", wo))
+    {
+        log_debug() << "writeOften not found";
+        m_softwarecontainerWorkspace.m_writeOften = false;
+    } else {
+        if (wo == "1")
+        {
```
breakreturn: bracket

breakreturn:
I did a brief style review, added comments where I saw something. Also, in general the vertical whitespace is not very consistent. I'm, nitpicking here of course.

agent/src/agent.cpp
```diff
@@ -1,508 +0,0 @@
...
-        }
-
-        return valid;
-    }
-
-    ReturnCode readConfigElement(const json_t *element)
-    {
-        std::string wo;
-        if(!read(element, "writeOften", wo))
-        {
```
tobsan: Pull this bracket up, please.

agent/src/agent.cpp
```diff
@@ -1,508 +0,0 @@
...
-            log_debug() << "writeOften not found";
-            m_softwarecontainerWorkspace.m_writeOften = false;
-        } else {
-            if (wo == "1")
-            {
```
tobsan: Pull this bracket up, also.

agent/src/main.cpp
tobsan: Funky indeed, I would prefer it to be **argv

agent/src/softwarecontaineragent.cpp
```diff
@@ -0,0 +1,231 @@
...
+            m_softwarecontainerWorkspace.m_writeOften = true;
+        } else {
+            m_softwarecontainerWorkspace.m_writeOften = false;
+        }
+    }
+    return ReturnCode::SUCCESS;
+}
+
+bool SoftwareContainerAgent::parseConfig(const std::__cxx11::string &config)
```
tobsan: Is there a reason we're not using std::string here?

agent/src/softwarecontaineragent.cpp
```diff
@@ -0,0 +1,231 @@
...
+        }
+    } else {
+        log_error() << "Root JSON element is not an array";
+        return false;
+    }
+
+    return true;
+}
+
+ContainerID SoftwareContainerAgent::createContainer(const std::__cxx11::string &prefix, const std::__cxx11::string &config)
```
tobsan: And why not std::string here?

tobsan:
Looks mostly good except the comments already there. Why are we changing to std::__cxx11::string from std::string everywhere?

agent/src/main.cpp
frznlogic: Not intentional. I don't think I did this? Hmmm... :)

agent/src/softwarecontaineragentadaptor.h
frznlogic: I would never indent like this, i think it's qtcreator who did the indentation when i reformatted the file. I'll have a look at my settings when I get time.

agent/src/softwarecontaineragent.cpp
frznlogic: NOOOOooooo..... I hate brackets on the same line :'( . I'll change it to be consistent if I must however.

agent/src/softwarecontaineragent.cpp
frznlogic: copy paste from other places. I don't think so, I'll have a look next week.

frznlogic: Actually, just realized that it might be qtcreator that did some magic...

agent/src/softwarecontaineragent.h
```diff
@@ -0,0 +1,115 @@
...
+ * WHETHER IN AN ACTION OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION,
+ * ARISING OUT OF OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS
+ * SOFTWARE.
+ *
+ * For further information see LICENSE
+ */
+
+
+/**
+ * That file contains an executable which register a DBUS service which lets client applications manipulate LXC containers.
```
thenor: "This"

agent/src/softwarecontaineragent.cpp
thenor: If brackets should not be on the line after then also fix them from line 132 and forward.

agent/src/softwarecontaineragentadaptor.cpp
```diff
@@ -0,0 +1,58 @@
+#include "softwarecontaineragentadaptor.h"
+
```
thenor, line 2: In the spirit of keeping a consistent code style, move the brackets to the function line.

agent/src/softwarecontaineragentadaptor.h
```diff
@@ -0,0 +1,56 @@
+#ifndef SOFTWARECONTAINERAGENTADAPTOR_H
+#define SOFTWARECONTAINERAGENTADAPTOR_H
+
+#include "softwarecontaineragent.h"
+
+namespace softwarecontainer {
+
+class SoftwareContainerAgentAdaptor :
+    public com::pelagicore::SoftwareContainerAgent_adaptor
```
thenor, line 9: One of the places where "pelagicore" is still used. Should it be changed here, or should I do it in my cleanup work?

thenor, line 9: Or is this a special case where it is still needed?

socec:
FYI,  QtCreator likes to change from std::string to std::__cxx11::string when you use the Refactor command :(

agent/src/softwarecontaineragentadaptor.cpp
frznlogic: No, but in other places. Function lines are a special case. according to our coding guideline.

agent/src/softwarecontaineragentadaptor.h
frznlogic: It's not needed but is not part of this story. I'd suggest you do it in your cleanup story ;)

breakreturn: That namespace is dictated by the generated D-Bus code I think, so we'd need to change the API (the xml etc.) in order to change this.

PR #7, opened 2016-09-08
Tests for DeviceNodeGateway

DunderRoffe:


libsoftwarecontainer/src/gateway/devicenodegateway.cpp
```diff
@@ -86,18 +91,32 @@ bool DeviceNodeGateway::activateGateway()
             parseInt(dev.mode.c_str(), &mode);
 
             // mknod dev.name c dev.major dev.minor
-            if (mknod(dev.name.c_str(), S_IFCHR | mode, makedev(majorVersion, minorVersion)) != 0) {
+            pid_t pid = INVALID_PID;
+            getContainer()->executeInContainer([&] () {
+                  return mknod(dev.name.c_str(), S_IFCHR | mode,
+                               makedev(majorVersion, minorVersion));
+                  }, &pid);
```
breakreturn: To me the indentation of the bracket here looks misaligned. But it looks consistent with the rest so perhaps this is how we want it?

libsoftwarecontainer/src/gateway/devicenodegateway.h
```diff
@@ -60,7 +60,7 @@
  *                        "mode":  "666"
  *                    }
  *                ]
- *  }
+ *
```
tobsan: Indentation is weird here, please fix.

breakreturn: jsonlint.com - paste the string and use the format the page turns it into, would make all y'all consistent

libsoftwarecontainer/src/gateway/devicenodegateway.cpp
DunderRoffe: No you are right I will change this

libsoftwarecontainer/src/gateway/devicenodegateway.h
DunderRoffe: I fix

tobsan:
Ok, looks good to me.

PR #11, opened 2016-09-13
Add more unit tests for createContainer

frznlogic:


agent/unit-test/softwarecontaineragent_unittest.cpp
```diff
@@ -24,28 +24,78 @@
...
-    ASSERT_EQ(true,true);
+TEST_F(SoftwareContainerAgentTest, CreateContainerWithNoConf) {
+    ContainerID id = sca->createContainer("iejr-", "");
+    // This is actually only true if no other containers have been created
+    // before this one. Might need to be fixed somehow.
+    ASSERT_TRUE(id == 0);
+}
+
+TEST_F(SoftwareContainerAgentTest, CreateContainerWithConf) {
+    ContainerID id = sca->createContainer("iejr-", "[{\"writeOften\": \"1\"}]");
```
tobsan: Should it really be "writeOften": "1" instead of "writeOften": true? JSON supports true/false, so we should use them imho.

tobsan:
Apart from the 1/true-issue, this looks good.

jacky405:
And I doubt the word "often" is a good choice.  What does it mean?

agent/unit-test/softwarecontaineragent_unittest.cpp
```diff
@@ -24,28 +24,78 @@
...
     bool m_shutdownContainers = true;
     int m_shutdownTimeout = 2;
+    SoftwareContainerWorkspace workspace;
 
     void SetUp() override
     {
         sca = new SoftwareContainerAgent(m_context
                                          , m_preloadCount
                                          , m_shutdownContainers
                                          , m_shutdownTimeout);
```
breakreturn: I know we haven't settled completely on the style, but the indentation here doesn't look like the one in the guide. Perhaps we can keep this until we've settled on a consistent guide...

frznlogic:
@jacky405 it might be unfortunate but it's the best description of what it actually is for now. It will be documented.

PR #13, opened 2016-09-13
Added Getting started chapter

breakreturn:
Updated user docs with a getting started chapter.
Also added comment to containerabstractinterface

### DIFF
--- a/doc/chapters/getting-started/index.rst
+++ b/doc/chapters/getting-started/index.rst
@@ -1,0 +1,133 @@
+
+Getting started
+***************
+
+Structure
+=========
+
+The user API is implemented by the SoftwareContainer Agent as a D-bus service. That API is used
+to start, configure, and control containers. On the bus, the Agent has the name
+`com.pelagicore.SoftwareContainer` and object path `/com/pelagicore/SoftwareContainerAgent`.
+
+The Agent in turn links to the SoftwareContainer library which is to be considered an internal
+API for most users.
+
+TODO: Explain the thoughts around root privileges and the Agent.
+
+
+Container config
+================
+
+When creating containers, a configuration is passed as a JSON string. The string is an array with objects, in the JSON sense.
+The currently supported configs are:
+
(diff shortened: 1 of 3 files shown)


doc/chapters/getting-started/index.rst
```diff
@@ -0,0 +1,106 @@
...
+From the build directory, run the agent::
+
+    sudo ./agent/softwarecontainer-agent
+
+Not that we are running this command with `sudo`, the Agent needs to be started with root privileges.
+
+
+Using `dbus-send`, introspect the Agent D-Bus service API::
+
+    dbus-send --system --print-reply --dest=com.pelagicore.SoftwareContainerAgent /com/pelagicore/SoftwareContainerAgent org.freedesktop.DBus.Introspectable.Introspect
```
DunderRoffe: I think you should split this over several lines to make it easier and faster to read.
Using '\' will persevere the "copy and paste"-ability if that is a concern.
Eg:

``` bash
dbus-send --system --print-reply \
    --dest=com.pelagicore.SoftwareContainerAgent \
      /com/pelagicore/SoftwareContainerAgent \
      org.freedesktop.DBus.Introspectable.Introspect
```

DunderRoffe:
Besides the note about multi line over long single line above, it looks really good :)

doc/chapters/getting-started/index.rst
tobsan: +1

doc/chapters/getting-started/index.rst
```diff
@@ -0,0 +1,106 @@
...
+
+Prerequisites:
+  * configure and build (see project `README.md`)
+
+
+From the build directory, run the agent::
+
+    sudo ./agent/softwarecontainer-agent
+
+Not that we are running this command with `sudo`, the Agent needs to be started with root privileges.
```
tobsan: Note*

breakreturn:
Updated now based on feedback.

PR #16, opened 2016-09-14
adds unit tests

tobsan:
Adds a test to explicitly set networkgateway to down state, and one to check if setting environment variables work inside the container.


thenor:
+1

PR #18, opened 2016-09-15
explicit byte array to string conversion

frznlogic:


tobsan:
Is there a problem with the earlier way we did it, or do we just want to be explicit about the types in general?

PR #23, opened 2016-09-16
Added diagram support in user docs

breakreturn:
We use 'seqdiag' for diagram generation in the Sphinx docs.
Updated the cookbook revision and the Sphinx config and added
a diagram in the docs.


frznlogic (approved):
This looks good to me. Correct commit hash. Nice seqdiag and works as a nice example of how to use it in the future ;)

PR #24, opened 2016-09-19
Added a chapter on design

e8johan:
This is just a very rough introduction. I wanted to get something in place. Please review and comment!

### DIFF
--- a/doc/chapters/design/index.rst
+++ b/doc/chapters/design/index.rst
@@ -1,0 +1,39 @@
+SoftwareContainer Design
+************************
+
+SoftwareContainer has been designed as a building block enabling sandboxing of application contents in a larger system.
+
+The SoftwareContainer achieves this by providing a structured way to limit what inter-process communication (IPC) mechanims are made available to each application. By targetting the IPC mechanisms rather than the applications or services directly, the sandboxing mechanism can be deployed in a non-intrusive way, i.e. completely transparent to both applications and services.
+
+Each IPC is managed by a *gateway*. The IPC needs of each service is described in *service manifest* files, defining *capabilities* and mapping these to individual *gateway configurations*. This can be individual objects or interfaces on d-bus, networking filter rules, access to device nodes or something completely different.
+
+Each application is then associated with a set of *capabilities*. This could be done through an *application manifest*, or any other method supported by the *launcher*. These are used to configure the sandbox in which the application is then launched.
+
+In a typical deployment the SoftwareContainer is coupled with a *launcher*, which is responsible for initiating containers and provide an entry point to start execution from. This means that SoftwareContainer is not intended to be used as a standalone tool, but rather as a toolbox to build the full experience from.
+
+The *launcher* is responsible for providing SoftwareContainer with a list of *capabilities*, a mount point for the application payload and an entrypoint to start execution from.The SoftwareContainer does not concern itself with application packaging, deployment of manifest formats. All it does it associate the given *capabilities* with *gateway configurations* to setup the sandbox with the appropriate IPC access to the surrounding world.
+
+The Launcher
+============
+
+The launcher is responsible for setting up sandboxes for contained applications and then launch the application entry point.
+
+The SoftwareContainer API allows the launcher great control over the creation of the container. This makes it possible to implement preloading schemes, doing much of the sandbox creation and configuration prior to actually loading an application.
+
+The setup of a container is done in four phases:
+
+1. **Creation**, where the container sandbox is instantiated. This sandbox has no access to the host platform.
+2. **Activation**, where the sandbox is given a set of capabilities and the gateways are configured accordingly.
+3. **Mounting**, where the application contents is mounted into the filesystem of the sandbox.
+4. **Launching**, where the application entry point is executed.
+
+The main launcher for SoftwareContainer is a part of the `QtApplicationManager <http://code.qt.io/cgit/qt/qtapplicationmanager.git/>`_.
+
+Service Manifests
+=================
+
+Service manifests are JSON files mapping capabilities to gateway configurations. Capabilities is what applications specify when they want access to a specific feature of the platform. By specifying a list of capabilities to SoftwareContainer, a sandbox with the correct gateway configurations is created, providing access to the selected parts of the platform services.
+
+Capabilities are named according to a reverse DNS scheme, e.g. ``org.genivi.dlt`` or ``com.pelagicore.popups``. In some cases, the same service provides capabilities of different levels, e.g. generic permission to show popup compared to permission to show high priority popups. This is generally added to the end of the name, e.g. ``com.pelagicore.popups`` and ``com.pelagicore.popups.important``.
+
+The service manifests are typically placed in one JSON file per service, but can be split into one file per capability if needed. The files are placed in ``/share/resource-access-manager/caps.d/``.
--- a/doc/index.rst
+++ b/doc/index.rst
@@ -9,6 +9,6 @@ Contents:
     :glob:
 
     chapters/intro/*
+    chapters/design/*
     chapters/lxc/*
     chapters/gateways/*
-


doc/chapters/design/index.rst
```diff
@@ -0,0 +1,39 @@
+SoftwareContainer Design
+************************
+
+SoftwareContainer has been designed as a building block enabling sandboxing of application contents in a larger system.
+
+The SoftwareContainer container achieves this by providing a structured way to limit what IPC mechanims are made available to each application. By targetting the IPC mechanisms rather than the applications or services directly, the sandboxing mechanism can be deployed in a non-intrusive way, i.e. completely transparent to both applications and services.
+
+Each IPC is managed by a *gateway*. The IPC needs of each service is described in *service manifest* files, defining *capabilities* and mapping these to individual *gateway configurations*. This can be individual objects or interfaces on d-bus, ipfilter rules, access to device nodes or something completely different.
```
tobsan: This is not true yet, since the component that handles capabilities has has not been integrated into SC.

doc/chapters/design/index.rst
```diff
@@ -0,0 +1,39 @@
...
+
+Each application is then associated with a set of *capabilities* in an *application manifest*. These are used to configure the sandbox in which the application is then launched.
```
tobsan: This is irrelevant to how SC works and would fit better in the discussion around the launcher below.

doc/chapters/design/index.rst
```diff
@@ -0,0 +1,39 @@
+SoftwareContainer Design
+************************
+
+SoftwareContainer has been designed as a building block enabling sandboxing of application contents in a larger system.
+
+The SoftwareContainer container achieves this by providing a structured way to limit what IPC mechanims are made available to each application. By targetting the IPC mechanisms rather than the applications or services directly, the sandboxing mechanism can be deployed in a non-intrusive way, i.e. completely transparent to both applications and services.
```
DunderRoffe: Since the target audience is larger than purely automotive people, maybe include something about what an IPC mechanism is?

tobsan (changes requested):
My general view is that this goes too far in mixing what is softwarecontainer design and what is coming from outside.

Also, the PR has merge conflicts, please resolv them first.

frznlogic:
There are merge conflicts.

doc/chapters/design/index.rst
frznlogic: Most likely not ipfilter rules as they are part of a different OS.

doc/chapters/design/index.rst
```diff
@@ -0,0 +1,39 @@
...
+
+Each IPC is managed by a *gateway*. The IPC needs of each service is described in *service manifest* files, defining *capabilities* and mapping these to individual *gateway configurations*. This can be individual objects or interfaces on d-bus, ipfilter rules, access to device nodes or something completely different.
+
+Each application is then associated with a set of *capabilities* in an *application manifest*. These are used to configure the sandbox in which the application is then launched.
+
+In a typical deployment the SoftwareContainer is coupled with a *launcher*, which is responsible for initiating containers and provide an entry point to start execution from. This means that SoftwareContainer is not intended to be used as a standalone tool, but rather as a toolbox to build the full experience from.
```
frznlogic, line 12: The idea of a CLI might change that I guess, but probably not... or?

doc/chapters/design/index.rst
```diff
@@ -0,0 +1,39 @@
...
+The launcher is responsible for setting up sandboxes for contained applications and then launch the application entry point.
+
+The SoftwareContainer API allows the launcher great control over the creation of the container. This makes it possible to implement preloading schemes, doing much of the sandbox creation and configuration prior to actually loading an application.
+
+The setup of a container is done in four phases:
+
+1. **Creation**, where the container sandbox is instantiated. This sandbox has no access to the host platform.
+2. **Activation**, where the sandbox is given a set of capabilities and the gateways are configured accordingly.
+3. **Mounting**, where the application contents is mounted into the filesystem of the sandbox.
+4. **Launching**, where the application entry point is executed.
```
DunderRoffe, line 28: We have had some discussions about a "convenience-API". I think we should use these phases as a base for calls to the DBUS-API in that purpose.

doc/chapters/design/index.rst
```diff
@@ -0,0 +1,39 @@
...
+
+The main launcher for SoftwareContainer is a part of the `QtApplicationManager <http://code.qt.io/cgit/qt/qtapplicationmanager.git/>`_.
```
frznlogic, line 30: Not sure I like having QtApplicationManager written together tbh. It feels like reading a fragment of a sentence. I digress.

frznlogic (commented):
Looks generally good.

doc/chapters/design/index.rst
breakreturn: Sure, the capabilities are not implemented yet (in this component), but that's soon there so I do think it's relevant. I'm not so sure about the application manifests though. Perhaps it should be made more clear that it's not part of this components domain?

doc/chapters/design/index.rst
```diff
@@ -0,0 +1,39 @@
...
+
+Service Manifests
+=================
+
+Service manifests are JSON files mapping capabilities to gateway configurations. Capabilities is what applications specify when they want access to a specific feature of the platform. By specifying a list of capabilities to SoftwareContainer, a sandbox with the correct gateway configurations is created, providing access to the selected parts of the platform services.
```
breakreturn, line 35: Please break the very long lines so it's nicer to work with in peoples editors. It has its drawbacks but this is the desired way from the maintainers.

doc/chapters/design/index.rst
DunderRoffe: Sure, but isn't it better to write future proof documentation since this will be implemented this sprint anyhow?

DunderRoffe: But isn't SC is Linux specific anyway?

doc/chapters/design/index.rst
DunderRoffe: I agree with you, but at the same time isn't it hard to really get a grasp of SC without some insight into the concepts around it?

doc/chapters/design/index.rst
DunderRoffe: +1

doc/chapters/design/index.rst
```diff
@@ -0,0 +1,39 @@
+SoftwareContainer Design
+************************
+
+SoftwareContainer has been designed as a building block enabling sandboxing of application contents in a larger system.
```
DunderRoffe, line 4: Personal preference: I like the text to come directly after a heading without any empty line in between. AFAIK this is not defined anywhere (nothing found when searching for rst or markdown guidelines on the wiki).

DunderRoffe (changes requested):

doc/chapters/design/index.rst
tobsan: Sure.

tobsan: I think the point is that ipfilter is more or less BSD specific.

doc/chapters/design/index.rst
breakreturn: We don't have any guidelines for that AFAIK. Perhaps we should talk pros and cons but this commit is at least consistent with the rest of the user docs.

doc/chapters/design/index.rst
breakreturn: I think the term IPC is pretty generic and most users will probably guess it means inter-process communication, but a compromise where the first occurrence is spelled out might help some readers.

doc/chapters/design/index.rst
breakreturn: We had an offline discussion that we believe "capabilities" are part of the SC domain, and that it is useful to mention the integration pattern that we promote, which contains e.g. "application manifests". For that reason we decided it could be good to try and separate the pure SC domain things and and keep the rest but have it in e.g. a "Integration pattern" section or similar.

doc/chapters/design/index.rst
DunderRoffe: That sounds as a good solution

doc/chapters/design/index.rst
DunderRoffe: Ok

frznlogic: Yes :D

doc/chapters/design/index.rst
frznlogic: Having a separate section with acronyms and branch specific terms might be a good idea imho.

DunderRoffe (approved):

doc/chapters/design/index.rst
thenor: "SoftwareContainer container": seems unnecessary to me to have the "container".

doc/chapters/design/index.rst
e8johan: well, the CLI would just be a launcher, I guess. But with very detailed control of SC

PR #29, opened 2016-09-21
docs: added chapter on design

e8johan:
Changed wording around app manifests
Documented what IPC is
Removed reference to ipfilter :-)

### DIFF
--- a/doc/chapters/design/index.rst
+++ b/doc/chapters/design/index.rst
@@ -1,0 +1,39 @@
+SoftwareContainer Design
+************************
+
+SoftwareContainer has been designed as a building block enabling sandboxing of application contents in a larger system.
+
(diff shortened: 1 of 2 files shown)


frznlogic:
Please fix the merge conflict? :)

e8johan:
Will fix the merge conflict and do some linebreaking :-)

breakreturn:
All in all I think this text introduces some nomenclature we are not using currently, and might be slightly confusing as what SC brings compared to what a launcher would bring w.r.t. integration and what's mandated and what's implied patterns etc. But I think that just means we need to revise and iterate as we go along. I am fine with merging this as it is now.

tobsan (changes requested):
Lines are still super long and this request still has merge conflicts. Please fix both things.

PR #32, opened 2016-09-22
Vagrantfile: use debian stable instead of testing

tobsan:
This removes the testing upgrade of the VM. It also makes latexpdf
generation of documentation optional.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -1,16 +1,11 @@
 # -*- mode: ruby -*-
 # vi: set ft=ruby :
 
-ram = 4 #GB
-cpus = 3
-
 Vagrant.configure(2) do |config|
     config.vm.box = "debian/contrib-jessie64"
     config.vm.provider "virtualbox" do |vb|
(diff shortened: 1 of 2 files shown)


Vagrantfile
```diff
@@ -1,16 +1,11 @@
 # -*- mode: ruby -*-
 # vi: set ft=ruby :
 
-ram = 4 #GB
```
frznlogic, line 4: Actually, maybe we should make this a bit useful instead? set vb.cpus and vb.memory to this X lala etc, and then add environment variables so if you have VAGRANT_RAM=16 you get 16GB ram used for the vagrant machine?

frznlogic (changes requested):
Some changes plix?

Vagrantfile
tobsan: I can live with that, I'll make the necessary changes.

tobsan: Updated with changes.

PR #33, opened 2016-09-22
New syntax for networkgateway configuration

DunderRoffe:


doc/chapters/gateways/index.rst
```diff
@@ -302,12 +302,39 @@ An example configuration can look like this::
 Network gateway
 ===============
 
-The Network Gateway is used to provide Internet access through a container-specific firewall.
+The Network Gateway is used to setup network connection and configure which traffic is allowed and not.
 
 Configuration
 -------------
+Types: "INCOMMING" and "OUTGOING"
```
tobsan: It should say INCOMING (one M)

doc/chapters/gateways/index.rst
```diff
@@ -302,12 +302,39 @@ An example configuration can look like this::
...
+            "type": "INCOMMING",
+            "rules": [
+                         { "host": "127.0.0.1/16", "port": 80, "target": "ACCEPT"},
+                         { "host": "google.com", "port": "80-85", "target": "ACCEPT"},
+                         { "host": "127.0.0.1/16", "port": [80, 8080], "target": "ACCEPT"},
+                         { "host": "50.63.202.33/24", "target": "REJECT"},
+                     ],
+            "default": "DROP"
+        }
+    ]
```
tobsan, line 384: What about a general allow-all rule? Would that be possible? I guess no rules and default: ACCEPT?

libsoftwarecontainer/src/gateway/networkgateway.cpp
```diff
@@ -41,28 +40,152 @@ NetworkGateway::~NetworkGateway()
 
 ReturnCode NetworkGateway::readConfigElement(const json_t *element)
 {
-    if (!read(element, "internet-access", m_internetAccess)) {
-        log_error() << "either key \"internet-access\" missing or not a bool in json configuration";
+    Entry e;
+    if (!read(element, "type", e.type)) {
+        log_error() << " \"internet-access\" ";
```
tobsan: This error message isn't very useful.

doc/chapters/gateways/index.rst
frznlogic: incoming with 1 m

doc/chapters/gateways/index.rst
```diff
@@ -302,12 +302,39 @@ An example configuration can look like this::
...
+            "rules": [
+                         { "host": "127.0.0.1/16", "port": 80, "target": "ACCEPT"},
+                         { "host": "google.com", "port": "80-85", "target": "ACCEPT"},
+                         { "host": "127.0.0.1/16", "port": [80, 8080], "target": "ACCEPT"},
+                         { "host": "50.63.202.33/24", "target": "DROP"},
+                     ],
+            "default": "REJECT"
+        },
+        {
+            "type": "INCOMMING",
```
frznlogic: incoming with 1 m

doc/chapters/gateways/index.rst
```diff
@@ -302,12 +302,39 @@ An example configuration can look like this::
...
 ===============
 
-The Network Gateway is used to provide Internet access through a container-specific firewall.
+The Network Gateway is used to setup network connection and configure which traffic is allowed and not.
 
 Configuration
 -------------
+Types: "INCOMMING" and "OUTGOING"
+Targets: "ACCEPT", "DROP" and "REJECT"
+host: hostname or ip-address with netmask
```
frznlogic: At some point it would be nice to know how hostname is expanded. It might also be nice to have ip address without netmask.

frznlogic (changes requested):
I would also like a set of service tests, but I guess thats for later when @breakreturn is done with the refactoring?

libsoftwarecontainer/unit-test/networkgateway_unittest.cpp
```diff
@@ -43,6 +43,62 @@ class NetworkGatewayTest : public SoftwareContainerGatewayTest
...
+                     "{ \"host\": \"google.com\", \"port\": \"80-85\", \"target\": \"ACCEPT\"},"
+                     "{ \"host\": \"127.0.0.1/16\", \"port\": [80, 8080], \"target\": \"ACCEPT\"},"
+                     "{ \"host\": \"50.63.202.33/24\", \"target\": \"REJECT\"}"
+                 "],"
+        "\"default\": \"ACCEPT\""
+    "}, {"
+        "\"type\": \"INCOMMING\","
+        "\"rules\": ["
+                     "{ \"host\": \"127.0.0.1/16\", \"port\": 80, \"target\": \"ACCEPT\"},"
+                     "{ \"host\": \"google.com\", \"port\": \"80-85\", \"target\": \"ACCEPT\"},"
```
tobsan: Something I've been thinking about for a while: shouldn't we really use example.org instead?

libsoftwarecontainer/unit-test/networkgateway_unittest.cpp
```diff
@@ -43,6 +43,62 @@ class NetworkGatewayTest : public SoftwareContainerGatewayTest
...
+                     "{ \"host\": \"50.63.202.33/24\", \"target\": \"REJECT\"}"
+                 "],"
+        "\"default\": \"DROP\""
+    "}, {"
+        "\"type\": \"INCOMMING\","
+        "\"rules\": ["
+                     "{ \"host\": \"127.0.0.1/16\", \"port\": 80, \"target\": \"ACCEPT\"},"
+                     "{ \"host\": \"google.com\", \"port\": \"80-85\", \"target\": \"ACCEPT\"},"
+                     "{ \"host\": \"127.0.0.1/16\", \"port\": [80, 8080], \"target\": \"ACCEPT\"},"
+                     "{ \"host\": \"50.63.202.33/24\", \"target\": \"REJECT\"}"
```
tobsan: Where does this IP come from? I noticed that it is a real IP that runs a webserver at least. Should we use some example ip instead? Something like example.org but for an IP?

tobsan (changes requested):
Looks good in general, I only have some minor issues.

doc/chapters/gateways/index.rst
DunderRoffe: Atm it is not allowed to not specify any rules. Maybe that is a mistake ?

tobsan: It would be nice to be able to have a general on/off setting.

libsoftwarecontainer/unit-test/networkgateway_unittest.cpp
DunderRoffe: I picked the ip of google and httpbin, can not remember which one this is.

libsoftwarecontainer/unit-test/networkgateway_unittest.cpp
DunderRoffe: We could. However, this code is from the unittest so it might be usefull to have an exisiting address ?

libsoftwarecontainer/src/gateway/networkgateway.cpp
DunderRoffe: Fixed

doc/chapters/gateways/index.rst
DunderRoffe: Fixed

doc/chapters/gateways/index.rst
DunderRoffe: OK I can change this if it something that is desired

doc/chapters/gateways/index.rst
DunderRoffe: fixed

libsoftwarecontainer/unit-test/networkgateway_unittest.cpp
tobsan: example.org is an existing address, existing only for example purposes.

doc/chapters/gateways/index.rst
DunderRoffe: fixed

doc/chapters/gateways/index.rst
DunderRoffe: I agree, will have to figure that out for myself first though :)

libsoftwarecontainer/unit-test/networkgateway_unittest.cpp
DunderRoffe: Ok I'll use that instead then

doc/chapters/gateways/index.rst
```diff
@@ -302,12 +302,39 @@ An example configuration can look like this::
...
 
 Configuration
 -------------
+Types: "INCOMING" and "OUTGOING"
+Targets: "ACCEPT", "DROP" and "REJECT"
+host: hostname or ip-address with netmask
+default: target to choose when not matching any other rule
 
-T.B.D.
+Example of network gateway config::
```
breakreturn: How do we express how to resolve contradicting rules? Is that in the domain of the service integrator that writes configs, and we just document it? Or do we need to allow prioritization? What currently happens when rules are contradictory, that should be documented and tested.

DunderRoffe: Suggestion on collision resolution:
Merge the rules lists of all entries of same type using the most permissive default.

Example:

``` json
 [
        {
            "type": "OUTGOING",
            "rules": [
                         { "host": "127.0.0.1/16", "port": 80, "target": "ACCEPT"},
                         { "host": "example.com", "port": "80-85", "target": "ACCEPT"}
                     ],
            "default": "REJECT"
        },
        {
            "type": "OUTGOING",
            "rules": [
                         { "host": "127.0.0.1/16", "port": [80, 8080], "target": "ACCEPT"},
                         { "host": "93.184.216.34/24", "target": "REJECT"}
                     ],
            "default": "DROP"
        }
    ]
```

Will become

``` json
 [
        {
            "type": "OUTGOING",
            "rules": [
                         { "host": "127.0.0.1/16", "port": 80, "target": "ACCEPT"},
                         { "host": "example.com", "port": "80-85", "target": "ACCEPT"},
                         { "host": "127.0.0.1/16", "port": [80, 8080], "target": "ACCEPT"},
                         { "host": "93.184.216.34/24", "target": "REJECT"}
                     ],
            "default": "REJECT"
        }
    ]
```

DunderRoffe: As we discussed offline, there will be a need for some kind of priority attribute for each entry. Otherwise app developers need to specify their capabilities in a specific order.

libsoftwarecontainer/unit-test/networkgateway_unittest.cpp
DunderRoffe: Fixed

libsoftwarecontainer/unit-test/networkgateway_unittest.cpp
DunderRoffe: Fixed

doc/chapters/gateways/index.rst
DunderRoffe: Actually empty lists are allowed but not specifying the rules parameter in **not** allowed. Which means that this is not an issue.

libsoftwarecontainer/unit-test/networkgateway_unittest.cpp
frznlogic: Just as a note on the topic https://tools.ietf.org/html/rfc6761, section 6.5. example.org is OK though.

libsoftwarecontainer/unit-test/networkgateway_unittest.cpp
frznlogic: Check https://tools.ietf.org/html/rfc6761for DNS examples and https://tools.ietf.org/html/rfc5737 for example IP addresses.

doc/chapters/gateways/index.rst
frznlogic: This is interesting imho. To begin with we could be lazy and just let the rules wind up in the order they are added, there is the possibility to add rules in different ordering from each rule (ie, inserting, prepending, appending, etc), which might get confusing... we could add a match/add value which marks the rule for appending or prepending perhaps? The inserting could become really hairy otherwise.

frznlogic: Just noticed you already added a priority attribute so I guess the point is moot.

DunderRoffe:
I have made a temporary workaround to make the new syntax map to the old backend as good as possible.

I would like feedback on what is missing in terms of documentation etc for this issue to be considered done.
- [x] Component tests for the story exists. (They exist as "unit-tests" in softwarecontainerlib_unittest.cpp i.e. the old unsorted stuff)
- [x] Unit tests for the story exists. (Only testing the new syntax, can be found in networkgateway_unittest.cpp)
- [x] Appropriate CI jobs have been updated/added. (There are none as far as I'm concerned)
- [ ] Documentation is added in either the repo or on the wiki, if documentation exist already it should be reviewed and updated if needed, see checklist.
- [ ] Documentation is generated and available on github. Do we have a process for this and is it relevant in this case ?
- [x] Build passes without any new warnings or errors.
- [x] All tests passes at least as well as before.

libsoftwarecontainer/src/gateway/networkgateway.h
```diff
@@ -115,10 +116,58 @@ class NetworkGateway :
...
+
+    std::vector<Entry> m_entries;
+
     bool m_internetAccess;
     bool m_interfaceInitialized;
 
     Generator m_generator;
+
+    /**
+     * @breif Parses a json element to a Rule
```
frznlogic: brief, not breif.

libsoftwarecontainer/src/gateway/networkgateway.h
```diff
@@ -115,10 +116,58 @@ class NetworkGateway :
...
+     * @return ReturnCode::SUCCESS if the rule is successfully parsed
+     * @return ReturnCode::FAILURE otherwise.
+     */
+    virtual ReturnCode parseRule(const json_t *element, std::vector<Rule> &rules);
+
+    /**
+     * @breif Parses a string to a Target
```
frznlogic: breif.

libsoftwarecontainer/unit-test/networkgateway_unittest.cpp
```diff
@@ -60,14 +122,255 @@ TEST_F(NetworkGatewayTest, TestNoContainer) {
  */
 TEST_F(NetworkGatewayTest, TestSetConfig) {
     givenContainerIsSet(gw);
-    const std::string config = "[{\"internet-access\": false, \"gateway\":\"\"}]";
+
+    ASSERT_TRUE(gw->setConfig(VALID_FULL_CONFIG));
+}
+
+/*! Test that config entries with no priority specified fails gracefully.
```
frznlogic: /*\* formatting etc.

libsoftwarecontainer/unit-test/networkgateway_unittest.cpp
```diff
@@ -60,14 +122,255 @@ TEST_F(NetworkGatewayTest, TestNoContainer) {
...
+                     "{ \"host\": \"127.0.0.1/16\", \"port\": [80, 8080], \"target\": \"ACCEPT\"},"
+                     "{ \"host\": \"50.63.202.33/24\", \"target\": \"REJECT\"}"
+                 "],"
+        "\"default\": \"ACCEPT\""
+    "}]";
+
+    ASSERT_FALSE(gw->setConfig(config));
+}
+
+/*! Test that config entries with no type specified fails gracefully.
```
frznlogic: Dito.

libsoftwarecontainer/unit-test/networkgateway_unittest.cpp
```diff
@@ -60,14 +122,255 @@ TEST_F(NetworkGatewayTest, TestNoContainer) {
...
+                     "{ \"host\": \"127.0.0.1/16\", \"port\": [80, 8080], \"target\": \"ACCEPT\"},"
+                     "{ \"host\": \"50.63.202.33/24\", \"target\": \"REJECT\"}"
+                 "],"
+        "\"default\": \"ACCEPT\""
+    "}]";
+
+    ASSERT_FALSE(gw->setConfig(config));
+}
+
+/*! Test that config entries with no rules specified fails gracefully.
```
frznlogic: Dito.

libsoftwarecontainer/unit-test/networkgateway_unittest.cpp
```diff
@@ -60,14 +122,255 @@ TEST_F(NetworkGatewayTest, TestNoContainer) {
...
+    "[{"
+        "\"type\": \"OUTGOING\","
+        "\"priority\": 1,"
+        "\"default\": \"ACCEPT\""
+    "}]";
+
+    ASSERT_FALSE(gw->setConfig(config));
+}
+
+/*! Test that config entries with empty rules specified works.
```
frznlogic: Dito.

libsoftwarecontainer/unit-test/networkgateway_unittest.cpp
```diff
@@ -60,14 +122,255 @@ TEST_F(NetworkGatewayTest, TestNoContainer) {
...
+        "\"priority\": 1,"
+        "\"rules\": [],"
+        "\"default\": \"ACCEPT\""
+    "}]";
 
     ASSERT_TRUE(gw->setConfig(config));
 }
 
-TEST_F(NetworkGatewayTest, TestSeConfigNoGw) {
+/*! Test that config entries with no default target specified fails gracefully.
```
frznlogic: Dito.

libsoftwarecontainer/unit-test/networkgateway_unittest.cpp
```diff
@@ -60,14 +122,255 @@ TEST_F(NetworkGatewayTest, TestNoContainer) {
...
+                     "{ \"host\": \"google.com\", \"port\": \"80-85\", \"target\": \"ACCEPT\"},"
+                     "{ \"host\": \"127.0.0.1/16\", \"port\": [80, 8080], \"target\": \"ACCEPT\"},"
+                     "{ \"host\": \"50.63.202.33/24\", \"target\": \"REJECT\"}"
+                 "]"
+    "}]";
+
+    ASSERT_FALSE(gw->setConfig(config));
+}
+
+/*! Test that config entries with type specified as an integer fails gracefully.
```
frznlogic: Dito.

libsoftwarecontainer/unit-test/networkgateway_unittest.cpp
```diff
@@ -60,14 +122,255 @@ TEST_F(NetworkGatewayTest, TestNoContainer) {
...
+                     "{ \"host\": \"127.0.0.1/16\", \"port\": [80, 8080], \"target\": \"ACCEPT\"},"
+                     "{ \"host\": \"50.63.202.33/24\", \"target\": \"REJECT\"}"
+                 "],"
+        "\"default\": \"ACCEPT\""
+    "}]";
+
+    ASSERT_FALSE(gw->setConfig(config));
+}
+
+/*! Test that config entries with type specified as a string fails gracefully.
```
frznlogic: Dito.

libsoftwarecontainer/unit-test/networkgateway_unittest.cpp
```diff
@@ -60,14 +122,255 @@ TEST_F(NetworkGatewayTest, TestNoContainer) {
...
+                     "{ \"host\": \"127.0.0.1/16\", \"port\": [80, 8080], \"target\": \"ACCEPT\"},"
+                     "{ \"host\": \"50.63.202.33/24\", \"target\": \"REJECT\"}"
+                 "],"
+        "\"default\": \"ACCEPT\""
+    "}]";
+
+    ASSERT_FALSE(gw->setConfig(config));
+}
+
+/*! Test that config entries with type specified as an object fails gracefully.
```
frznlogic: Dito.

libsoftwarecontainer/unit-test/networkgateway_unittest.cpp
```diff
@@ -60,14 +122,255 @@ TEST_F(NetworkGatewayTest, TestNoContainer) {
...
+                     "{ \"host\": \"127.0.0.1/16\", \"port\": [80, 8080], \"target\": \"ACCEPT\"},"
+                     "{ \"host\": \"50.63.202.33/24\", \"target\": \"REJECT\"}"
+                 "],"
+        "\"default\": \"ACCEPT\""
+    "}]";
+
+    ASSERT_FALSE(gw->setConfig(config));
+}
+
+/*! Test that config entries with type specified as an array fails gracefully.
```
frznlogic: Dito.

libsoftwarecontainer/unit-test/networkgateway_unittest.cpp
```diff
@@ -60,14 +122,255 @@ TEST_F(NetworkGatewayTest, TestNoContainer) {
...
+                     "{ \"host\": \"127.0.0.1/16\", \"port\": [80, 8080], \"target\": \"ACCEPT\"},"
+                     "{ \"host\": \"50.63.202.33/24\", \"target\": \"REJECT\"}"
+                 "],"
+        "\"default\": \"ACCEPT\""
+    "}]";
+
+    ASSERT_FALSE(gw->setConfig(config));
+}
+
+/*! Test that config entries with rules specified as an object fails gracefully.
```
frznlogic: Dito.

libsoftwarecontainer/unit-test/networkgateway_unittest.cpp
```diff
@@ -60,14 +122,255 @@ TEST_F(NetworkGatewayTest, TestNoContainer) {
...
+                     "\"rule2\": { \"host\": \"google.com\", \"port\": \"80-85\", \"target\": \"ACCEPT\"},"
+                     "\"rule3\": { \"host\": \"127.0.0.1/16\", \"port\": [80, 8080], \"target\": \"ACCEPT\"},"
+                     "\"rule4\": { \"host\": \"50.63.202.33/24\", \"target\": \"REJECT\"}"
+                 "},"
+        "\"default\": \"ACCEPT\""
+    "}]";
+    ASSERT_FALSE(gw->setConfig(config));
+}
+
+/*! Test that config entries with rules specified as a string fails gracefully.
```
frznlogic: Dito.

libsoftwarecontainer/unit-test/networkgateway_unittest.cpp
```diff
@@ -60,14 +122,255 @@ TEST_F(NetworkGatewayTest, TestNoContainer) {
...
+                     "{ \"host\": \"google.com\", \"port\": \"80-85\", \"target\": \"ACCEPT\"},"
+                     "{ \"host\": \"127.0.0.1/16\", \"port\": [80, 8080], \"target\": \"ACCEPT\"},"
+                     "{ \"host\": \"50.63.202.33/24\", \"target\": \"REJECT\"}"
+                 "\","
+        "\"default\": \"ACCEPT\""
+    "}]";
+    ASSERT_FALSE(gw->setConfig(config));
+}
+
+/*! Test that config entries with rules specified as an integer fails gracefully.
```
frznlogic: Dito.

libsoftwarecontainer/unit-test/networkgateway_unittest.cpp
```diff
@@ -60,14 +122,255 @@ TEST_F(NetworkGatewayTest, TestNoContainer) {
...
+    "[{"
+        "\"type\": \"INCOMING\","
+        "\"priority\": 1,"
+        "\"rules\": 123"
+        "\"default\": \"ACCEPT\""
+    "}]";
+    ASSERT_FALSE(gw->setConfig(config));
+}
+
+/*! Test that config entries with default specified as a invali target fails gracefully.
```
frznlogic: Dito.

libsoftwarecontainer/unit-test/networkgateway_unittest.cpp
```diff
@@ -60,14 +122,255 @@ TEST_F(NetworkGatewayTest, TestNoContainer) {
...
+}
+
+/*! Test that config entries with type specified as an object fails gracefully.
+ */
+TEST_F(NetworkGatewayTest, TestSetConfigTypeIsObject) {
+    givenContainerIsSet(gw);
+
+    const std::string config =
+    "[{"
+        "\"type\": {\"type\": \"INCOMMING\"},"
```
frznlogic: INCOMMING

frznlogic (changes requested):
There's some minor changs I'd like to have first, but overall good.

libsoftwarecontainer/unit-test/networkgateway_unittest.cpp
DunderRoffe: Fixed

libsoftwarecontainer/unit-test/networkgateway_unittest.cpp
DunderRoffe: Fixed

libsoftwarecontainer/unit-test/networkgateway_unittest.cpp
DunderRoffe: Fixed

libsoftwarecontainer/unit-test/networkgateway_unittest.cpp
DunderRoffe: Fixed

libsoftwarecontainer/unit-test/networkgateway_unittest.cpp
DunderRoffe: Fixed

libsoftwarecontainer/unit-test/networkgateway_unittest.cpp
DunderRoffe: Fixed

libsoftwarecontainer/unit-test/networkgateway_unittest.cpp
DunderRoffe: Fixed

libsoftwarecontainer/unit-test/networkgateway_unittest.cpp
DunderRoffe: Fixed

libsoftwarecontainer/unit-test/networkgateway_unittest.cpp
DunderRoffe: Fixed

libsoftwarecontainer/unit-test/networkgateway_unittest.cpp
DunderRoffe: Fixed

libsoftwarecontainer/unit-test/networkgateway_unittest.cpp
DunderRoffe: Fixed

libsoftwarecontainer/unit-test/networkgateway_unittest.cpp
DunderRoffe: Fixed

libsoftwarecontainer/unit-test/networkgateway_unittest.cpp
DunderRoffe: Fixed

libsoftwarecontainer/unit-test/networkgateway_unittest.cpp
DunderRoffe: Fixed

libsoftwarecontainer/src/gateway/networkgateway.h
DunderRoffe: Fixed

libsoftwarecontainer/src/gateway/networkgateway.h
DunderRoffe: Fixed

tobsan (approved):

frznlogic (approved):

doc/chapters/gateways/index.rst
```diff
@@ -302,12 +302,51 @@ An example configuration can look like this::
 Network gateway
 ===============
 
-The Network Gateway is used to provide Internet access through a container-specific firewall.
+The Network Gateway is used to setup network connection and configure which traffic is allowed and not.
 
 Configuration
 -------------
+The configuration is structured as a list of JSON objects that each describe rules for ``OUTGOING``
+or ``INCOMING`` network traffic. During evaluation the rules are read in order starting with the
```
breakreturn: Is the order relevant given the "priority" stuff described below? If it's not relevant it would be nice to skip it as it is something the user hopefully should not assume and rely on.

libsoftwarecontainer/src/gateway/networkgateway.cpp
```diff
@@ -41,28 +40,173 @@ NetworkGateway::~NetworkGateway()
...
     }
 
-    if (m_internetAccess) {
-        log_info() << "Internet access will be enabled";
+    if (e.type != "INCOMING" && e.type != "OUTGOING") {
+        log_error() << e.type << " is not a valid type ('INCOMING' or 'OUTGOING')";
+        return ReturnCode::FAILURE;
+    }
+
+    const json_t *prio = json_object_get(element, "priority");
```
breakreturn: Why do we use jansson directly in these parts of the code?

libsoftwarecontainer/src/gateway/networkgateway.cpp
```diff
@@ -41,28 +40,173 @@ NetworkGateway::~NetworkGateway()
...
+    if (prio == nullptr) {
```
breakreturn: Is it always safe to check the jansson NULL with a nullptr comparison? Perhaps the JSONParser should be used and convert NULL checks to nullptr or something like that?

libsoftwarecontainer/src/gateway/networkgateway.cpp
```diff
@@ -41,28 +40,173 @@ NetworkGateway::~NetworkGateway()
...
+    // in the wait of activate() being rewritten
+    if (e.defaultTarget == Target::ACCEPT) {
+        m_internetAccess = true;
+        m_gateway = "10.0.3.1";
+    }
+    // ----------------------------
+    return ReturnCode::SUCCESS;
+}
+
+ReturnCode NetworkGateway::parseRule(const json_t *element, std::vector<Rule> &rules)
```
breakreturn, line 116: Pretty big method. Would it be nicer to break out the string and array specific parsing in the big nested else ifs?

libsoftwarecontainer/src/gateway/networkgateway.cpp
DunderRoffe: There are no implemented workaround for this

libsoftwarecontainer/src/gateway/networkgateway.cpp
DunderRoffe: NULL == nullptr, so it is the same thing.

libsoftwarecontainer/unit-test/networkgateway_unittest.cpp
```diff
@@ -60,14 +122,268 @@ TEST_F(NetworkGatewayTest, TestNoContainer) {
...
+    "}]";
+
+    ASSERT_FALSE(gw->setConfig(config));
+}
+
+/**
+ * @brief Test that config entries with no type specified fails gracefully.
+ */
+TEST_F(NetworkGatewayTest, TestSetConfigNoType) {
+    givenContainerIsSet(gw);
```
breakreturn, line 154: Perhaps out of scope for this change, but isn't the name givenContainerIsSet weird? Doesn't in practice mean "given this gateway is set"?

libsoftwarecontainer/src/gateway/networkgateway.cpp
DunderRoffe: I've been thinking about that. It is non trivial as it is so closely coupled with parsing the rule. On the other hand I agree with the function being to big.

libsoftwarecontainer/unit-test/networkgateway_unittest.cpp
DunderRoffe: I don't think so as the tests are written from the Gateways point of view. Using your suggested approach would introduce ambiguity as to what context the gateway is set in.

As a compromise some thing like "given that container is set for gateway" but that  is a bit wordy.

libsoftwarecontainer/src/gateway/networkgateway.cpp
breakreturn: What do you mean? Isn't the JSONParser a wrapper around jansson? Which we use elsewhere when parsing json?

libsoftwarecontainer/src/gateway/networkgateway.cpp
breakreturn: OK

libsoftwarecontainer/unit-test/networkgateway_unittest.cpp
breakreturn: So when I read that method call, I read "given that a container is set". I don't know what that means. The call just adds a gateway to a container right? So, then perhaps givenAddingThisGateway() or something would make sense. Or how should I read the method as it is now?

libsoftwarecontainer/src/gateway/networkgateway.cpp
DunderRoffe: No we only use it to parse strings and booleans

PR #36, opened 2016-09-26
Design documentation updates efter review

thenor:
Updates after review to fix formatting, grammar and typos.

### DIFF
--- a/doc/chapters/lxc/index.rst
+++ b/doc/chapters/lxc/index.rst
@@ -1,36 +1,39 @@
 SoftwareContainer using LXC
 ***************************
 
-SoftwareContainer uses the Linux Containers [#lxc] (LXC) project as its container backend. LXC uses namespaces and cgroups,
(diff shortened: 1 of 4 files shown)


doc/chapters/lxc/index.rst
```diff
@@ -76,5 +83,3 @@ LXC API
 -------
 TBD
 
-.. [#lxc] https://linuxcontainers.org/
-.. [#fhs] https://en.wikipedia.org/wiki/Filesystem_Hierarchy_Standard
```
tobsan, line 80: Why are we removing these references?

tobsan (approved):

doc/chapters/lxc/index.rst
thenor: They are in-line instead now. Do we also want to have them in a "Links"/"References" chapter?

tobsan: Actually, I don't have a preference. Inline are fine by me.

PR #39, opened 2016-09-26
Moved JSONParser to common

DunderRoffe:


common/jsonparser.cpp
```diff
@@ -0,0 +1,58 @@
...
+ * WHETHER IN AN ACTION OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION,
+ * ARISING OUT OF OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS
+ * SOFTWARE.
+ *
+ * For further information see LICENSE
+ */
+
+#include "jsonparser.h"
+
+namespace softwarecontainer {
```
frznlogic, line 22: Did we make any decisions on how to handle the namespacing? @breakreturn

common/jsonparser.cpp
```diff
@@ -0,0 +1,58 @@
...
+
+LOG_DECLARE_CLASS_CONTEXT("JSON", "JSONParser");
+
+bool JSONParser::read(const json_t *element, const char *key, std::string &result)
```
frznlogic, line 26: Do we have any policy on using the ReturnCode stuff instead of bool or why are we using that one in other places? (I do see this is only a copy paste but thought i'd ask)

common/jsonparser.h
```diff
@@ -0,0 +1,31 @@
...
+#pragma once
+
+#include "softwarecontainer-log.h"
+#include <jansson.h>
+
+namespace softwarecontainer {
+    class JSONParser {
+
+    protected:
+        bool read(const json_t *element, const char *key, std::string &result);
```
frznlogic, line 47: Documentation plix? :)

frznlogic (commented):
I generally agree to it, but would like some documentation to be added as well.

frznlogic (approved):

PR #40, opened 2016-09-27
Split softwarecontainer-common into separate classes.

frznlogic:


common/filetoolkitwithundo.cpp
```diff
@@ -0,0 +1,204 @@
...
+    if(!success) {
+        log_error() << "One or more cleanup handlers returned error status, please check the log";
+    }
+}
+
+ReturnCode FileToolkitWithUndo::createParentDirectory(const std::string &path)
+{
+    log_debug() << "Creating parent directories for " << path;
+    std::string parent = parentPath(path);
+    if (!isDirectory(parent) && parent != "") {
```
tobsan: I know this is just code that has been moved, but shouldn't we do && parent.empty() instead of comparing with empty string?

common/filetoolkitwithundo.cpp
```diff
@@ -0,0 +1,204 @@
...
+    char *dir = const_cast<char*>(templ.c_str());
+    dir = mkdtemp(dir);
+    if (dir == nullptr) {
+        log_warning() << "Failed to create buffered Directory: " << strerror(errno);
+        return nullptr;
+    }
+
+    m_cleanupHandlers.push_back(new DirectoryCleanUpHandler(templ));
+
+    return StringBuilder() << dir;
```
tobsan: How about just std::string(dir) instead? And superfluous blank line.

common/filetoolkitwithundo.cpp
```diff
@@ -0,0 +1,204 @@
...
+                    << " , dst=" << dst << " err=" << strerror(errno);
+        return ReturnCode::FAILURE;
+    }
+
+    return ReturnCode::SUCCESS;
+}
+
+ReturnCode FileToolkitWithUndo::createSharedMountPoint(const std::string &path)
+{
+    // MS_MGC_VAL |
```
tobsan: What is this contraption?

common/filetoolkitwithundo.cpp
```diff
@@ -0,0 +1,204 @@
...
+#include "filecleanuphandler.h"
+#include "mountcleanuphandler.h"
+
+namespace softwarecontainer {
+
+FileToolkitWithUndo::~FileToolkitWithUndo()
+{
+    bool success = true;
+    // Clean up all created directories, files, and mount points
+    for (auto it = m_cleanupHandlers.rbegin(); it != m_cleanupHandlers.rend(); ++it) {
```
tobsan: Is it possible to use for-each style loop here? In that case it would look nicer imho. But then we would have to clear() the vector afterwards.

tobsan (approved):
This is just copy/paste code, but we should really write documentation for this also.

common/filetoolkitwithundo.cpp
frznlogic: Done

common/filetoolkitwithundo.cpp
frznlogic: Done.

common/filetoolkitwithundo.cpp
frznlogic: Done.

common/filetoolkitwithundo.cpp
frznlogic: Would you be ok with 

for (auto it : m_cleanupHandlers) {
.... 
}

?

PR #43, opened 2016-09-27
Renamed SoftwareContainerLib and split Workspace

DunderRoffe:
SoftwareContainerLib has been renamed to SoftwareContainer since the class itself is not a library and the naming is inconsistent.
The source files libsoftwarecontainer.cpp and .h are now called softwarecontainer.cpp and .h.

SoftwareContainerWorkspace has been renamed to Workspace and been split into a separate .h and .cpp file.
getDefaultWorkspace was not used by anything but one test in softwarecontainerlib_unittest.cpp and has been causing problems
and has therefore been removed.

### DIFF
--- a/agent/src/softwarecontaineragent.cpp
+++ b/agent/src/softwarecontaineragent.cpp
@@ -9,7 +9,7 @@ SoftwareContainerAgent::SoftwareContainerAgent(
     , m_preloadCount(preloadCount)
     , m_shutdownContainers(shutdownContainers)
 {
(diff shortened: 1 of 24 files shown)


tobsan (changes requested):
Abstract/Function/Command-jobs have inconsistent naming of the SoftwareContainer pointer, and the Workspace class is lacking documentation. Apart from that, it looks good.

DunderRoffe:
Fixed

tobsan (approved):

PR #44, opened 2016-09-28
Fixed broken cleanup phase

DunderRoffe:


common/softwarecontainer-common.cpp
```diff
@@ -157,10 +157,6 @@ SignalConnectionsHandler::~SignalConnectionsHandler()
     }
 }
 
-bool operator&& (ReturnCode lhs, ReturnCode rhs) {
-    return lhs == rhs;
-}
```
tobsan, line 162: Why is this being removed?

common/softwarecontainer-common.h
```diff
@@ -61,8 +61,6 @@ enum class ReturnCode
     SUCCESS
 };
 
-bool operator&& (ReturnCode lhs, ReturnCode rhs);
```
tobsan, line 64: And why is this being removed?

tobsan (commented):
Looks good, just wondering about the ReturnCode stuff that is being removed.

common/softwarecontainer-common.cpp
DunderRoffe: This operator makes no sense.

``` cpp
ReturnCode::FAILURE && ReturnCode::FAILURE
```

does **not** equal

``` cpp
ReturnCode::FAILURE == ReturnCode::FAILURE
```

Use isError() or isSuccess() instead

common/softwarecontainer-common.h
DunderRoffe: See above

common/softwarecontainer-common.cpp
tobsan: OK!

PR #47, opened 2016-09-29
Updated service test framework and profiling tests

breakreturn:


servicetest/conftest.py
```diff
@@ -32,21 +31,32 @@
...
+    p = subprocess.Popen('dbus-launch', stdout=subprocess.PIPE, stderr=subprocess.STDOUT)
     for var in p.stdout:
         sp = var.split('=', 1)
-        os.environ[sp[0]] = sp[1][:-1]
+        env_var = sp[0]
+        env_val = sp[1][:-1]
+        os.environ[env_var] = env_val
+    yield
+    # The DBUS_SESSION_BUS_PID is set to sp[1] in the last loop iteration above...
+    os.kill(int(sp[1]), 9)
```
DunderRoffe: I think we should use the singal module instead of magic numbers for SIGKILL, SIGTERM etc...
https://docs.python.org/2/library/signal.html

servicetest/timingprofiling/test_profiling.py
```diff
@@ -17,307 +17,181 @@
...
+    log_file.seek(0)
+    for line in log_file:
+        if re.search(point_name, line):
             match = match + 1
-            if match == matchNumber:
-                return removeAnsi(string.split(line)[2])
-    print "Nothing found! " + pointName + " " + str(matchNumber)
+            if match == match_number:
+                return remove_ansi(string.split(line)[2])
+    print "Nothing found! " + point_name + " " + str(match_number)
```
DunderRoffe, line 110: Please use python3 compatible code

servicetest/dbus/test_dbus.py
```diff
@@ -23,7 +23,42 @@
...
+
+# This function is used by the 'agent' fixture to know where the log should be stored
+def logfile_path():
+    return CURRENT_DIR + "/test.log"
+
+
+# The path to where the test app is located, will be passed in the 'data' dict when
+# starting the container. The test app is assumed to be located in the
+# same directory as this test.
+HOST_PATH = os.path.dirname(os.path.abspath(__file__))
```
frznlogic, line 40: Can't you add this directly in the DATA section? Or is it used somewhere else also?

servicetest/timingprofiling/test_profiling.py
```diff
@@ -17,307 +17,181 @@
...
+    for line in log_file:
+        if re.search(function_name + " end", line):
             match = match + 1
-            if match == matchNumber:
+            if match == match_number:
                 print line
-                return removeAnsi(string.split(line)[6])
-    print "Nothing found! " + functionName + " " + str(matchNumber)
+                return remove_ansi(string.split(line)[6])
+    print "Nothing found! " + function_name + " " + str(match_number)
```
DunderRoffe, line 123: Please use python3 compatible code

servicetest/timingprofiling/test_profiling.py
```diff
@@ -17,307 +17,181 @@
...
+    log_file.seek(0)
+    for line in log_file:
+        if point_name in line:
             match = match + 1
-            if match == matchNumber:
-                return removeAnsi(string.split(line)[4])
-    print "Nothing found! " + pointName + " " + str(matchNumber)
+            if match == match_number:
+                return remove_ansi(string.split(line)[4])
+    print "Nothing found! " + point_name + " " + str(match_number)
```
DunderRoffe, line 141: Please use python3 compatible code

DunderRoffe (changes requested):
Overall good but some things needs to be fixed.

servicetest/conftest.py
```diff
@@ -32,21 +31,32 @@
...
-    agent_handler = SoftwareContainerAgentHandler(log_file_path)
+def agent(request):
+    """ Start the Agent and return an interface to it.
+
+        This fixture requires the test module to have a function named 'logfile_path'
+        on the module level
+    """
+    # Introspect the consuming module for the logfile path. The scope of this
+    # fixture means the module is received through the 'request' parameter
+    # and thus the path should be defined on the module level.
```
frznlogic, line 58: The mixture of comment style feels weird in this place, probably ok, but still sticks in my eyes.

servicetest/testframework/lib.py
```diff
@@ -21,153 +20,197 @@
...
+    """
+    def __init__(self):
+        threading.Thread.__init__(self)
+        GObject.threads_init()
+        self.__msg_queue = Queue.Queue()
+
+    def msg_queue(self):
+        return self.__msg_queue
+
+    def handler(self, gob, gob2, gob3):
```
frznlogic: These variable names should probably be renamed.... :D

frznlogic (approved):
Looks good to me, some minor comments.

DunderRoffe:
We should get rid of all time.sleep() calls.
I suggest using locks from the thread class. https://docs.python.org/2/library/thread.html

``` python
import thread

class Test (threading.Thread)

    def __init__(self):
         self.lock = thread.allocate_lock()
         self.lock.aquire()

    def run(self):
         # Do some setup
         self.lock.release()

    def wait_until_setup_is_done(self):
        a_lock.aquire()
        a_lock.release()


if __name__ == "__main__":
    t = Test()
    t.start()
    t.wait_until_setup_is_done()

    # Now all is setup and nice
```

servicetest/conftest.py
breakreturn: One comment is the method comment and the other one is in the body of the actual code. It's supposed to be different style, but I agree it's not very pretty when they happen to follow each other like this. But I think this is the correct way though.

servicetest/timingprofiling/test_profiling.py
breakreturn: We've never used any of our Python code with Python3 ever. Usually, that's not what's available and preferred in various contexts. Maybe that's changing, I'm not sure. I'm not sure I'd like to start checking our stuff for compatibility in general, e.g. use of bindings and whatnot.

servicetest/dbus/test_dbus.py
breakreturn: I could but I wanted to keep the variable separate so I could document it. If the comment is not needed, we can have it in the dict as it is not used elsewhere now. But I think this is more descriptive.

DunderRoffe (approved):

PR #49, opened 2016-09-30
Pulse fix

DunderRoffe:
## Fixed bug TestPulseaudioEnabled fail

The reason for the bug was that it took too long time before any stream were used
and as such pulseaudio exited due to being idle for too long. 
## Also

Checking that the source for bind mount exists on the host file system before attempting a bind mount


frznlogic (approved):
Looks good to me.

PR #50, opened 2016-09-30
Sync overlays

frznlogic:


DunderRoffe (approved):

.gitignore
```diff
@@ -14,3 +14,4 @@ build-*/
 .vagrant/
 vagrant_key
 vagrant_key.pub
+.cache/
```
breakreturn, line 15: Remove the cruft from .gitignore. I assume we don't need the key files anymore?

common/copycleanuphandler.h
```diff
@@ -0,0 +1,18 @@
+#ifndef COPYCLEANUPHANDLER_H
+#define COPYCLEANUPHANDLER_H
```
breakreturn: We've said we should use pragma once here.

frznlogic: I noticed it just now as well, is it ok if i add copyright headers all over at the same time?

common/copycleanuphandler.cpp
```diff
@@ -0,0 +1,34 @@
...
+
+CopyCleanupHandler::CopyCleanupHandler(std::string src, std::string dst)
+{
+    m_src = src;
+    m_dst = dst;
+}
+
+ReturnCode CopyCleanupHandler::clean()
+{
+    return RecursiveCopy::getInstance().copy(m_src, m_dst);
```
tobsan: It feels counter-intuitive (when no docs are given in the .h file) that the clean step copies something. I think the naming of this is a bit strange.

common/softwarecontainer-common.cpp
```diff
@@ -47,6 +47,23 @@ bool isDirectory(const std::string &path)
...
 }
 
+bool isDirectoryEmpty(const std::string &path) {
+  int n = 0;
+  struct dirent *d;
+  DIR *dir = opendir(path.c_str());
+  if (dir == NULL) //Not a directory or doesn't exist
+    return true;
+  while ((d = readdir(dir)) != NULL) {
+    if(++n > 2)
```
tobsan: Not really a nice check. Is there not a better way to exclude "." and ".." ?

common/unit-test/recursivecopy_unittest.cpp
```diff
@@ -0,0 +1,98 @@
...
+    ASSERT_TRUE(isFile(dstdir + "/lala2.txt"));
+}
+
+TEST_F(RecursiveCopyTest, copyDirFilesToFilesOverWrite) {
+    createFile(srcdir + "/lala.txt");
+    createFile(dstdir + "/lala.txt");
+    ASSERT_TRUE(RecursiveCopy::getInstance().copy(srcdir, dstdir) == ReturnCode::SUCCESS);
+    ASSERT_TRUE(isFile(srcdir + "/lala.txt"));
+    ASSERT_TRUE(isFile(dstdir + "/lala.txt"));
+
```
tobsan: The last two ASSERT_TRUE in this test will be true even if copy() failed (or returned success without succeeding). A better check would be to write different contents into the two "lala.txt" files and match after copy.

common/filetoolkitwithundo.cpp
```diff
@@ -161,6 +167,12 @@ ReturnCode FileToolkitWithUndo::overlayMount(
     return ReturnCode::SUCCESS;
 }
 
+ReturnCode FileToolkitWithUndo::syncOverlayMount(const std::string &lower,
+        const std::string &upper)
```
breakreturn, line 198: Indented too far?

common/recursivecopy.cpp
```diff
@@ -0,0 +1,70 @@
...
+ * and used by the RecursiveCopy::copy() function.
+ * @param src_path The path of the file to copy
+ * @param sb not used in this function
+ * @param typeflag Points to what type of node is traversed by ftw().
+ * @sideeffect Uses the global variables dst_root and src_root to know the root directories
+ * being copied to and from. These must not be changed while ftw is running or the function will
+ * misbehave.
+ * @return always 0 for now.
+ */
+int copy_file(const char* src_path, const struct stat* sb, int typeflag) {
```
breakreturn: Mixing naming conventions - we use camelStyle for methods

common/recursivecopy.cpp
```diff
@@ -0,0 +1,70 @@
...
+    return instance;
+}
+
+ReturnCode RecursiveCopy::copy(std::string src, std::string dst) {
+    ReturnCode retval = ReturnCode::SUCCESS;
+    m_copyLock.lock();
+    dst_root.assign(dst);
+    src_root.assign(src);
+    if (ftw(src.c_str(), copy_file, 20) != 0)
+    {
```
breakreturn: Bracket on the wrong line.

common/recursivecopy.cpp
breakreturn: bracket location

common/softwarecontainer-common.cpp
```diff
@@ -47,6 +47,23 @@ bool isDirectory(const std::string &path)
...
+      break;
+  }
+  closedir(dir);
+  if (n <= 2) //Directory Empty
```
breakreturn: Missing brackets

common/unit-test/recursivecopy_unittest.cpp
```diff
@@ -0,0 +1,98 @@
...
+#include <gmock/gmock.h>
+#include <gtest/gtest.h>
+#include <unistd.h>
+
+#include <filetoolkitwithundo.h>
+
+#include <recursivecopy.h>
+
+#include <fstream>
```
breakreturn, line 28: Is there any logic behind this grouping of includes?

tobsan (changes requested):
Looks good, my main concern is the copycleanuphandler naming.

common/filetoolkitwithundo.cpp
frznlogic: Hmm, too far? you want single indentation I take it?

servicetest/filesystem/test_filesystem.py
```diff
@@ -0,0 +1,88 @@
...
+    Container.CONFIG: '[{"enableWriteBuffer": false}]',
+    Container.BIND_MOUNT_DIR: "app",
+    Container.HOST_PATH: CURRENT_DIR
+}
+
+
+@pytest.mark.usefixtures("dbus_launch", "agent", "assert_no_proxy")
+class TestFileSystem(object):
+    """ This suite should do some basic testing of the Filesystem within the
+    containers and that the whole chain is working properly.
```
breakreturn: I suggest you indent this line to line up with the above.

common/copycleanuphandler.cpp
frznlogic: Would CopyHandler be a better name for it?

servicetest/filesystem/test_filesystem.py
```diff
@@ -0,0 +1,88 @@
...
+    """
+
+    def test_enable_write_buffer_flag(self):
```
breakreturn: The tests are very similar. Can they be written in the same way? In that case it would be more clear that we are testing the same things but with the flag toggled. This can then be parametrized and you could get rid of one of the methods completely.

common/softwarecontainer-common.cpp
frznlogic: There probably are, but this is fairly simple at least.. I could add a comment about it at least?

common/unit-test/recursivecopy_unittest.cpp
frznlogic: I agree, something could be improved. I'll have a look.

servicetest/filesystem/test_filesystem.py
```diff
@@ -0,0 +1,95 @@
...
+}
+
+
+@pytest.mark.usefixtures("dbus_launch", "agent", "assert_no_proxy")
+class TestFileSystem(object):
+    """ This suite should do some basic testing of the Filesystem within the
+        containers and that the whole chain is working properly.
+    """
+
+    def test_enable_write_buffer_flag(self):
```
breakreturn: Does this suite mean that we can remove the writeBuffer flag test from the dbus test suite?

frznlogic: Yes, consider it done. It was added temporarily there as there was no good way of breaking it out until your job was done.

servicetest/filesystem/test_filesystem.py
frznlogic: Done

servicetest/filesystem/test_filesystem.py
frznlogic: Done

common/unit-test/recursivecopy_unittest.cpp
frznlogic: Done

common/copycleanuphandler.cpp
frznlogic: Fixed

common/unit-test/recursivecopy_unittest.cpp
frznlogic: Nope. Fixed.

frznlogic: Sidenote, I usually add local includes top level, then project specific includes  and finally system level includes so the system level don't change failures in project level, and project level in local etc... Hope that makes sense.

common/softwarecontainer-common.cpp
frznlogic: Done.

common/recursivecopy.cpp
frznlogic: Done

common/recursivecopy.cpp
frznlogic: Done

common/recursivecopy.cpp
frznlogic: Done

common/copycleanuphandler.h
frznlogic: Done

common/filetoolkitwithundo.cpp
frznlogic: Done

.gitignore
frznlogic: Done

common/softwarecontainer-common.cpp
frznlogic: Done

common/copycleanuphandler.h
```diff
@@ -0,0 +1,50 @@
...
+ *
+ * For further information see LICENSE
+ */
+
+#pragma once
+
+#include <cleanuphandler.h>
+
+/**
+ * @brief The CopyCleanupHandler class is used to copy files on cleanup and can be added to the
```
tobsan: You mean the OverlaySyncCleanupHandler? ;)

tobsan (changes requested):
Good, only a comment that needs updating.

common/copycleanuphandler.h
frznlogic: WRONG!

PR #51, opened 2016-09-30
Removing old docs

DunderRoffe:
I went through the old_docs directory to see if it was worth keeping. I found that there is no introduction or text about "an eagle eye perspective" on the components that reside int the repo in the "new" docs. Although this was mentioned in the old_docks, that text can not be reused as it is quite sparse.

As such I think it the old_docs folder can be removed.


DunderRoffe:
Maybe we should add a story for writing the missing documentation?

tobsan (approved):

breakreturn:
Part of the story to remove this is to salvage anything we want to keep right? Bump the existing docs with what's useful and then we just remove it as you propose.

PR #54, opened 2016-10-03
Changed lxc-template path. No longer named 'hl'

DunderRoffe:
LXC_TEMPLATE_PATH redefined from "/var/hl/lxc-template" to "/var/softwarecontainer/lxc-templates"


tobsan (changes requested):
While we're at it, can we install it into something like /usr/share instead? /var should be for files that are subject to change (like logs). The template is static and arch-independent. If someone finds a better location, I'm all for that, though.

https://en.wikipedia.org/wiki/Filesystem_Hierarchy_Standard

tobsan (approved):

PR #58, opened 2016-10-04
Adding Jenkins file

frznlogic:


Jenkinsfile
```diff
@@ -0,0 +1,31 @@
+#!/usr/bin/groovy
+def workspace = "/mnt/ssd/jenkins/workspace/SoftwareContainerPipeline"
```
DunderRoffe: Seems awfully specific. Will this run on any of the nodes ?

tobsan: Can we really have this hard-coded? We should be able to capture the workspace of the first workspace (build stage) and then use that in runInVagrant. I don't know about groovy scripts, but in Jenkins shell scripts you can use ${WORKSPACE}

Jenkinsfile
```diff
@@ -0,0 +1,31 @@
...
+        "
+}
+
+node {
+    stage 'Download'
+    git branch: 'master', url: 'https://github.com/Pelagicore/softwarecontainer.git'
+    sh 'git submodule update --init'
+    
+    stage 'Build'
+    sh 'cd /mnt/ssd/jenkins/workspace/SoftwareContainerPipeline && vagrant up --provision'
```
tobsan: Also, if we want the workspace variable, it should be used here as well.

tobsan (changes requested):

Jenkinsfile
```diff
@@ -0,0 +1,31 @@
...
+
+// TBD: Haven't figured out how to make the parallel jobs run on the same slave/agent 
+// or if it is possible. Very annoying. Let's run sequentially in a single slave for now.
+//    parallel ('Documentation': {
+        stage 'Documentation'
+        runInVagrant('cd softwarecontainer/build && make doc')
+        
+//    }, 'UnitTest':{
+        stage 'UnitTest' 
+        runInVagrant("cd softwarecontainer/build && sudo ./run-tests.sh")
```
DunderRoffe: Are there not problems with running the tests in sudo ?

DunderRoffe (changes requested):

Jenkinsfile
DunderRoffe: The solution @tobsan suggested sounds good

frznlogic: There is no $WORKSPACE in pipeline scripts unfortunately..... :/ I did find a workaround however. pwd()....

Jenkinsfile
frznlogic: Yepp, that was the idea ;)

Jenkinsfile
frznlogic: Nope :)

Jenkinsfile
frznlogic: aaah, that explains why the pulseaudio test fails perhaps :)

DunderRoffe (approved):

PR #60, opened 2016-10-05
Add netlink code and replace networkgateway busybox calls.

tobsan:


common/netlink.h
```diff
@@ -0,0 +1,64 @@
+#include <linux/rtnetlink.h>
```
frznlogic: This should be documented more properly imho.

common/netlink.cpp
```diff
@@ -0,0 +1,468 @@
...
+    return 0;
+}
+
+/**
+ * To get a dump of all links/addresses/routes we only need to use the generic
+ * netlink family message type, and for that we only need to set family (AF_PACKET).
+ */
+bool Netlink::get_dump()
+{
+    m_links.clear();
```
frznlogic: indentation in this function...

common/netlink.cpp
```diff
@@ -0,0 +1,468 @@
...
+	size_t PAGE_SIZE = getpagesize();
+    char buf[PAGE_SIZE];
+
+    while (!end) {
+		int len;
+		struct nlmsghdr *msg;
+		struct msghdr reply;
+		memset(&reply, 0, sizeof(reply));
+
+        struct iovec io = { buf, PAGE_SIZE };
```
frznlogic: indentation.

common/netlink.cpp
```diff
@@ -0,0 +1,468 @@
...
+        if (hasBridge) {
+            break;
+        }
+    }
+
+    if (!hasBridge) {
+        return false;
+    }
+
+	bool bridgeHasGateway = false;
```
frznlogic: indentation

common/netlink.cpp
```diff
@@ -0,0 +1,468 @@
...
+        // Second, set IP address
+        // TODO: Support for ipv6
+		netlink_request<ifaddrmsg> msg_setip = alloc_msg<ifaddrmsg>(RTM_NEWADDR, NLM_F_CREATE | NLM_F_REPLACE);
+        msg_setip.pay.ifa_family = AF_INET; // ipv4
+        msg_setip.pay.ifa_prefixlen = 24; // netmask TODO: parse from argument netmask
+        msg_setip.pay.ifa_scope = RT_SCOPE_UNIVERSE;
+        msg_setip.pay.ifa_index = ifinfo.ifi_index; // interface (link)
+
+		struct in_addr bcast_addr; // TODO: calculate from ip
+		if (inet_pton(msg_setip.pay.ifa_family, m_broadcast.c_str(), &bcast_addr) == 0) {
```
frznlogic: indentation.

common/netlink.cpp
```diff
@@ -0,0 +1,468 @@
...
+    set_gw.pay.rtm_table = RT_TABLE_MAIN;
+    set_gw.pay.rtm_protocol = RTPROT_STATIC;
+    set_gw.pay.rtm_scope = RT_SCOPE_UNIVERSE;
+    set_gw.pay.rtm_type = RTN_UNICAST;
+
+    struct in_addr gw_addr;
+    if (inet_aton(gateway_address, &gw_addr) == 0) {
+        return false;
+    }
+	add_attribute(set_gw, RTA_GATEWAY, sizeof(gw_addr), &gw_addr);
```
frznlogic: indentation.

common/netlink.cpp
```diff
@@ -0,0 +1,468 @@
...
+
+#include <linux/if_arp.h> // ARPHRD_ defines
+
+#include <unistd.h> // getpid(), getpagesize()
+#include <string.h> // memcpy, memset etc
+
+Netlink::Netlink()
+{
+    m_sequence_number = 1;
+    m_ip = "10.0.3.100";
```
frznlogic: These hardcoded IP addresses are really scary imho. Can we do it some other way please?

common/netlink.cpp
```diff
@@ -0,0 +1,468 @@
...
+    // Setup kernel sockaddr
+    memset(&m_kernel, 0, sizeof(m_kernel));
+    m_kernel.nl_family = AF_NETLINK; /*  fill-in kernel address (destination) */
+
+    return true;
+}
+
+template<typename payload> Netlink::netlink_request<payload> Netlink::alloc_msg(const int type, const int flags)
+{
+	netlink_request<payload> request;
```
frznlogic: indentation.

common/netlink.cpp
```diff
@@ -0,0 +1,468 @@
...
+	memset(request.attr, 0, sizeof(request.attr));
+
+	return request;
+}
+
+template<typename payload>
+void Netlink::add_attribute(netlink_request<payload> &req, int type, size_t length, void *data)
+{
+	// First, get a pointer to memory at the end of the payload (pointed out by nlmsg_len)
+	struct rtattr *rta = (struct rtattr *)(((char *) &req) + NLMSG_ALIGN(req.hdr.nlmsg_len));
```
frznlogic: Can this be simplified somehow? It looks really complex because of the typecasts etc, maybe break it down into parts making it simpler to read?

common/netlink.cpp
```diff
@@ -0,0 +1,468 @@
+#include "netlink.h"
+
+#include <linux/if_arp.h> // ARPHRD_ defines
+
+#include <unistd.h> // getpid(), getpagesize()
+#include <string.h> // memcpy, memset etc
+
+Netlink::Netlink()
```
frznlogic: What do you think, should we name the class NetworkManager and the CapabilityManager?

common/netlink.h
tobsan: You mean, in general? I agree! Or do you mean just the rtnetlink header?

common/netlink.cpp
```diff
@@ -0,0 +1,468 @@
...
+    }
+}
+
+Netlink::~Netlink()
+{
+}
+
+bool Netlink::setup_netlink()
+{
+    if ((m_fd = socket(AF_NETLINK, SOCK_RAW, NETLINK_ROUTE)) == -1) {
```
frznlogic: What happens if we run setup_netlink multiple times by accident? Do we need some kind of state?

common/netlink.cpp
```diff
@@ -0,0 +1,468 @@
...
+                        break;
+                    }
+					case NLMSG_DONE:
+						end = true;
+						break;
+                    case RTM_GETLINK:
+					case RTM_NEWLINK:
+                        save_msg<ifinfomsg, LinkInfo>(msg, m_links);
+                        break;
+                    case RTM_NEWADDR:
```
frznlogic: All the breaks feel odd, gather them all up at the end with a default handler perhaps?

frznlogic (commented):
In general good, some indentation and other comments. I think more documentation of the netlink.h would be really good, possibly some state/state checking of the connection, etc...?

common/netlink.h
frznlogic: The whole file :D

common/netlink.cpp
```diff
@@ -0,0 +1,468 @@
...
+                char *ifname = (char *) data;
+                if (strcmp(ifname, bridgeName) == 0) {
+                    hasBridge = true;
+                    bridge_ifindex = linkinfo.first.ifi_index;
+                    break;
+                }
+            }
+        }
+
+        // TODO: Make this nicer, please
```
DunderRoffe: please do

DunderRoffe (changes requested):
Should not be put in common as it is not used by agent. Create a separate directory in softwarecontainer for it

common/netlink.h
tobsan: Gotcha! Good idea!

common/netlink.cpp
tobsan: Maybe NetlinkManager? But on the other hand, the underlying implementation should not be that interesting to whoever uses this, as long as one can use them both inside and outside containers.

NetworkManager could work, but it has a very GNOMEish ring to it ;)

common/netlink.cpp
tobsan: Already done!

common/netlink.cpp
tobsan: Yes, we do need that, thanks for pointing it out!

common/netlink.cpp
tobsan: That is indeed possible.

common/netlink.cpp
tobsan: Will do!

libsoftwarecontainer/src/netlink.cpp
```diff
@@ -0,0 +1,573 @@
...
+
+        if (bridgeHasGateway) {
+            break;
+        }
+    }
+
+    return bridgeHasGateway;
+}
+
+int Netlink::readMessage()
```
DunderRoffe, line 348: This method should maybe be split into multiple parts. E.g. One method dealing with link, one method dealing with routes ....  Or maybe it could be solved with some nice template magic ? Thinking mostly of the switch statement.

libsoftwarecontainer/src/netlink.cpp
```diff
@@ -0,0 +1,658 @@
...
+        
+        if (isError(sendMessage(down_msg))) {
+            return ReturnCode::FAILURE;
+        } else {
+            return ReturnCode::SUCCESS;
+        }
+    }
+
+    return ReturnCode::FAILURE;
+}
```
DunderRoffe, line 279: This code does not do what it is supposed to

DunderRoffe (changes requested):

tobsan:
I just updated this PR, please have a look again.

libsoftwarecontainer/include/netlink.h
```diff
@@ -0,0 +1,301 @@
...
+
+#include "softwarecontainer-common.h"
+
+#include <linux/rtnetlink.h>
+#include <arpa/inet.h>
+
+#include <string>
+#include <vector>
+
+/*
```
frznlogic: /*\* to mark as doxygen documentation.

libsoftwarecontainer/include/netlink.h
```diff
@@ -0,0 +1,301 @@
...
+         */
+        virtual ReturnCode setupNetlink();
+
+        /*
+         * @brief clears the cache
+         *
+         * This clears all the links, addresses and routes. It also has the responsibility to free
+         * all allocated attributes for links, addresses och routes.
+         */
+        void clearCache();
```
frznlogic, line 178: Where is the cache created? It might be worth adding a link to it or some note about it.

libsoftwarecontainer/include/netlink.h
```diff
@@ -0,0 +1,301 @@
...
+         * more reasonable.
+         *
+         * @note this function allocates memory on the heap. Use freeAttributes to free the list
+         *
+         * @param h the netlink message header
+         * @param result a reference to the list in which to store results
+         * @result ReturnCode::SUCCESS on success, ReturnCode::FAILURE otherwise
+         */
+        template<typename msgtype>
+            ReturnCode getAttributes(const struct nlmsghdr *header, AttributeList &result);
```
DunderRoffe: Is it possible to use a referance instead of a pointer here ?

libsoftwarecontainer/src/gateway/networkgateway.cpp
```diff
@@ -262,64 +261,91 @@ bool NetworkGateway::generateIP()
...
-    return true;
+    return isSuccess(ret);
 }
 
 bool NetworkGateway::up()
 {
-    log_debug() << "Attempting to configure eth0 to 'up state'";
-    std::string cmd;
-
     if (!m_interfaceInitialized) {
```
DunderRoffe: You could negate this and return true inside to lower indentation

libsoftwarecontainer/src/netlink.cpp
```diff
@@ -0,0 +1,559 @@
...
+            if (attr.rta_type != IFA_ADDRESS && attr.rta_type != IFA_LOCAL) {
+                continue;
+            }
+
+            void *data = attrPair.second;
+            char out[INET6_ADDRSTRLEN];
+
+            // It is ok here for inet_ntop to fail (data could be bad)
+            if (inet_ntop(addressFamily, data, out, sizeof(out))) {
+                if (strcmp(out, needle) == 0) {
```
DunderRoffe: if (a) if (b) ... == if (a && b)
Maybe it could be rewritten to increase readability ?

libsoftwarecontainer/src/netlink.cpp
```diff
@@ -0,0 +1,559 @@
...
+        memset(&reply, 0, sizeof(reply));
+
+        struct iovec io = { buf, sizeof(buf) };
+        reply.msg_iov = &io;
+        reply.msg_iovlen = 1;
+        reply.msg_name = &m_kernel;
+        reply.msg_namelen = sizeof(m_kernel);
+
+        len = recvmsg(m_fd, &reply, 0); /* read lots of data */
+        if (len > 0) {
```
DunderRoffe: Could negate and return to lower indentation

DunderRoffe (commented):
I guess it is fine either way but here are some suggestions

libsoftwarecontainer/include/netlink.h
tobsan: Yes, I will update it to use that.

libsoftwarecontainer/src/gateway/networkgateway.cpp
tobsan: Good point, will do!

libsoftwarecontainer/src/netlink.cpp
tobsan: Good point, will do!

libsoftwarecontainer/src/netlink.cpp
tobsan: Done!

DunderRoffe (approved):

DunderRoffe (approved):

PR #64, opened 2016-10-06
Add artifact archivation

frznlogic:
Artifacts should be archived for this to be approved.


tobsan (commented):
Looks good, but fails to build, apparently.

PR #65, opened 2016-10-07
Added user docs for service manifests

breakreturn:


doc/chapters/service-manifests/index.rst
```diff
@@ -0,0 +1,65 @@
...
+Manifest format
+===============
+
+A service manifest is a text file containing JSON. It is required to define
+a version set to "1", and an array "capabilities" which contains a
+"name" key with the capability name as value, and a "gateways" key which
+contains an array of objects each defining a gateway configuration identified
+with a gateway ID.
+
+Examples
```
rpannek, line 44: I'm confused about where this should go in, into a special file somewhere or what?

doc/chapters/service-manifests/index.rst
```diff
@@ -0,0 +1,65 @@
...
+For example, a platform might have a capability `com.acme.SomeResource`
+defined. This capability requires access to service `A`. Service `A`
+will then need to deploy a service manifest which contains all gateway
+configurations needed for the `com.acme.SomeResource` capability.
+
+A service manifest might also contain gateway configurations for more
+capabilities.
+
+More services than one might be involved to realize a capability and
+then each relevant service would need to be integrated in this way.
```
thenor: "More than one service might be needed to fulfill a capability, meaning that each relevant service would need to be integrated in this way."

doc/chapters/service-manifests/index.rst
```diff
@@ -0,0 +1,65 @@
...
+
+For information about how to work with the manifest files, e.g. where to
+install them etc., see :ref:`Integration <integration>`.
```
thenor: I feel this phrase is a bit confusing since the next chapter is called "Manifest format". Maybe move this text to the next chapter and add this text as a "disclaimer". E.g. "this chapter describes the format of the Manifest, for information on where to place them and hwo to integrate them, please refer to :ref:`Integration <integration>`.

doc/chapters/service-manifests/index.rst
breakreturn: Will use your sentence instead.

doc/chapters/service-manifests/index.rst
breakreturn: The "Manifest format" is a sub-heading to this text. My intention is to describe what the reader should expect to find in the rest of the chapter. I will add a note on that this chapter is for the format of the file, and then the reference to how to use it might make more sense.

doc/chapters/service-manifests/index.rst
thenor: An addition to this comment: I am confused by the "this way". Since this is a stand-alone paragraph for me it shouldn't refer to a "this". The context for the paragraph is only the paragraph. :P

doc/chapters/service-manifests/index.rst
breakreturn: I will try to put a more clear explanation in the beginning what the reader can expect to find in this text. I'd like to have integration guidelines separate.

thenor (approved):

PR #69, opened 2016-10-11
WIP: Reading capabilities from disk

thenor:
Parses capability from Service Manifest files and returns GatewayConfigurations.

### DIFF
--- a/agent/src/capability/configstore.cpp
+++ b/agent/src/capability/configstore.cpp
@@ -1,0 +1,174 @@
+/*
+ * Copyright (C) 2016 Pelagicore AB
+ *
+ * Permission to use, copy, modify, and/or distribute this software for
+ * any purpose with or without fee is hereby granted, provided that the
+ * above copyright notice and this permission notice appear in all copies.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL
+ * WARRANTIES WITH REGARD TO THIS SOFTWARE INCLUDING ALL IMPLIED
+ * WARRANTIES OF MERCHANTABILITY AND FITNESS. IN NO EVENT SHALL THE AUTHOR
+ * BE LIABLE FOR ANY SPECIAL, DIRECT, INDIRECT, OR CONSEQUENTIAL DAMAGES
+ * OR ANY DAMAGES WHATSOEVER RESULTING FROM LOSS OF USE, DATA OR PROFITS,
+ * WHETHER IN AN ACTION OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION,
+ * ARISING OUT OF OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS
+ * SOFTWARE.
+ *
+ * For further information see LICENSE
+ */
+
+#include "configstore.h"
+
+
+ConfigStore::ConfigStore()
+{
+  
+}
+
+ConfigStore::~ConfigStore()
+{
+}
+
+std::vector<GatewayConfiguration> softwarecontainer::capFromFile(std::string capID,
+								 std::string &filePath)
+{
+    std::vector<GatewayConfiguration> GWconfs;
+
+    if (!configFilePath.empty()) {
+        GWconfs = getGatewaysByCap(capID, filePath);
+    }
+
+    return GWconfs;
+}
+/*
+std::vector<Capability> ConfigParser::capsFromConfigs(std::string &configDirPath)
+{
+    if (configDirPath.empty()) {
+        log_debug() << "Path to configuration files is empty";
+    }
+
+    if (!hasEnding(configDirPath, "/")) {
+        configDirPath += "/";
+    }
+
+    std::vector<Capability> caps;
+    std::vector<std::string> configFiles;
+
+    configFiles = configFileList(configDirPath);
+
+    if (configFiles.empty()) {
+        log_debug() << "No configuration files found";
+    }
+
+    for (unsigned int i = 0; i != configFiles.size(); i++) {
+        std::string configFilePath = configDirPath + configFiles[i];
+
+        std::vector<Capability> configCaps = capsFromConfig(configFilePath);
+
+        caps.insert(caps.end(), configCaps.begin(), configCaps.end());
+    }
+
+    return caps;
+}
+*/
+
+std::vector<GatewayConfiguration> softwarecontainer::getGatewaysByCap(std::string capID,
+								      const std::string &filePath) {
+    std::vector<GatewayConfiguration> gateways;
+
+    JSonElement root = JSonElement::fromFile(filePath);
+
+    if (root.isValid()) {
+
+        auto arrayElement = root.getField("capabilities");
+
+        if (arrayElement.isArray()) {
+            std::vector<JSonElement> children;
+            arrayElement.readChildren(children);
+            for (auto& child : children) {
+
+                std::string capName;
+                child.read("name", capName);
+
+		// TODO (thenor): Only push gateways for capID
+		if (capName == capID) {
+		    auto gatewaysElement = child.getField("gateways");
+		    if (arrayElement.isArray()) {
+
+		        std::string strError;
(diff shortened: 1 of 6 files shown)


agent/src/capability/configstore.cpp
```diff
@@ -0,0 +1,173 @@
...
+ConfigStore::ConfigStore()
+{
+}
+
+ConfigStore::~ConfigStore()
+{
+}
+
+    static std::vector<GatewayConfiguration> capFromFile(std::string capID, std::string &configFilePath);
+    static std::vector<GatewayConfiguration> capFromDir(std::string capID, std::string &configDirPath);
```
tobsan: Indentation error?

tobsan (commented):
From the looks of this it looks like we'll be reading in capabilities from file on demand. Am I reading it correct then? If so, that should be changed to reading all of them at once and then do a lookup in some data structure on demand.

agent/src/capability/configstore.cpp
thenor: Copy paste error from the h file. I saw this myself. I also saw that my cpp and h files are not aligned at all. Sorry about that.

agent/src/capability/configstore.h
```diff
@@ -0,0 +1,98 @@
...
+    /**
+     * @brief Returns all Gateways for a certain Capability.
+     */
+    static std::vector<GatewayConfiguration> getGatewaysByCap(std::string capID, std::vector<Capability> caps);
+
+    /**
+     * @brief Populates a vector with Gateway objects and returns true on
+     * 	success and false on failure.
+     */
+    static bool parseGatewayData(const pelagicore::JSonElement &gatewayData, std::vector<Gateway> &gateways, std::string *error);
```
thenor: "const json_t *element" ?? Not "const pelagicore::JSonElement &gatewayData" ??

agent/src/capability/configstore.cpp
```diff
@@ -0,0 +1,174 @@
...
+{
+  
+}
+
+ConfigStore::~ConfigStore()
+{
+}
+
+std::vector<GatewayConfiguration> softwarecontainer::capFromFile(std::string capID,
+								 std::string &filePath)
```
DunderRoffe, line 33: fix indentaion

agent/src/capability/configstore.cpp
```diff
@@ -0,0 +1,174 @@
...
+
+        caps.insert(caps.end(), configCaps.begin(), configCaps.end());
+    }
+
+    return caps;
+}
+*/
+
+std::vector<GatewayConfiguration> softwarecontainer::getGatewaysByCap(std::string capID,
+								      const std::string &filePath) {
```
DunderRoffe, line 76: align `const std::string &filePath` with `std::string capID`

agent/src/capability/configstore.cpp
DunderRoffe: Since `filePath` is const in getGatewayCap(...), maybe `filePath` could be const here as well?

DunderRoffe (commented):
Seems good overall, but the indentation is a bit off.

agent/src/capability/configstore.cpp
DunderRoffe: This applies generally in the file

agent/src/capability/configstore.h
```diff
@@ -0,0 +1,94 @@
...
+//#include "SoftwareContainerAgent_dbuscpp_adaptor.h"
+#include "softwarecontainer-common.h"
+
+#include <jsonparser.h>
+#include "commandjob.h"
+
+
+namespace softwarecontainer {
+
+class ConfigStore
```
frznlogic, line 46: I'd love some documentation on what ConfigStore is.

agent/src/capability/configstore.h
```diff
@@ -0,0 +1,94 @@
...
+     *                                                                                                                          
+     */
+    static std::vector<GatewayConfiguration> capFromDir(std::string capID, std::string &configDirPath);
+
+private:
+
+    /**
+     * @brief Returns all Gateways for a certain Capability.
+     */
+    static std::vector<GatewayConfiguration> getGatewaysByCap(std::string capID, const std::string &filePath);
```
frznlogic, line 80: Some of these declarations wind up really long, not sure if what we say about using namespace, and also maybe declare a new datatype for vector<GatewayConfiguration>. Just a thought.

agent/src/capability/configstore.cpp
frznlogic, line 76: the { of functions are supposed to be on a new line?

agent/src/capability/configstore.cpp
```diff
@@ -0,0 +1,174 @@
...
+		    auto gatewaysElement = child.getField("gateways");
+		    if (arrayElement.isArray()) {
+
+		        std::string strError;
+			if (!parseGatewayData(gatewaysElement, gateways, &strError)) {
+			  log_debug() << "Error in " << pathString << ": " << strError;
+			  gateways.clear();
+			}
+			else {
+			  GatewayConfiguration gw = GatewayConfiguration(gatewayName, gwConf);
```
frznlogic, line 104: There's a lot of nesting going on here... can the function be broken down?

agent/src/capability/configstore.cpp
```diff
@@ -0,0 +1,174 @@
...
+            if (pelagicore::has_suffix(filename, ".json"))
+                configFiles.push_back(filename);
+        }
+        closedir(dirFile);
+    }
+
+    return configFiles;
+}
+
+bool softwarecontainer::hasEnding(std::string const &fullString, std::string const &ending)
```
frznlogic, line 167: These  functions are declared in the softwarecontainer::ConfigStore in the header file, but here they are in softwarecontaner:: only?

agent/src/capability/configstore.cpp
```diff
@@ -0,0 +1,174 @@
...
+{
+    std::vector<GatewayConfiguration> GWconfs;
+
+    if (!configFilePath.empty()) {
+        GWconfs = getGatewaysByCap(capID, filePath);
+    }
+
+    return GWconfs;
+}
+/*
```
frznlogic, line 43: Why is this commented out?

agent/src/capability/configstore.cpp
```diff
@@ -0,0 +1,174 @@
...
+
+    DIR *dirFile = opendir(configPath.c_str());
+    if (dirFile == nullptr) {
+        log_debug() << "Could not open directory: " << configPath << "\n";
+    } else {
+        struct dirent *hFile;
+        while ((hFile = readdir(dirFile)) != nullptr) {
+            std::string filename(hFile->d_name);
+
+            if (pelagicore::has_suffix(filename, ".json"))
```
frznlogic, line 158: No if without {}

agent/src/capability/configstore.cpp
```diff
@@ -0,0 +1,174 @@
...
+            for (auto& child : children) {
+
+                std::string capName;
+                child.read("name", capName);
+
+		// TODO (thenor): Only push gateways for capID
+		if (capName == capID) {
+		    auto gatewaysElement = child.getField("gateways");
+		    if (arrayElement.isArray()) {
+
```
frznlogic, line 97: Too many empty lines.

agent/src/capability/configstore.cpp
```diff
@@ -0,0 +1,174 @@
...
+		        std::string strError;
+			if (!parseGatewayData(gatewaysElement, gateways, &strError)) {
+			  log_debug() << "Error in " << pathString << ": " << strError;
```
frznlogic, line 100: Indentation in general is wrong. (2 spaces only here)

frznlogic (changes requested):
Looks like you are started. Some minor comments here and there. I think the getGatewaysByCap could use some breaking apart, possibly some generalization as well. Perhaps something in the jsonparser?

thenor:
Thanks for all the comments. I will go through them, but some might be inapplicable in the new pull request. Stay tuned!

PR #70, opened 2016-10-12
Remove -P option which isn't available in all versions of busybox

frznlogic:


DunderRoffe (approved):

tobsan (approved):
Good, but it is an xargs option, not busybox

PR #71, opened 2016-10-12
CreateContainer ID pool for creating better id mechanism

kursatkobya:
Change container of m_containers from vector to map
Install proper id pool mechanism

### DIFF
--- a/agent/src/softwarecontaineragent.cpp
+++ b/agent/src/softwarecontaineragent.cpp
@@ -2,18 +2,18 @@
 #include <cstdint>
 
 SoftwareContainerAgent::SoftwareContainerAgent(
-		Glib::RefPtr<Glib::MainContext> mainLoopContext
-		, int preloadCount
-		, bool shutdownContainers
-		, int shutdownTimeout)
+        Glib::RefPtr<Glib::MainContext> mainLoopContext
+        , int preloadCount
+        , bool shutdownContainers
+        , int shutdownTimeout)
 : m_mainLoopContext(mainLoopContext)
 , m_preloadCount(preloadCount)
 , m_shutdownContainers(shutdownContainers)
 {
-	m_containerIdPool.push_back(0);
-	m_softwarecontainerWorkspace = std::make_shared<Workspace>();
-	triggerPreload();
-	m_softwarecontainerWorkspace->m_containerShutdownTimeout = shutdownTimeout;
+    m_containerIdPool.push_back(0);
+    m_softwarecontainerWorkspace = std::make_shared<Workspace>();
+    triggerPreload();
+    m_softwarecontainerWorkspace->m_containerShutdownTimeout = shutdownTimeout;
 }
 
 SoftwareContainerAgent::~SoftwareContainerAgent()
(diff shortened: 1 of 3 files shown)


agent/src/softwarecontaineragent.cpp
```diff
@@ -106,29 +116,35 @@ bool SoftwareContainerAgent::parseConfig(const std::string &config)
...
+
+	SoftwareContainer *container;
+	if (m_preloadedContainers.size() != 0) {
+		container = m_preloadedContainers[0].release();
+		m_preloadedContainers.erase(m_preloadedContainers.begin());
+	} else {
+		container = new SoftwareContainer(m_softwarecontainerWorkspace);
+	}
+
+	ContainerID availableID = m_containerIdPool.back();
```
breakreturn: How about having the ID generation as a separate entity, e.g. a class or method so it can be tested separately?

agent/src/softwarecontaineragent.cpp
```diff
@@ -31,24 +33,32 @@ void SoftwareContainerAgent::triggerPreload()
 
 void SoftwareContainerAgent::deleteContainer(ContainerID containerID)
 {
-    bool valid = ((containerID < m_containers.size()) && (m_containers[containerID] != nullptr));
-    if (valid) {
-        m_containers[containerID] = nullptr;
-    } else {
-        log_error() << "Invalid container ID " << containerID;
-    }
+	bool valid = ((containerID < UINT32_MAX )
```
breakreturn: Put the two identical checks for validity in a private method, or inside the "ID generator class" I proposed below, if that works out.

agent/unit-test/softwarecontaineragent_unittest.cpp
```diff
@@ -53,12 +53,33 @@ class SoftwareContainerAgentTest: public ::testing::Test
...
+	bool retval = sca->checkContainer(id, container);
+	ASSERT_TRUE(retval == false);
 }
+/*
+ * Stress test written for creation + deletion
+ * It takes too much time, thus commented out
+ *
+TEST_F(SoftwareContainerAgentTest, StressCreateDeleteContainer) {
+	ContainerID id = 0;
+	for (auto i = 0 ; i < UINT32_MAX; i++) {
```
frznlogic: I don't think stress tests should be in the unit tests, what do @breakreturn think? 

If we keep the test here, we should be able to  just lower how many times we run it to like ... 2-3000 iterations or something that takes a reasonable amount of time.

agent/src/softwarecontaineragent.cpp
```diff
@@ -106,29 +116,35 @@ bool SoftwareContainerAgent::parseConfig(const std::string &config)
...
-
-    return id;
+	profilepoint("createContainerStart");
+	profilefunction("createContainerFunction");
+
+	parseConfig(config);
+
+	SoftwareContainer *container;
+	if (m_preloadedContainers.size() != 0) {
+		container = m_preloadedContainers[0].release();
```
frznlogic: Indentation feels a bit off, 8 instead of 4 spaces?

agent/src/softwarecontaineragent.cpp
kursatkobya: Sure, It makes lots of sense to make an inline function to check valid id. I am on it

agent/src/softwarecontaineragent.cpp
kursatkobya: Finding suitable id can be a function you are right. But testing only id pool won't have more meaning testing std::vectors. 
Please see the changes

agent/unit-test/softwarecontaineragent_unittest.cpp
kursatkobya: It was just commented out, i can remove it entirely if needed,

frznlogic: I think the test could be lowered in iterations and still be valid, but I'd also like to hear what Joakim says about it first.

agent/src/softwarecontaineragent.cpp
```diff
@@ -20,225 +22,245 @@ SoftwareContainerAgent::~SoftwareContainerAgent()
...
-                return false;
-            }
-        }
-    } else {
-        log_error() << "Root JSON element is not an array";
-        return false;
-    }
-
-    return true;
+	if (config.size() == 0) {
```
breakreturn: This chunk seems to be indented wrong. Please review the whole commit according to the style guidelines.

agent/unit-test/softwarecontaineragent_unittest.cpp
breakreturn: If we have stress tests that would require a lot of time or some special treatment I think we should put them in separate suites and make it possible to deselect them easily so the CI system essentially can choose to run them and developers can choose to skip. W.r.t. unit-tests or not, these types of tests will IMHO fall into the third, slightly undefined, category between proper unit-tests and the service-tests. In any case, are stress testing the creation of container related to container ID? This should perhaps be done with a different scope.

agent/src/softwarecontaineragent.cpp
kursatkobya: Sorry for inconvenience, indentation is fixed. Please check again.

agent/src/softwarecontaineragent.cpp
```diff
@@ -1,14 +1,16 @@
...
 
 SoftwareContainerAgent::SoftwareContainerAgent(
         Glib::RefPtr<Glib::MainContext> mainLoopContext
         , int preloadCount
         , bool shutdownContainers
         , int shutdownTimeout)
-    : m_mainLoopContext(mainLoopContext)
-    , m_preloadCount(preloadCount)
-    , m_shutdownContainers(shutdownContainers)
+: m_mainLoopContext(mainLoopContext)
```
frznlogic, line 9: I actually think this should be indented, see the style guide please ;).

agent/src/softwarecontaineragent.cpp
kursatkobya: Sorry for inconvenience, indentation is fixed. Please check again.

agent/src/softwarecontaineragent.cpp
```diff
@@ -29,26 +31,33 @@ void SoftwareContainerAgent::triggerPreload()
     }
 }
 
+inline bool SoftwareContainerAgent::isIdValid (ContainerID containerID)
+{
+    return ((containerID < UINT32_MAX )
```
frznlogic: Final space between UINT32_MAX and )

agent/src/softwarecontaineragent.cpp
```diff
@@ -29,26 +31,33 @@ void SoftwareContainerAgent::triggerPreload()
...
+            && (containerID >= 0 )
```
frznlogic: Here also

agent/unit-test/softwarecontaineragent_unittest.cpp
kursatkobya: I agree, so it is already commented out in the softwarecontaineragent_unittest.cpp just for notice.
The aim of stress test was actually to observe correctness of creation/deletion sequence.
I can remove it if you feel it is bettter way.

agent/unit-test/softwarecontaineragent_unittest.cpp
```diff
@@ -37,28 +37,49 @@ class SoftwareContainerAgentTest: public ::testing::Test
...
 
     void SetUp() override
-    {
+            {
         sca = new SoftwareContainerAgent(
-                      m_context
-                    , m_preloadCount
-                    , m_shutdownContainers
-                    , m_shutdownTimeout);
+                m_context
```
frznlogic: I'd move the , to the end of the lines as well, we changed that style a few weeks ago as it makes more sense.

agent/unit-test/softwarecontaineragent_unittest.cpp
```diff
@@ -37,28 +37,49 @@ class SoftwareContainerAgentTest: public ::testing::Test
     std::shared_ptr<Workspace> workspace;
 
     void SetUp() override
-    {
+            {
```
frznlogic: Why is this indented so far?

agent/unit-test/softwarecontaineragent_unittest.cpp
frznlogic: So basically one more test suite which can be run on CI machines with component tests?

breakreturn: Yes, another suite sounds easiest. What exactly is the easiest way to select/deselect and organize tests depends on the framework used I'd say, but both in the case of gtest and py.test I think it would be "suites". Let's move stress tests to a different separate update perhaps?

agent/src/softwarecontaineragent.cpp
kursatkobya: I don't believe there is a problem with this. parameters are not in the same indentation level with initializer lists. And i have checked with style guide. I couldn't find any contradiction .

Please guide me if you feel this is wrong.

agent/unit-test/softwarecontaineragent_unittest.cpp
kursatkobya: I did not write this part of the code but i have seen C++ formatting guidelines was suggesting to use ',' at the start .

agent/unit-test/softwarecontaineragent_unittest.cpp
kursatkobya: So i removed commented out lines of stress test.

agent/src/softwarecontaineragent.cpp
kursatkobya: Fixed

agent/src/softwarecontaineragent.cpp
kursatkobya: Fixed

agent/src/softwarecontaineragent.cpp
```diff
@@ -29,26 +31,33 @@ void SoftwareContainerAgent::triggerPreload()
     }
 }
 
+inline bool SoftwareContainerAgent::isIdValid (ContainerID containerID)
+{
+    return ((containerID < UINT32_MAX)
+            && (containerID >= 0)
+            && (m_containers.find(containerID)!= m_containers.end()));
```
tobsan: Can we use m_container.count(containerID) instead?

tobsan (approved):

agent/src/softwarecontaineragent.cpp
kursatkobya: You are right and done

PR #77, opened 2016-10-14
WIP: Reading capabilities from disk

thenor:
Part 1 of reading from file:
Parse capabilities from Service Manifest files and save them internally.
Can read from file or from directory.

Note: Does not include tests.

### DIFF
--- a/agent/src/capability/capability.cpp
+++ b/agent/src/capability/capability.cpp
@@ -1,0 +1,40 @@
+
+/*
+ * Copyright (C) 2016 Pelagicore AB
+ *
+ * Permission to use, copy, modify, and/or distribute this software for
+ * any purpose with or without fee is hereby granted, provided that the
+ * above copyright notice and this permissiobilityn notice appear in all copies.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL
+ * WARRANTIES WITH REGARD TO THIS SOFTWARE INCLUDING ALL IMPLIED
+ * WARRANTIES OF MERCHANTABILITY AND FITNESS. IN NO EVENT SHALL THE AUTHOR
+ * BE LIABLE FOR ANY SPECIAL, DIRECT, INDIRECT, OR CONSEQUENTIAL DAMAGES
+ * OR ANY DAMAGES WHATSOEVER RESULTING FROM LOSS OF USE, DATA OR PROFITS,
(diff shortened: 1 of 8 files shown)


agent/src/capability/configstore.cpp
```diff
@@ -0,0 +1,190 @@
...
+    if (!json_is_array(capabilities)) {
+      return ReturnCode::FAILURE;
+    }
+
+    std::map<std::string, std::map<std::string, json_t*>> caps;
+
+    size_t i;
+    json_t *capability;
+
+    json_array_foreach(capabilities,i,capability) {
```
tobsan, line 63: It would be nice to move this foreach to its own method instead

agent/src/capability/configstore.cpp
```diff
@@ -0,0 +1,190 @@
...
+      }
+      if (!json_is_array(gateways)) {
+        return ReturnCode::FAILURE;
+      }
+      size_t j;
+      json_t *gateway;
+
+      std::map<std::string, json_t*> gwConfs;
+
+      json_array_foreach(gateways, j, gateway) {
```
tobsan, line 85: It would be nice to move this inner foreach to its own method instead.

agent/src/capability/configstore.cpp
```diff
@@ -0,0 +1,190 @@
...
+std::vector<std::string> ConfigStore::fileList(const std::string &filePath)
+{
+    std::vector<std::string> files;
+
+    DIR *dirFile = opendir(filePath.c_str());
+    if (dirFile == nullptr) {
+        log_debug() << "Could not open directory: " << filePath << "\n";
+    } else {
+        struct dirent *hFile;
+        while ((hFile = readdir(dirFile)) != nullptr) {
```
tobsan, line 170: I think we should check for . and .. here.

agent/src/capability/capability.cpp
```diff
@@ -0,0 +1,40 @@
+
```
tobsan, line 1: Imho this capability class does not make much sense with the current setup. HOWEVER, if we want to use it to merge gateway configurations together, it would start to make more sense though. And it is nice to capture things in classes anyway.

thenor: You mean to move some of the functionality I currently have in configstore, or to add functionality related to merging capabilities?

tobsan: Yes. I mean you have two simple options: have the capabilities merge themselves (if they are captured in classes), or have some method in the config store that does the merging.

PR #78, opened 2016-10-18
Vagrant no longer autobuilds on provision

tobsan:


kursatkobya (approved):
Jenkins build seems okay and changes looks nice

Jenkinsfile
```diff
@@ -1,52 +1,55 @@
...
+        sh 'git submodule update --init'
 
-    stage 'Build'
-    // STart the machine and provision it
-    // TODO This should be split into two parts, one for provisioning and one for actually
-    // building softwarecontainer
-    sh "cd ${workspace} && vagrant up --provision"
+    stage 'StartVM'
+        // Start the machine (destroy it if present) and provision it
+        sh "cd ${workspace} && vagrant destroy -f"
```
frznlogic: Will vagrant destroy -f always return 0 or should we or it with true?

Vagrantfile
```diff
@@ -57,21 +57,4 @@ Vagrant.configure(2) do |config|
...
-        # clang analysis of the code
-        config.vm.provision "shell", privileged: false,
-            args: ["softwarecontainer", "clang", "-DENABLE_DOC=1 -DENABLE_TEST=ON -DENABLE_COVERAGE=1"],
-            path: "cookbook/build/clang-code-analysis.sh"
-
-        config.vm.provision "shell", inline: <<-SHELL
-            cd softwarecontainer
-            sudo su -c ./run-all-tests.sh
-        SHELL
-    end
```
frznlogic, line 76: Do we really want to delete all this, we dont want to keep it but with a if ENV thingy around it for "convenience"?

frznlogic (changes requested):

Jenkinsfile
tobsan: Good point, will update.

Vagrantfile
tobsan: Sure, we could do the same with the building in the provision step.

PR #79, opened 2016-10-18
Write user documentation of DBus API

kursatkobya:
DBus API documentation is added

### DIFF
--- a/doc/chapters/api/index.rst
+++ b/doc/chapters/api/index.rst
@@ -3,7 +3,7 @@
 API
 ***
 
-This chapter documents the DBus and C API for interacting with the SoftwareContainerAgent. This is
+This chapter documents the D-Bus API for interacting with the SoftwareContainerAgent. This is
 mostly useful for integrators.
 
 .. _dbus-api:
@@ -11,18 +11,10 @@ mostly useful for integrators.
 D-Bus API
 ========
 
-D-Bus API is an IPC interface to call SoftwareContainer agent methods. The API provides following path and interface.
+D-Bus API is an IPC interface to call SoftwareContainer agent methods. The API provides following object path and interface.
 
 :Object Path: /com/pelagicore/SoftwareContainer
 :Interface: com.pelagicore.SoftwareContainerAgent
-
-In this document object path and interface used in examples below with following order.::
-
-        dbus-send --system --print-reply \
-        --dest=Interface \
-        Object Path \
-        Method \
-        Arguments
 
 Methods
 -------
@@ -36,15 +28,6 @@ Controls availability of com.pelagicore.SoftwareContainerAgent interface
 |
 :Return Value:
         *None*
-|
-:Example Usage:
-
-::
-
-        dbus-send --system --print-reply \
-        --dest=com.pelagicore.SoftwareContainerAgent \
-        /com/pelagicore/SoftwareContainerAgent \
-        com.pelagicore.SoftwareContainerAgent.Ping
 
 CreateContainer
 ---------------
@@ -60,16 +43,6 @@ Creates a new container and returns created container id.
 |
 :Return Value:
         :containerID: ``uint32`` ID of created SoftwareContainer.
-|
-:Example Usage:
-
-::
-
-        dbus-send --system --print-reply \
-        --dest=com.pelagicore.SoftwareContainerAgent \
-        /com/pelagicore/SoftwareContainerAgent \
-        com.pelagicore.SoftwareContainerAgent.CreateContainer \
-        string:"00" string:'[{"enableWriteBuffer": true}]'
 
 SetContainerName
 ----------------
@@ -81,21 +54,10 @@ Sets the name of container with unique containerID.
 |
 :Return Value:
         *None*
-|
-:Example Usage:
-
-::
-
-        dbus-send --system --print-reply \
-        --dest=com.pelagicore.SoftwareContainerAgent \
-        /com/pelagicore/SoftwareContainerAgent \
-        com.pelagicore.SoftwareContainerAgent.SetContainerName \
-        uint32:0 string:"myContainer"
-
 
 LaunchCommand
 -------------
-Launches specific application/code int he container.
+Launches specific application/code in the container.
 
 :Parameters:
         :containerID: ``uint32`` The ID obtained by CreateContainer method.
@@ -107,47 +69,22 @@ Launches specific application/code int he container.
 |
 :Return Value:
         :pid: ``uint32`` PID of the process run inside the container.
-|
-:Example Usage:
-
-::
-
-        dbus-send --system --print-reply \
-        --dest=com.pelagicore.SoftwareContainerAgent \
-        /com/pelagicore/SoftwareContainerAgent \
-        com.pelagicore.SoftwareContainerAgent.LaunchCommand \
-        uint32:0 \
-        uint32:0 \
-        string:"touch hello" \
-        string:"/gateways/app/" \
-        string:"" \
-        dict:string:string:""
-        
+       
 
 ShutdownContainer
 -----------------
-Teardowns all active gateways related to container and shutdowns the container with all reserved sources.
+Tears down all active gateways related to container and shuts down the container with all reserved sources.
 
 :Parameters:
         :containerID: ``uint32`` The ID obtained by CreateContainer method.
 |
 :Return Value:
         *None*
-|
-:Example Usage:
-
-::
-
-        dbus-send --system --print-reply \
-        --dest=com.pelagicore.SoftwareContainerAgent \
-        /com/pelagicore/SoftwareContainerAgent \
-        com.pelagicore.SoftwareContainerAgent.ShutDownContainer \
-        uint32:0
 
 
 ShutdownContainerWithTimeout
 ----------------------------
-Teardowns all active gateways related to container and shutdowns the container and all reserved sources after given timeout.
+Tears down all active gateways related to container and shuts down the container and all reserved sources after given timeout.
 
 :Parameters:
         :containerID: ``uint32`` The ID obtained by CreateContainer method.
@@ -155,18 +92,6 @@ Teardowns all active gateways related to container and shutdowns the container a
 |
 :Return Value:
         *None*
-|
-:Example Usage:
-
-::
-
-        dbus-send --system --print-reply \
-        --dest=com.pelagicore.SoftwareContainerAgent \
-        /com/pelagicore/SoftwareContainerAgent \
-        com.pelagicore.SoftwareContainerAgent.ShutDownContainerWithTimeout \
-        uint32:0 \
-        uint32:5
-
 
 WriteToStdIn
 ------------
@@ -178,18 +103,6 @@ Send a character array to the standard input of a particular process.
 |
 :Return Value:
         *None*
-|
-:Example Usage:
-
-::
-
-        bus-send --system --print-reply \
-        --dest=com.pelagicore.SoftwareContainerAgent \
-        /com/pelagicore/SoftwareContainerAgent \
-        com.pelagicore.SoftwareContainerAgent.WriteToStdIn \
-        uint32:14859 \
-        array:byte:'a','b'        
-
 
 BindMountFolderInContainer
 --------------------------
@@ -203,20 +116,6 @@ Binds a directory on the host to the container.
 |
 :Return Value:
         :pathInContainer: ``string`` path to the bind folder in container. 
-|
-:Example Usage:
-
-::
-
-        dbus-send --system --print-reply \
-        --dest=com.pelagicore.SoftwareContainerAgent \
-        /com/pelagicore/SoftwareContainerAgent \
-        com.pelagicore.SoftwareContainerAgent.BindMountFolderInContainer \
-        uint32:1 \
-        string:"/home/myUser/myBindFolder" \
-        string:"/home/vagrant/softwarecontainer/build" \
-        boolean:false
-        
 
 SetGatewayConfigs
 -----------------
@@ -228,18 +127,6 @@ Sets the configuration of a particular gateway. The gateway configuration contai
 |
 :Return Value:
         *None*
-| 
-:Example Usage:
-
-::
-
-        dbus-send --system --print-reply \
-        --dest=com.pelagicore.SoftwareContainerAgent \
-        /com/pelagicore/SoftwareContainerAgent \
-        com.pelagicore.SoftwareContainerAgent.SetGatewayConfigs \
-        uint32:1 \
-        dict:string:string:"configItem1","1","configItem2","2"
-
 
 SetCapabilities
 ---------------
@@ -251,25 +138,13 @@ Currently This method has no applicable usage.
 |
 :Return Value:
         :success: ``boolean`` either true or false.
-|
-:Example Usage:
-
-::
-
-         dbus-send --system --print-reply \
-         --dest=com.pelagicore.SoftwareContainerAgent \
-         /com/pelagicore/SoftwareContainerAgent \
-         com.pelagicore.SoftwareContainerAgent.SetCapabilities \
-         uint32:1 \
-         array:string:"I","Can","not","provide","functionality"
-
 
 Signals
 -------
 
 ProcessStateChanged
 -------------------
-The D-Bus API sends signal when process state is changed. There are four value to be emitted.
+The D-Bus API sends signal when process state is changed. There are four values to be emitted.
 
 :containerID: ``uint32`` The ID obtained by CreateContainer method.
 
@@ -285,12 +160,4 @@ Introspection
 
 Using ``org.freedesktop.DBus.Introspectable.Introspect`` interface, methods of SoftwareContainerAgent D-Bus API can be observed.
 
-:Example Usage:
-
-::
-
-        dbus-send --system --print-reply \
-        --dest=com.pelagicore.SoftwareContainerAgent \
-        /com/pelagicore/SoftwareContainerAgent \
-        org.freedesktop.DBus.Introspectable.Introspect
 
--- a/doc/chapters/api/index.rst
+++ b/doc/chapters/api/index.rst
@@ -34,7 +34,7 @@ CreateContainer
 Creates a new container and returns created container id.
 
 :Parameters:
-        :IDPrefix: ``string`` prefix for name of back-end container.
+        :prefix: ``string`` prefix for name of back-end container.
         :config: ``string`` config file in json format.
         
                 Example config JSON::
@@ -57,7 +57,7 @@ Sets the name of container with unique containerID.
 
 LaunchCommand
 -------------
-Launches specific application/code in the container.
+Launches the specified application/code in the container.
 
 :Parameters:
         :containerID: ``uint32`` The ID obtained by CreateContainer method.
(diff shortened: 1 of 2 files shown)


doc/chapters/api/index.rst
```diff
@@ -1,21 +1,10 @@
...
-API
-***
-
-This chapter documents the DBus and C API for interacting with the SoftwareContainerAgent. This is
-mostly useful for integrators.
-
-.. TODO::
-    This chapter is still to be written properly. Some bits and pieces have
-    been but they are under development.
-
```
tobsan, line 13: Why is all of this removed? The main heading and introduction are still valid, so just remove the TODO

doc/chapters/api/index.rst
```diff
@@ -1,21 +1,10 @@
...
 .. _dbus-api:
 
 DBus API
 ========
 
+DBus API is an IPC interface to call software container agent methods.
```
tobsan: SoftwareContainer, not software container

doc/chapters/api/index.rst
```diff
@@ -24,88 +13,276 @@ Methods
...
+        - **None** 
+
+:Example Usage:
+
+::
+
+        dbus-send --system --print-reply \
+        --dest=com.pelagicore.SoftwareContainerAgent \
+        /com/pelagicore/SoftwareContainerAgent \
+        com.pelagicore.SoftwareContainerAgent.Ping
```
tobsan: I like that there are four lines here, but It would be nice with a description of why it has been split like that.

doc/chapters/api/index.rst
```diff
@@ -24,88 +13,276 @@ Methods
...
+
+:Parameters:
+        - **IDPrefix** : ``string`` prefix for random container id.
+        - **config** : ``string`` config file in json format.
+        
+                Example config JSON::
+        
+                [{"enableWriteBuffer": true}]
+
+
```
tobsan: Lots of blank lines here

doc/chapters/api/index.rst
```diff
@@ -24,88 +13,276 @@ Methods
...
-uint32 containerID CreateContainer(string containerIDPrefix, string config)
+:Example Usage:
+
+::
+
+        dbus-send --system --print-reply \
+        --dest=com.pelagicore.SoftwareContainerAgent \
+        /com/pelagicore/SoftwareContainerAgent \
+        com.pelagicore.SoftwareContainerAgent.CreateContainer \
+        string:"00" string:'[{"enableWriteBuffer": true}]'
```
tobsan: My comment about four lines is obviously valid here as well (even though there are five lines)

doc/chapters/api/index.rst
```diff
@@ -24,88 +13,276 @@ Methods
...
+        uint32:0 \
+        uint32:5
 
-void ShutdownContainerWithTimeout(uint32 containerID, uint32 timeout)
 
 WriteToStdIn
 ------------
-:processID: uint32
-:bytes: array<char>
+Send characters array to the standard input of particular process.
```
tobsan: of a particular process

DunderRoffe (commented):
Looks good overall but I do not think that the contents of the boxes should be bullet lists.

doc/chapters/api/index.rst
```diff
@@ -24,88 +13,276 @@ Methods
...
+        uint32:1 \
+        dict:string:string:"configItem1","1","configItem2","2"
 
-void SetGatewayConfigs(uint32 containerID, map<string, string> configs)
 
 SetCapabilities
 ---------------
-:containerID: uint32
-:capabilities: array<string>
+Currently This method has no applicable usage.
```
tobsan, line 133: This should probably be described as to how it is meant to work, not the current implementation.

doc/chapters/api/index.rst
```diff
@@ -24,88 +13,276 @@ Methods
...
 
-void WriteToStdIn(uint32 processID, array<char> bytes)
 
 BindMountFolderInContainer
 --------------------------
-:containerID: uint32
-:pathInHost: string
-:subPathInContainer: string
-:readOnly: bool
+Binds a directory in the container to a directory in the host/
```
tobsan: This binds a directory on the host to the container, not the other way around.

doc/chapters/api/index.rst
```diff
@@ -24,88 +13,276 @@ Methods
...
+        string:"/home/myUser/myBindFolder" \
+        string:"/home/vagrant/softwarecontainer/build" \
+        boolean:false
+        
 
 SetGatewayConfigs
 -----------------
-:containerID: uint32
-:configs: map<string, string>
+Sets the configuration of particular gateway. The gateway configuration contains settings as key/value pairs.
```
tobsan: Of **a** particular gateway.

tobsan (changes requested):
Looks good, but some things to be changed.

doc/chapters/api/index.rst
breakreturn: Use "D-Bus" for consistency with the rest of the docs.

doc/chapters/api/index.rst
breakreturn: Tobias, isn't it common practice to split long commands into readable lines that still can be pasted directly? Do we really need to explain formatting like that when there's no practical difference to e.g. have it as one long line?

doc/chapters/api/index.rst
```diff
@@ -24,88 +13,276 @@ Methods
...
+:Parameters:
+        - **containerID** ``uint32`` The ID obtained by CreateContainer method.
+        - **capabilities** ``array<string>``
+
+
+:Return Value:
+       - **success** ``boolean`` either true or false.
+
+:Example Usage:
+        - Althought there is no meaning to use this method, it can be used as following example.
```
breakreturn: Spelling of although

doc/chapters/api/index.rst
```diff
@@ -24,88 +13,276 @@ Methods
...
 Signals
 -------
 
 ProcessStateChanged
 -------------------
-:containerID: uint32
-:processID: uint32
-:isRunning: bool
-:exitCode: uint32
+The Dbus API sends signal when process state is changed. There are four value to be emitted.
```
breakreturn: Another variant of D-Bus, change them all to "D-Bus"

breakreturn (changes requested):
Looks nice in general, please see comments about  changes. Also, I think this should relate to the Getting started, e.g. linking between them and consolidate the descriptions etc.

doc/chapters/api/index.rst
tobsan: Maybe not, I thought that since the D-Bus parts was one on each line one might want to say what is ObjectPath, what is Interface etc. But perhaps one can expect the user to understand D-Bus since they are reading about the D-Bus API...

doc/chapters/api/index.rst
```diff
@@ -24,88 +13,276 @@ Methods
...
-void ShutdownContainerWithTimeout(uint32 containerID, uint32 timeout)
 
 WriteToStdIn
 ------------
-:processID: uint32
-:bytes: array<char>
+Send characters array to the standard input of particular process.
+
+:Parameters:
+        - **processID** ``uint32`` PID of the process; obtained LaunchCommand.
```
frznlogic: you obtained LaunchCommand? ;)

doc/chapters/api/index.rst
```diff
@@ -24,88 +13,276 @@ Methods
...
+        - **bytes** ``array<char>`` character array to sent stdin.
```
frznlogic: Grammar, sent i guess "character array sent to stdin." is more correct.

doc/chapters/api/index.rst
frznlogic: a character array

doc/chapters/api/index.rst
```diff
@@ -24,88 +13,276 @@ Methods
...
-:pathInHost: string
-:subPathInContainer: string
-:readOnly: bool
+Binds a directory in the container to a directory in the host/
+
+:Parameters:
+        - **containerID** ``uint32`` The ID obtained by CreateContainer method.
+        - **pathInHost** ``string`` path to the directory in host.
+        - **subPathInContainer** ``string`` path to the directory in container.
+        - **readOnly** ``bool`` availability of binded folder
```
frznlogic: Not availability I think... if so, this parameter should be renamed?

doc/chapters/api/index.rst
kursatkobya: So simply, no need to make any changes related this, am i right ?

doc/chapters/api/index.rst
kursatkobya: It is indeed but there are no blank lines in the html. For code snippet reStructuredText uses this structure.

doc/chapters/api/index.rst
```diff
@@ -24,88 +13,276 @@ Methods
...
 
 SetGatewayConfigs
 -----------------
-:containerID: uint32
-:configs: map<string, string>
+Sets the configuration of particular gateway. The gateway configuration contains settings as key/value pairs.
+
+:Parameters:
+        - **containerID** ``uint32`` The ID obtained by CreateContainer method.
+        - **configs** ``map<string, string>`` A map to key/value pairs.
```
frznlogic: what kind of key/value pairs?

doc/chapters/api/index.rst
```diff
@@ -24,88 +13,276 @@ Methods
...
-void SetGatewayConfigs(uint32 containerID, map<string, string> configs)
 
 SetCapabilities
 ---------------
-:containerID: uint32
-:capabilities: array<string>
+Currently This method has no applicable usage. 
+
+:Parameters:
+        - **containerID** ``uint32`` The ID obtained by CreateContainer method.
```
frznlogic: Actually... I just realized, it's good to know where the ID comes from, but would also be interesting to know what it is... Kind of obvious I guess, but none the less..

doc/chapters/api/index.rst
```diff
@@ -24,88 +13,276 @@ Methods
...
 
 ProcessStateChanged
 -------------------
-:containerID: uint32
-:processID: uint32
-:isRunning: bool
-:exitCode: uint32
+The Dbus API sends signal when process state is changed. There are four value to be emitted.
+
+:containerID: ``uint32`` The ID obtained by CreateContainer method.
```
frznlogic, line 149: A different formatting standard than in the previous D-Bus calls is used here. Same formatting all over would be nice.

doc/chapters/api/index.rst
```diff
@@ -24,88 +13,276 @@ Methods
...
+
+:isRunning: ``bool`` Whether the process is running or not.
+
+:exitCode: ``uint32`` exit code of Process.
+
+
+Introspection
+-------------
+
+Using ``org.freedesktop.DBus.Introspectable.Introspect`` interface, methhods of SoftwareContainerAgent DBus API can be observed.
```
frznlogic: methhods is misspelled

doc/chapters/api/index.rst
frznlogic: I agree with tobsan, this chapter should be used to document all public API's imho. For now, that only means the D-Bus API, but I'd like to have a top level chapter. Also, this chapter is pointed to by other chapters using the top level reference so if you insist on removing it you need to re-point the api references in the other chapters.

frznlogic (changes requested):
Some comments added, I would like some changes but overall, it's a good start.

doc/chapters/api/index.rst
breakreturn: I thought the comment was about why the command is split into four parts. I don't think it's necessary to explain that. However, if we will use dbus-send in the examples etc. then I agree it's a good idea to point out what the opath and iface is etc. But i think that's better suited then in the Getting started (or whatever that turns out to be in the end) rather than here. See my comment about consolidating with the rest of the docs.

Kursat, why have example invocations here as well as in the Getting started?

doc/chapters/api/index.rst
kursatkobya: So it is agreeed that it will be changed when implemented.

doc/chapters/api/index.rst
kursatkobya: i mean it is a character aray which is there to be sent to stdin, so maybe "character array to sent to the stdin" would be better. What do you think ?

doc/chapters/api/index.rst
kursatkobya: you mean is it better to express it like "ID of container"

doc/chapters/api/index.rst
kursatkobya: But this is not actually a D-Bus method to call

doc/chapters/api/index.rst
kursatkobya: Will be added again.

doc/chapters/api/index.rst
kursatkobya: We have used some of similar examples at Getting started chapter. So i just want to use same schema. But if it would be beteer we can create a small section which explain very basics of dbus-send. Or it is possible to give reference to its web page.

breakreturn: I think it would be enough to document the API and have the usage somewhere else. Perhaps add missing parts to the Getting started chapter, or just reference it from this chapter? I'm mainly concerned about maintenance.

I don't think we need to explain dbus-send much, just point out what parts of the command is the object path etc., that should be enough for the user to figure out any details about dbus-send.

doc/chapters/api/index.rst
kursatkobya: I will do so

doc/chapters/api/index.rst
kursatkobya: Some small cheat sheet added before Methods section.

tobsan (approved):

frznlogic (approved):
Looks good to me.

doc/chapters/api/index.rst
```diff
@@ -7,105 +6,291 @@ API
...
-DBus API
+D-Bus API
 ========
 
+D-Bus API is an IPC interface to call SoftwareContainer agent methods. The API provides following path and interface.
+
 :Path: /com/pelagicore/SoftwareContainer
 :Interface: com.pelagicore.SoftwareContainerAgent
 
+In this document path and interface above used in all method call examples below with following order.::
```
breakreturn: This sentence is difficult to grasp, could do with a rewrite. Also, I think we should use the words "object path" (possibly abbreviated as "opath" if needed) instead of "path". Object path is less ambiguous in the D-Bus domain.

doc/chapters/api/index.rst
```diff
@@ -7,105 +6,291 @@ API
...
 CreateContainer
 ---------------
-:containerIDPrefix: string
-:config: string
-
-uint32 containerID CreateContainer(string containerIDPrefix, string config)
+Creates a new container and returns created container id.
+
+:Parameters:
+        :IDPrefix: ``string`` prefix for random container id.
```
breakreturn: Currently, this prefix is part of the container name (in the LXC domain), and I think referring to it as "ID" in these parts of the docs might confuse the reader to think it has something to do with the container ID used to identify containers when using this API.

kursatkobya: Changed due to your comments.

doc/chapters/api/index.rst
kursatkobya: Changed due to your comments.

doc/chapters/api/index.rst
```diff
@@ -7,105 +6,291 @@ API
 This chapter documents the DBus and C API for interacting with the SoftwareContainerAgent. This is
```
thenor: Since you have used "D-Bus" further down in the document, I think you should also change this instance.

doc/chapters/api/index.rst
```diff
@@ -7,105 +6,291 @@ API
...
 -------------
-:containerID: uint32
-:userID: uint32
-:commandLine: string
-:workingDirectory: string
-:outputFile: string
-:env: map<string, string>
-
-uint32 pid LaunchCommand(uint32 containerID, uint32 userID, string commandLine, string workingDirectory, string outputFile, map<string, string> env)
+Launches specific application/code int he container.
```
thenor: Typo: "int he" should be "in the"

doc/chapters/api/index.rst
```diff
@@ -7,105 +6,291 @@ API
...
+        string:"touch hello" \
+        string:"/gateways/app/" \
+        string:"" \
+        dict:string:string:""
+        
 
 ShutdownContainer
 -----------------
-:containerID: uint32
+Teardowns all active gateways related to container and shutdowns the container with all reserved sources.
```
thenor: Should be "Tears down" and "shuts down", since this is the verb and not the action teardown/shutdown.

doc/chapters/api/index.rst
```diff
@@ -7,105 +6,291 @@ API
...
+        com.pelagicore.SoftwareContainerAgent.ShutDownContainer \
+        uint32:0
 
-void ShutdownContainer(uint32 containerID)
 
 ShutdownContainerWithTimeout
 ----------------------------
-:containerID: uint32
-:timeout: uint32
+Teardowns all active gateways related to container and shutdowns the container and all reserved sources after given timeout.
```
thenor: Same as above: should be "tears down" and "shuts down"

doc/chapters/api/index.rst
thenor: I am currently writing this method. So we could describe it on a high level, but doesn't it feel like the effort might be a bit wasted if I re-work the code a bit?
That is why I felt that @kursatkobya could leave it like this until I am finished.
Should we instead write something high level, as you say, @tobsan ?

doc/chapters/api/index.rst
```diff
@@ -7,105 +6,291 @@ API
...
 Signals
 -------
 
 ProcessStateChanged
 -------------------
-:containerID: uint32
-:processID: uint32
-:isRunning: bool
-:exitCode: uint32
+The D-Bus API sends signal when process state is changed. There are four value to be emitted.
```
thenor: "this signal" perhaps?
"four values" (s at the end)

doc/chapters/api/index.rst
```diff
@@ -7,105 +6,291 @@ API
...
+-------------
+
+Using ``org.freedesktop.DBus.Introspectable.Introspect`` interface, methods of SoftwareContainerAgent D-Bus API can be observed.
+
+:Example Usage:
 
-void ProcessStateChanged(uint32 containerID, uint32 processID, bool isRunning, uint32 exitCode)
+::
 
-C API
```
thenor, line 109: So the C API is no longer valid?

kursatkobya: I suspect if if there was any real C API ever.

doc/chapters/api/index.rst
```diff
@@ -1,111 +1,163 @@
...
 -------------
-:containerID: uint32
-:userID: uint32
-:commandLine: string
-:workingDirectory: string
-:outputFile: string
-:env: map<string, string>
-
-uint32 pid LaunchCommand(uint32 containerID, uint32 userID, string commandLine, string workingDirectory, string outputFile, map<string, string> env)
+Launches specific application/code in the container.
```
thenor: I think "specific" should be changed, perhaps write "Launches the specified ..."

thenor (changes requested):
One small change. Otherwise I approve.

thenor (approved):
Looks good!

breakreturn (approved):

PR #80, opened 2016-10-18
Enables building of ONLY docs with cmake

tobsan:
You can now run cmake on the doc directory directly, and then
run make to get only the docs. It will not check for other
build dependencies, just sphinx and doxygen.


frznlogic (changes requested):
Would be nice with documentation on this. Otherwise, really nice.

tobsan:
Actually, that this works was already (incorrectly) stated in the README.md file. So, docs are there.

DunderRoffe (approved):

frznlogic (approved):

PR #83, opened 2016-10-19
Created script that pushes to gh-pages

DunderRoffe:


scripts/publish_docs.sh
```diff
@@ -0,0 +1,32 @@
+#!/bin/bash
+set -e
+
+### Fetch the gh-pages branch and switch to it ###
+git fetch origin gh-pages
+git checkout -qf FETCH_HEAD
+git checkout -f gh-pages || git checkout -bf gh-pages
+
+### Setting git author ###
+if [[ "`git config user.name`" == "" ]]
```
tobsan, line 11: Would be nicer to user -n instead of check if empty manually imho

scripts/publish_docs.sh
```diff
@@ -0,0 +1,32 @@
...
+
+API_DOCS="api-docs"
+USER_DOCS="user-docs"
+
+git status
+git rm -rf ${API_DOCS} ${USER_DOCS} || true
+mkdir ${API_DOCS}
+mkdir ${USER_DOCS}
+cp -r build/doc/doxygen-docs/html/* ${API_DOCS}
+cp -r build/doc/html/* ${USER_DOCS}
```
tobsan: This can be simplified into `cp -r build/doc/doxygen-docs/html ${API_DOCS}` and then you don't need the mkdir

scripts/publish_docs.sh
```diff
@@ -0,0 +1,32 @@
...
+
+API_DOCS="api-docs"
+USER_DOCS="user-docs"
+
+git rm -rf ${API_DOCS}  || true
+git rm -rf ${USER_DOCS} || true
+mkdir ${API_DOCS}  || true
+mkdir ${USER_DOCS} || true
+cp -r build/doc/doxygen-docs/html/* ${API_DOCS}
+cp -r build/doc/html/* ${USER_DOCS}
```
tobsan: This can be simplified into `cp -r build/doc/doxygen-docs/html ${API_DOCS}` and remove the mkdir above.

tobsan (changes requested):
Looks good in general.

scripts/publish_docs.sh
DunderRoffe: Nice :D

scripts/publish_docs.sh
DunderRoffe: OK, I fix

scripts/publish_docs.sh
```diff
@@ -0,0 +1,32 @@
...
+    git config user.email "noreply@pelagicore.com"
+    git config user.name "Jenkins"
+fi
+
+### Moving docs to root directory and committing ###
+
+API_DOCS="api-docs"
+USER_DOCS="user-docs"
+
+git rm -rf ${API_DOCS}  || true
```
frznlogic: Fnuttar runt API_DOCS

scripts/publish_docs.sh
DunderRoffe: Did not work :/

scripts/publish_docs.sh
```diff
@@ -0,0 +1,32 @@
+#!/bin/bash
+set -e
+
+### Fetch the gh-pages branch and switch to it ###
+git fetch origin gh-pages
+git checkout -qf FETCH_HEAD
+git checkout -f gh-pages || git checkout -bf gh-pages
```
frznlogic, line 7: Why aren't we running with -b directly instead of first failing to checkout and then doing it?

frznlogic (changes requested):

scripts/publish_docs.sh
DunderRoffe: Fixed

scripts/publish_docs.sh
DunderRoffe: It fails if there already is a such branch

frznlogic (approved):

tobsan (approved):

PR #89, opened 2016-10-21
Build with -pedantic and -Wextra

tobsan:
In doing this, some warnings from dbus-c++ are suppressed
by using GCC pragmas for that.

### DIFF
--- a/agent/src/softwarecontaineragent.cpp
+++ b/agent/src/softwarecontaineragent.cpp
@@ -48,7 +48,6 @@ bool SoftwareContainerAgent::triggerPreload()
 inline bool SoftwareContainerAgent::isIdValid (ContainerID containerID)
 {
     return ((containerID < UINT32_MAX)
-            && (containerID >= 0)
             && (1 == m_containers.count(containerID)));
 }
 
@@ -228,7 +227,7 @@ void SoftwareContainerAgent::shutdownContainer(ContainerID containerID, unsigned
     if (m_shutdownContainers) {
         SoftwareContainer *container = nullptr;
         if (checkContainer(containerID, container)) {
-            container->shutdown();
+            container->shutdown(timeout);
             deleteContainer(containerID);
         }
     } else {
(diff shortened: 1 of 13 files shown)


agent/src/softwarecontaineragent.cpp
```diff
@@ -48,7 +48,6 @@ bool SoftwareContainerAgent::triggerPreload()
 inline bool SoftwareContainerAgent::isIdValid (ContainerID containerID)
 {
     return ((containerID < UINT32_MAX)
-            && (containerID >= 0)
```
tobsan, line 51: FYI: ContainerID type is unsigned.

kursatkobya: Yes but you can call function as IdValid(-1). 
Although in this case, it will return false due to map search, there is a meaning for this in my opinion. At least in this case, it will save us from searching map overhead.

tobsan: I might be wrong here, but -1 would be 0xFFFFFFFF in an unsigned case, which is >= 0.

kursatkobya: If compiler is coherent to c99 you are right, so ignore my comment then

tobsan: I guess if we have code that ever does this it would in turn generate a warning from the compiler that we're passing a signed int to this function, if it's even possible.

kursatkobya: Warning will be there for sure but i expect it to be compiled anyway. In any case you are right, the line should be removed.

thenor (commented):
I don't feel I know enough to Request Changes or Approve, but it looks good to me. 
Only part I would like clarified is why the DBus API doc is changed.

tobsan:
@thenor Sphinx generated warnings about missing newlines and unexpected unindents for the DBus API docs.

frznlogic (approved):

PR #90, opened 2016-10-21
Removes launchCommand from softwarecontainer.cpp

tobsan:
This method was not used at all.


kursatkobya (approved):
Looks reasonable

PR #91, opened 2016-10-23
Cmake cleanup

tobsan:
Adding message levels, some error checking and more configuration to network bridge setup.

Also only builds the unit tests for specific gateways if those gateways are enabled.


CMakeLists.txt
```diff
@@ -154,40 +154,58 @@ if(ENABLE_COVERAGE)
...
 
-if(CREATE_BRIDGE)
-    message("SoftwareContainer will create ${BRIDGE_DEVICE} if not present at runtime")
-    file(READ scripts/setup_create_nobridge.sh.in NOBRIDGE)
-else()
-    message("SoftwareContainer will exit if ${BRIDGE_DEVICE} is not present at runtime")
-    file(READ scripts/setup_exit_nobridge.sh.in NOBRIDGE)
-endif()
+    #
+    # Should SoftwareContainer create a bridge if not present?
```
breakreturn: Nitpicking: this is almost phrased as a TODO, it could perhaps be more clear that we know what we want to check and how to behave.

frznlogic (approved):

thenor (approved):

libsoftwarecontainer/unit-test/CMakeLists.txt
```diff
@@ -31,17 +31,28 @@ set(TEST_LIBRARY_DEPENDENCIES
...
+
+add_gateway_test(ENABLE_CGROUPSGATEWAY cgroupsgateway_unittest.cpp)
+add_gateway_test(ENABLE_DBUSGATEWAY dbusgateway_unittest.cpp)
+add_gateway_test(ENABLE_DEVICENODEGATEWAY devicenodegateway_unittest.cpp)
+add_gateway_test(ENABLE_NETWORKGATEWAY networkgateway_unittest.cpp)
+add_gateway_test(ENABLE_NETWORKGATEWAY netlink_unittest.cpp)
+
+set(DEFAULT_GATEWAYS TRUE)
+add_gateway_test(DEFAULT_GATEWAYS envgateway_unittest.cpp)
+add_gateway_test(DEFAULT_GATEWAYS filegateway_unittest.cpp)
```
DunderRoffe, line 49: Hard coding DEFAULT_GATEWAYS to TRUE and then using it to add test seems excessive.
Maybe use a comment instead?

tobsan: Actually, it follows the pattern used for adding gateways in libsoftwarecontainer/src/CMakeLists.txt. It could be changed of course.

DunderRoffe: I does not really matter. If there is an established scheme then it might be better to use that

PR #98, opened 2016-10-25
Changes ASSERT_TRUE to ASSERT_EQ for lots of cases.

tobsan:
Also removes all asserts in the code and replaces them with
proper error handling.


common/softwarecontainer-common.cpp
```diff
@@ -35,6 +35,8 @@ LOG_DECLARE_DEFAULT_CONTEXT(defaultLogContext, "MAIN", "Main context");
 struct stat getStat(const std::string &path)
 {
     struct stat st;
+    memset(&st, 0, sizeof(st));
+
```
DunderRoffe: Should be in a separate commit

DunderRoffe (changes requested):

thenor (approved):
Looks good to me.

DunderRoffe (approved):

PR #102, opened 2016-10-26
The container ID set by Agent is used for LXC ID

DunderRoffe:
Refactored how the ids for the LXC containers is set.
Now the agent sets the entire ID of the container instead of
only a prefix followed by a generated string.

However, if SoftwareContainer::init() is called before the
container ID is set an ID will ge generated prefixed by "SC-".

### DIFF
--- a/agent/src/softwarecontaineragent.cpp
+++ b/agent/src/softwarecontaineragent.cpp
@@ -38,7 +38,9 @@ bool SoftwareContainerAgent::triggerPreload()
             log_error() << "Preloading failed";
             return false;
         }
-        m_preloadedContainers.push_back(SoftwareContainerPtr(container));
+        auto availableID = findSuitableId();
+        auto pair = std::pair<ContainerID, SoftwareContainerPtr>(availableID, SoftwareContainerPtr(container));
+        m_preloadedContainers.push_back(std::move(pair));
     }
 
     return true;
@@ -139,14 +141,17 @@ ContainerID SoftwareContainerAgent::createContainer(const std::string &prefix, c
     parseConfig(config);
 
     SoftwareContainer *container;
-    if (m_preloadedContainers.size() != 0) {
-        container = m_preloadedContainers[0].release();
-        m_preloadedContainers.erase(m_preloadedContainers.begin());
+    ContainerID availableID;
+    if (!m_preloadedContainers.empty()) {
+        auto pair = std::move(m_preloadedContainers.back());
+        m_preloadedContainers.pop_back();
+
+        availableID = pair.first;
+        container   = pair.second.release();
     } else {
+        availableID = findSuitableId();
         container = new SoftwareContainer(m_softwarecontainerWorkspace);
     }
-
-    ContainerID availableID = findSuitableId();
 
     m_containers[availableID] = SoftwareContainerPtr(container);
     log_debug() << "Created container with ID :" << availableID;
--- a/agent/src/softwarecontaineragent.cpp
+++ b/agent/src/softwarecontaineragent.cpp
@@ -33,7 +33,6 @@ bool SoftwareContainerAgent::triggerPreload()
 {
     while (m_preloadedContainers.size() < m_preloadCount) {
         auto container = new SoftwareContainer(m_softwarecontainerWorkspace);
-        container->setContainerIDPrefix("Preload-");
 
         if (isError(container->preload())) {
             log_error() << "Preloading failed";
@@ -151,7 +150,7 @@ ContainerID SoftwareContainerAgent::createContainer(const std::string &prefix, c
 
     m_containers[availableID] = SoftwareContainerPtr(container);
     log_debug() << "Created container with ID :" << availableID;
-    container->setContainerIDPrefix(prefix);
+    container->setContainerID(prefix + std::to_string(availableID));
     container->setMainLoopContext(m_mainLoopContext);
     container->init();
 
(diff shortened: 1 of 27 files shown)


kursatkobya (approved):

libsoftwarecontainer/src/softwarecontainer.cpp
```diff
@@ -81,16 +79,13 @@ void SoftwareContainer::setContainerName(const std::string &name)
...
 
-void SoftwareContainer::validateContainerID()
+ReturnCode SoftwareContainer::preload()
 {
-    if (m_containerID.size() == 0) {
-        setContainerIDPrefix("PLC-");
+    if (m_containerID.empty()) {
+        const std::string newID = "SC-PRELOADED-" + Generator::gen_ct_name();
+        setContainerID(newID);
     }
```
breakreturn: Can we have all IDs (and thus also LXC names) use the same scheme? I.e. have the preloaded containers also use prefix-nextavailableid?

libsoftwarecontainer/include/container.h
```diff
@@ -75,8 +75,7 @@ class Container :
      * @param shutdownTimeout Timeout for shutdown of container.
      */
     Container(
-            const std::string &id
-            , const std::string &name
+            const std::string id
```
frznlogic, line 78: No longer taken by reference, is this correct?

agent/src/softwarecontaineragent.h
```diff
@@ -226,7 +228,7 @@ class SoftwareContainerAgent : protected softwarecontainer::JSONParser
     // List of containers in use
     std::map<ContainerID, SoftwareContainerPtr> m_containers;
     // List of pre-loaded containers
-    std::vector<SoftwareContainerPtr> m_preloadedContainers;
+    std::vector<std::pair<ContainerID, SoftwareContainerPtr>> m_preloadedContainers;
```
tobsan: I'm guessing there is a good reason for having a vector of pairs instead of a map, but I don't see it.

frznlogic: I know I thought a vector would be good here, but maybe a std::map is better after all?

libsoftwarecontainer/src/container.cpp
```diff
@@ -57,15 +57,15 @@ void Container::init_lxc()
...
+Container::Container(const std::string id, const std::string &configFile,
         const std::string &containerRoot, bool enableWriteBuffer, int shutdownTimeout) :
     m_configFile(configFile),
     m_id(id),
-    m_name(name),
     m_containerRoot(containerRoot),
     m_enableWriteBuffer(enableWriteBuffer),
     m_shutdownTimeout(shutdownTimeout)
 {
+    log_debug() << "Container constructed with " + id;
```
frznlogic: I'd << operator instead of + operator for style reasons, not sure what you think.

frznlogic (approved):
I got some questions, but nothing that blocks this from being approved really imho.

agent/src/softwarecontaineragent.h
DunderRoffe: Yes, so the reason is that I need to store the ContainerID but I never want to lookup values. The only functionality that is needed is to pop from the collection. Then maybe a stack would be better.

libsoftwarecontainer/include/container.h
DunderRoffe: Yes. Container holds this value now instead of SoftwareContainer since SoftwareContainer does not use it.

agent/src/softwarecontaineragent.h
DunderRoffe: I am thinking of changing to a stack actually. The only operations that are applied to the collection is push and pop anyway.

libsoftwarecontainer/src/container.cpp
DunderRoffe: Oh, will fix

agent/src/softwarecontaineragent.h
DunderRoffe: Actually a queue is even better

agent/src/softwarecontaineragent.cpp
```diff
@@ -132,26 +133,34 @@ ContainerID SoftwareContainerAgent::findSuitableId()
...
+    if (!m_preloadedContainers.empty()) {
+        auto pair = std::move(m_preloadedContainers.front());
+        m_preloadedContainers.pop();
+
+        availableID = pair.first;
+        container   = pair.second.release();
     } else {
-        container = new SoftwareContainer(m_softwarecontainerWorkspace);
+        availableID = findSuitableId();
+        container = makeSoftwareContainer(availableID);
```
tobsan, line 160: Since we only want to modify the container below, and only if it's not already preloaded (then init() is superfluous), can't we use insert on m_containers with a pair in both clauses somehow? I think it would be worth a shot. If it can be easily simplified, I would like to see that. If not, then nevermind.

tobsan (approved):

agent/src/softwarecontaineragent.h
```diff
@@ -213,6 +212,9 @@ class SoftwareContainerAgent : protected softwarecontainer::JSONParser
     std::shared_ptr<Workspace> getWorkspace();
 
 private:
+    // Helper for creating software container instances
+    SoftwareContainer* makeSoftwareContainer(const ContainerID &containerID);
```
breakreturn: We put '*' by the variable name not by the type.

PR #103, opened 2016-10-27
README: Add 'apt update' step

jonte:
On systems with outdated package repositories, the apt install staps
will fail without this.

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,8 +73,9 @@ On a debian-based system, issue the following commands to build using Vagrant:
 git submodule init
 git submodule update
 
-sudo apt install virtualbox
-sudo apt install vagrant
+sudo apt update
+sudo apt install virtualbox vagrant
+
 vagrant up
 ```
 


jonte:
This fails to build due to the fix in #105 not being included yet.

tobsan:
Should we really tell users to update their package repos before installing new software? Isn't that kind of obvious?

jonte:
I assume the purpose of this code block is to be able to cut and paste the instructions? That's not possible currently.

tobsan:
The fact that your apt cache was out of date doesn't mean it's not possible. I don't think I have ever seen instructions for installing packages through apt that remind you to update your cache first, since that step is kinda implicit or expected to be done if something fails, for example with a 404.

jonte:
Oops. I accidentally closed this PR..

tobsan:
Actually, looking more at the context now, I tend to change my mind.

PR #104, opened 2016-10-27
Vagrantfile: Enable examples

jonte:


tobsan (approved):

jonte:
This fails to build due to the fix in #105 not being included yet.

PR #106, opened 2016-10-27
examples: Unset RPATH for all examples

jonte:
The default RPATH set by a parent CMakeLists adds a superflous RPATH
which causes QA errors in some build systems (Yocto).

### DIFF
--- a/examples/CMakeLists.txt
+++ b/examples/CMakeLists.txt
@@ -16,5 +16,7 @@
 #
 # For further information see LICENSE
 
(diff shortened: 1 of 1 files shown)


tobsan:
Do we need to set RPATH for anything? I don't know enough about it, but I know we got Yocto warnings for it a while back when it was first added.

frznlogic (approved):

frznlogic:
The examples should not neeed the rpath, no, that's true.

PR #107, opened 2016-10-28
Adds an example for running wayland

tobsan:
This actually does not add a wayland application, but relies on the user
of this example to have weston-simple-egl and weston-flower installed.
These are part of the weston repository.

### DIFF
--- a/examples/simple/launch.sh
+++ b/examples/simple/launch.sh
@@ -62,22 +62,44 @@ AGENTPID="$!"
 # Let the agent start up
 sleep 2
 
-export PCCMD="dbus-send --${BUS} --print-reply --dest=com.pelagicore.SoftwareContainerAgent /com/pelagicore/SoftwareContainerAgent"
+# Destination
+PCNAME="com.pelagicore.SoftwareContainerAgent"
+# Object path
+PCOBJPATH="/com/pelagicore/SoftwareContainerAgent"
+# Prefix for dbus methods
+AGENTPREFIX="com.pelagicore.SoftwareContainerAgent"
+export PCCMD="dbus-send --${BUS} --print-reply --dest=$PCNAME $PCOBJPATH"
 
 # Introspect the agent
 $PCCMD org.freedesktop.DBus.Introspectable.Introspect
 
 # Ping the agent
-$PCCMD com.pelagicore.SoftwareContainerAgent.Ping
+$PCCMD $AGENTPREFIX.Ping
 
 # Create a new container
-$PCCMD com.pelagicore.SoftwareContainerAgent.CreateContainer string:'[{"writeOften": "0"}]'
+$PCCMD $AGENTPREFIX.CreateContainer string:prefix string:'[{"writeOften": "0"}]'
+
+# A few thing that we use for more or less every call below
+CONTAINERID="uint32:0"
+ROOTID="uint32:0"
+OUTFILE="/tmp/stdout"
 
 # Expose a directory to the container
-$PCCMD com.pelagicore.SoftwareContainerAgent.BindMountFolderInContainer uint32:0 string:${SCRIPTPATH} string:app boolean:true
+$PCCMD $AGENTPREFIX.BindMountFolderInContainer \
+    $CONTAINERID \
+    string:${SCRIPTPATH} \
+    string:app \
+    boolean:true
+APPBASE="/gateways/app"
 
 # Run the simple example
-$PCCMD com.pelagicore.SoftwareContainerAgent.LaunchCommand uint32:0 uint32:0 string:/gateways/app/simple string:/gateways/app/ string:/tmp/stdout dict:string:string:""
+$PCCMD $AGENTPREFIX.LaunchCommand \
+    $CONTAINERID \
+    $ROOTID \
+    string:$APPBASE/simple \
+    string:$APPBASE \
+    string:$OUTFILE \
+    dict:string:string:""
 
 # Let the example run for a while
 sleep 30
--- a/examples/simple/launch.sh
+++ b/examples/simple/launch.sh
@@ -36,7 +36,7 @@ help() {
 
 BUS="system"
 while getopts ":b:h" opt; do
-    case $opt in 
+    case $opt in
     b)
         BUS="$OPTARG"
         ;;
@@ -63,21 +63,21 @@ AGENTPID="$!"
 sleep 2
 
 # Destination
-PCNAME="com.pelagicore.SoftwareContainerAgent"
+SCNAME="com.pelagicore.SoftwareContainerAgent"
 # Object path
-PCOBJPATH="/com/pelagicore/SoftwareContainerAgent"
+SCOBJPATH="/com/pelagicore/SoftwareContainerAgent"
 # Prefix for dbus methods
 AGENTPREFIX="com.pelagicore.SoftwareContainerAgent"
-export PCCMD="dbus-send --${BUS} --print-reply --dest=$PCNAME $PCOBJPATH"
+export SC_CMD="dbus-send --${BUS} --print-reply --dest=$SCNAME $SCOBJPATH"
 
 # Introspect the agent
-$PCCMD org.freedesktop.DBus.Introspectable.Introspect
+$SC_CMD org.freedesktop.DBus.Introspectable.Introspect
 
 # Ping the agent
-$PCCMD $AGENTPREFIX.Ping
+$SC_CMD $AGENTPREFIX.Ping
 
 # Create a new container
-$PCCMD $AGENTPREFIX.CreateContainer string:prefix string:'[{"writeOften": "0"}]'
+$SC_CMD $AGENTPREFIX.CreateContainer string:prefix string:'[{"writeOften": "0"}]'
 
 # A few thing that we use for more or less every call below
 CONTAINERID="uint32:0"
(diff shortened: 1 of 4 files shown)


tobsan:
I would like some input on whether or not this is enough as an example, or if we should provide an actual wayland application of our own to test with. The weston-examples feels like a safe bet however.

examples/wayland/launch.sh
```diff
@@ -0,0 +1,119 @@
...
+WESTONPID="$!"
+
+# Launch a new agent
+softwarecontainer-agent &
+AGENTPID="$!"
+
+# Let the agent start up
+sleep 2
+
+export PCCMD="dbus-send --${BUS} --print-reply --dest=com.pelagicore.SoftwareContainerAgent /com/pelagicore/SoftwareContainerAgent"
```
DunderRoffe: Really long line

examples/wayland/launch.sh
```diff
@@ -0,0 +1,119 @@
...
+
+# Create a new container
+$PCCMD com.pelagicore.SoftwareContainerAgent.CreateContainer string:prefix string:'[{"writeOften": "0"}]'
+
+# Set gateway config allowing wayland access
+WAYLANDKEY="wayland"
+WAYLANDVALUE="[{\"enabled\": true}]"
+
+# Enable wayland gateway
+$PCCMD com.pelagicore.SoftwareContainerAgent.SetGatewayConfigs uint32:0 dict:string:string:"$WAYLANDKEY","$WAYLANDVALUE"
```
DunderRoffe: Really long line

examples/wayland/launch.sh
```diff
@@ -0,0 +1,119 @@
...
+
+# Expose a directory to the container
+$PCCMD com.pelagicore.SoftwareContainerAgent.BindMountFolderInContainer uint32:0 string:${SCRIPTPATH} string:app boolean:true
```
DunderRoffe: Really long line

examples/wayland/launch.sh
```diff
@@ -0,0 +1,119 @@
...
+
+# Run the simple egl example from weston
+$PCCMD com.pelagicore.SoftwareContainerAgent.LaunchCommand uint32:0 uint32:0 string:"$SIMPLEWESTON" string:/gateways/app string:/tmp/stdout dict:string:string:""
```
DunderRoffe: Really long line

examples/wayland/launch.sh
```diff
@@ -0,0 +1,119 @@
...
+
+# Run weston-flower
+$PCCMD com.pelagicore.SoftwareContainerAgent.LaunchCommand uint32:0 uint32:0 string:"$FLOWERWESTON" string:/gateways/app string:/tmp/stdout dict:string:string:""
```
DunderRoffe: Really long line

examples/wayland/launch.sh
```diff
@@ -0,0 +1,119 @@
...
+        help
+        ;;
+    *)
+        echo "Unknown option $OPTARG"
+        help
+    esac
+done
+
+SIMPLEWESTON=$(which weston-simple-egl)
+if [ "$?" != "0" ]; then
```
DunderRoffe, line 49: Is not double brackets required to do these types of comparisons?

examples/wayland/launch.sh
```diff
@@ -0,0 +1,119 @@
...
+
+# Let the examples run for a while
+sleep 5
+
+# Shutdown the container
+$PCCMD com.pelagicore.SoftwareContainerAgent.ShutDownContainer uint32:0
+
+# Clean up
+kill $WESTONPID
+kill $AGENTPID
```
DunderRoffe, line 137: What if they do not cleanly shutdown?

examples/wayland/README.md
```diff
@@ -0,0 +1,21 @@
+Example for running wayland applications in SoftwareContainer
+
+Requirements
+============
```
DunderRoffe: Presonal preference: second heading is better for this case

examples/wayland/README.md
```diff
@@ -0,0 +1,21 @@
...
+* weston
+* weston-simple-egl and weston-flower examples
+* softwarecontainer
+
+How to run
+==========
```
DunderRoffe: Presonal preference: second heading is better for this case

examples/wayland/README.md
```diff
@@ -0,0 +1,21 @@
...
+```
+# ./launch.sh
+```
+
+Known issues
+============
```
DunderRoffe: Presonal preference: second heading is better for this case

DunderRoffe (commented):
Overall good but I have some suggestions.

examples/wayland/README.md
```diff
@@ -0,0 +1,21 @@
...
+* Running the example from ssh in a VM does not seem to work, because
+  weston refuses to start without a proper tty.
+** WORKAROUND: Have X11 or another wayland compositor running first
+   and run the example from xterm or weston-terminal (for example).
```
frznlogic, line 21: You can somehow get around this using stty, I can't remember exactly now however.

examples/wayland/launch.sh
frznlogic: This is an example, not a test, and should as cleanly as possibly show "den röda tråden". I think it should not take into account that softwarecontaineragent might be fucked up in this instance.

examples/wayland/launch.sh
frznlogic: I really don't like all the magic numbers we have (uint32:0 in this case). It would be nice to make variables of them with a hint at what they actually do, for clarity.

examples/wayland/launch.sh
```diff
@@ -0,0 +1,119 @@
...
+$PCCMD com.pelagicore.SoftwareContainerAgent.LaunchCommand uint32:0 uint32:0 string:"$SIMPLEWESTON" string:/gateways/app string:/tmp/stdout dict:string:string:""
+
+# Run weston-flower
+$PCCMD com.pelagicore.SoftwareContainerAgent.LaunchCommand uint32:0 uint32:0 string:"$FLOWERWESTON" string:/gateways/app string:/tmp/stdout dict:string:string:""
+
+# Let the examples run for a while
+sleep 5
+
+# Shutdown the container
+$PCCMD com.pelagicore.SoftwareContainerAgent.ShutDownContainer uint32:0
```
frznlogic: Same here with the uint32:0, would be nice with it set in a named variable so it makes more sense.

examples/wayland/launch.sh
```diff
@@ -0,0 +1,119 @@
...
+eval `dbus-launch --sh-syntax`
+
+# Print the environment
+env
+
+# Launch weston
+export XDG_RUNTIME_DIR=/tmp/wayland/
+mkdir -p $XDG_RUNTIME_DIR
+chmod 0700 $XDG_RUNTIME_DIR
+weston &
```
frznlogic, line 70: Failing out if weston isn't available would also be good I guess, the system would be broken if it has weston-flower byt not weston I guess but... who knows...

frznlogic (changes requested):
I'm fine with using the weston-flower etc, they should be available and as long as you dockument the requirements (already one), it's good.

examples/wayland/launch.sh
DunderRoffe: +1

examples/wayland/launch.sh
tobsan: Actually, this would behave the same in double brackets. See http://tldp.org/LDP/abs/html/comparison-ops.html#SCOMPARISON2

examples/wayland/launch.sh
tobsan: Good point, I might update that for the simple example as well then.

examples/wayland/README.md
tobsan: That would be really nice, then one could run the example without having x11 on the host. Will have a look.

examples/wayland/launch.sh
DunderRoffe: That is true. Maybe it is not as important in that case then.

tobsan:
Updated after some comments. Also updated the simple example to make a nicer launch.sh

examples/simple/launch.sh
```diff
@@ -62,22 +62,44 @@ AGENTPID="$!"
...
 sleep 2
 
-export PCCMD="dbus-send --${BUS} --print-reply --dest=com.pelagicore.SoftwareContainerAgent /com/pelagicore/SoftwareContainerAgent"
+# Destination
+PCNAME="com.pelagicore.SoftwareContainerAgent"
+# Object path
+PCOBJPATH="/com/pelagicore/SoftwareContainerAgent"
+# Prefix for dbus methods
+AGENTPREFIX="com.pelagicore.SoftwareContainerAgent"
+export PCCMD="dbus-send --${BUS} --print-reply --dest=$PCNAME $PCOBJPATH"
```
breakreturn, line 72: This example, and e.g. the simple example, all use "PC" as part of a lot of variable names. This doesn't make any sense anymore. No prefix or "SC" is more suitable.

examples/wayland/launch.sh
tobsan: Well...in that case we should do the same with anything that is not a shell builtin... The reason I checek for the examples is that they are not normally built with weston, but one has to enable them explicitly.

PR #108, opened 2016-10-28
Added service-test suite for environment GW

breakreturn:
A suite with tests for the basic functionality of the
env GW, e.g. making sure the setting of vars works,
and the interaction between different configs etc.

### DIFF
--- a/servicetest/environment/test_environment.py
+++ b/servicetest/environment/test_environment.py
@@ -1,0 +1,277 @@
+
+# Copyright (C) 2016 Pelagicore AB
+#
+# Permission to use, copy, modify, and/or distribute this software for
+# any purpose with or without fee is hereby granted, provided that the
+# above copyright notice and this permission notice appear in all copies.
+#
+# THE SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL
+# WARRANTIES WITH REGARD TO THIS SOFTWARE INCLUDING ALL IMPLIED
+# WARRANTIES OF MERCHANTABILITY AND FITNESS. IN NO EVENT SHALL THE AUTHOR
+# BE LIABLE FOR ANY SPECIAL, DIRECT, INDIRECT, OR CONSEQUENTIAL DAMAGES
+# OR ANY DAMAGES WHATSOEVER RESULTING FROM LOSS OF USE, DATA OR PROFITS,
+# WHETHER IN AN ACTION OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION,
+# ARISING OUT OF OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS
+# SOFTWARE.
+#
+# For further information see LICENSE
+
+import pytest
+
+# External utility imports
+import os
+import time
+
+# Test framework and local test helper imports
+import testhelper
+from testframework import Container
+
+
+##### Useful globals #####
+
+CURRENT_DIR = os.path.dirname(os.path.realpath(__file__))
+TESTOUTPUT_DIR = CURRENT_DIR + "/testoutput/"
+
+
+##### Provide what the testframework requires #####
+
+# This function is used by the 'agent' fixture to know where the log should be stored
+def logfile_path():
+    return TESTOUTPUT_DIR + "/environment-test.log"
+
+
+##### Local fixtures #####
+
+@pytest.fixture
+def clear_env_files(scope="function"):
+    """ Removes the files used by the helper to avoid having false
+        positives from previous test runs.
+
+        TODO: Should this be a method on the helper?
+    """
+    file_path = TESTOUTPUT_DIR + testhelper.ENV_VARS_FILE_NAME
+    if os.path.exists(file_path):
+        os.remove(file_path)
+
+
+@pytest.fixture
+def create_testoutput_dir(scope="module"):
+    """ Create a directory for the generated test files.
+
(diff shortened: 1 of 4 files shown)


servicetest/environment/testhelper.py
```diff
@@ -0,0 +1,103 @@
...
+def env_var(name, base_path):
+    all_vars = all_env_vars(base_path)
+    return all_vars[name]
+
+
+if __name__ == "__main__":
+    """ TODO: Parsing nicely will be a thing in this helper, make sure to use argparse
+    """
+#    parser = argparse.ArgumentParser()
+#    parser.add_argument("--env-vars", action="store")
```
DunderRoffe: Is this something that will be fixed in a later issue or should this be removed?

servicetest/environment/testhelper.py
```diff
@@ -0,0 +1,103 @@
...
+# WARRANTIES OF MERCHANTABILITY AND FITNESS. IN NO EVENT SHALL THE AUTHOR
+# BE LIABLE FOR ANY SPECIAL, DIRECT, INDIRECT, OR CONSEQUENTIAL DAMAGES
+# OR ANY DAMAGES WHATSOEVER RESULTING FROM LOSS OF USE, DATA OR PROFITS,
+# WHETHER IN AN ACTION OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION,
+# ARISING OUT OF OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS
+# SOFTWARE.
+#
+# For further information see LICENSE
+
+import argparse
```
DunderRoffe: unused import

servicetest/environment/test_environment.py
```diff
@@ -0,0 +1,276 @@
...
+    Container.HOST_PATH: HOST_PATH
+}
+
+
+#####  Various gateway configs used by the tests. #####
+#
+# The configs below might be dependent on each other so beware
+# when changing them.
+
+GW_CONFIG_ONE_VAR = [
```
DunderRoffe, line 92: Are these configs useful for other testcases as well? Otherwise, I think they should be included in the testcase class.

servicetest/environment/test_environment.py
```diff
@@ -0,0 +1,276 @@
...
+# For further information see LICENSE
+
+import pytest
+
+import os
+import time
+
+from testframework import Container
+
+import testhelper
```
DunderRoffe, line 26: Picky: The import statements could be arranged better

servicetest/environment/test_environment.py
```diff
@@ -0,0 +1,276 @@
...
+            sc.launch_command("python " +
+                              sc.get_bind_dir() +
+                              "/testhelper.py --get-env-vars " +
+                              sc.get_bind_dir() + "/testoutput/")
+
+            # The D-Bus LaunchCommand is asynch so let it take effect before assert
+            time.sleep(0.5)
+            variable = testhelper.env_var("LD_LIBRARY_PATH", TESTOUTPUT_DIR)
+            # Assert the append resultet in the creation of the variable
+            assert variable == ":/another/path"
```
frznlogic, line 207: Would be nice with a hint that :/another/path comes from the GW_CONFIG_APPEND var. I guess the hint is there, just not obvious unless you read the whole test ;) .

frznlogic (approved):
Looks good.

servicetest/environment/test_environment.py
breakreturn: I will try to clarify the method comment a bit. I'd rather not explain it again in the comment above the actual assert, as the tests comment itself says that this will happen, the comment above the assert I thought should just clarify what is asserted, not the whole test.

servicetest/environment/testhelper.py
breakreturn: I'll remove the unused things, the import above as well.

servicetest/environment/test_environment.py
breakreturn: Will change.

servicetest/environment/test_environment.py
breakreturn: I think we should add another suite that tests the interaction between the GW configs and variables set explicitly when calling LaunchCommand, and I expect these configs to be reused then.

servicetest/environment/test_environment.py
```diff
@@ -0,0 +1,277 @@
...
+        "value": ":/another/path",
+        "append": True
+    }
+]
+
+
+##### Test suites #####
+
+""" TODO: Create a suite that tests the interaction between GW config and
+          explicitly setting variables when doing LaunchCommand
```
tobsan, line 131: This should be added, I guess.

servicetest/environment/test_environment.py
```diff
@@ -0,0 +1,277 @@
...
+            # The D-Bus LaunchCommand is asynch so let it take effect before assert
+            time.sleep(0.5)
+            ld_library_path = testhelper.env_var("LD_LIBRARY_PATH", TESTOUTPUT_DIR)
+            # Assert the "appended" value is kept
+            assert ld_library_path == ":/another/path"
+        finally:
+            sc.terminate()
+
+    def test_appending_non_existing_var_creates_it(self):
+        """ Append to a non existing env var and assert that the varaible is created
```
tobsan: typo

tobsan (commented):
Looks good in general.

DunderRoffe (approved):

PR #109, opened 2016-10-28
Reading capabilities from disk

thenor:


agent/CMakeLists.txt
```diff
@@ -25,6 +25,8 @@ add_executable(softwarecontainer-agent
     src/softwarecontaineragent.cpp
     src/softwarecontaineragentadaptor.cpp
     ${GENERATED_ADAPTOR_FILENAME}
+    src/capability/configstore.cpp
+    src/capability/capability.cpp
```
tobsan: This class still does nothing.

agent/src/capability/configstore.cpp
```diff
@@ -0,0 +1,237 @@
...
+}
+
+ConfigStore::~ConfigStore()
+{
+}
+
+ReturnCode ConfigStore::readCapsFromFile(const std::string &filePath)
+{
+    if (!is_json(filePath)) {
+        log_debug() << "File is not a json file: \n" << filePath;
```
tobsan: log_ macros will append newlines automatically, and you probably don't want newline before the filePath anyway. This goes for all uses of these log macros.

agent/src/capability/configstore.cpp
```diff
@@ -0,0 +1,237 @@
...
+        }
+    }
+
+    return ReturnCode::SUCCESS;
+}
+
+
+GatewayConfiguration ConfigStore::readConfigs(const std::string capID)
+{
+    return m_configStore[capID];
```
tobsan, line 130: Perhaps a warning about non-existing cap in case the cap is not in the store?

agent/src/capability/configstore.cpp
```diff
@@ -0,0 +1,237 @@
...
+
+
+ReturnCode ConfigStore::parseCapabilities(json_t *capabilities)
+{
+    std::map<std::string, GatewayConfiguration> caps;
+
+    size_t i;
+    json_t *capability;
+
+    json_array_foreach(capabilities,i,capability) {
```
tobsan: Would be nice with spaces here

agent/src/capability/configstore.cpp
```diff
@@ -0,0 +1,237 @@
...
+        if (!json_is_object(capability)) {
+            log_error() << "Capability is not a json object";
+            return ReturnCode::FAILURE;
+        }
+
+        std::string capName;
+        if (!JSONParser::read(capability,"name",capName)) {
```
tobsan: Spaces here also.

agent/src/capability/configstore.cpp
```diff
@@ -0,0 +1,237 @@
...
+        if (gateways == nullptr) {
+            log_error() << "Could not read Gateways";
+            return ReturnCode::FAILURE;
+        }
+        if (!json_is_array(gateways)) {
+            log_error() << "Gateways is not an array";
+            return ReturnCode::FAILURE;
+        }
+
+        return parseGateways(capName,gateways);
```
tobsan: Space please

agent/src/capability/configstore.cpp
```diff
@@ -0,0 +1,237 @@
...
+    }
+    return ReturnCode::SUCCESS;
+}
+
+ReturnCode ConfigStore::parseGateways(std::string capName, json_t *gateways)
+{
+    size_t j;
```
tobsan: Just personal preference, but I would prefer if this was called i.

agent/src/capability/configstore.cpp
```diff
@@ -0,0 +1,237 @@
...
+    GatewayConfiguration gwConfs;
+
+    json_array_foreach(gateways, j, gateway) {
+        if (!json_is_object(gateway)) {
+            log_error() << "Gateway is not a json object";
+            return ReturnCode::FAILURE;
+        }
+
+        std::string gwID;
+        if (!JSONParser::read(gateway,"id",gwID)) {
```
tobsan: Spaces please

agent/src/capability/configstore.cpp
```diff
@@ -0,0 +1,237 @@
...
+            log_error() << "Could not read Gateway ID";
+            return ReturnCode::FAILURE;
+        }
+
+        json_t *confs = json_object_get(gateway,"config");
```
tobsan: Spaces please.

agent/src/capability/configstore.cpp
```diff
@@ -0,0 +1,237 @@
...
+
+    }
+
+    GatewayConfiguration oldGwConfs = m_configStore[capName];
+
+    /* If there is already a Gateway Configuration saved with the
+     * same ID, then append the configs from this file to the
+     * existing Gateway in the m_configstore
+     */
+    for (auto it=gwConfs.begin(); it!=gwConfs.end(); ++it) {
```
tobsan: Using C++11 style for loop would be nicer here:

```
for (auto const &it : gwConfs) {

}
```

agent/src/capability/configstore.cpp
```diff
@@ -0,0 +1,237 @@
...
+{
+    std::vector<std::string> files;
+
+    DIR *dirFile = opendir(filePath.c_str());
+    if (dirFile == nullptr) {
+        log_error() << "Could not open directory: " << filePath << "\n";
+    } else {
+        struct dirent *hFile;
+        while ((hFile = readdir(dirFile)) != nullptr) {
+            // TODO (thenor): we should check for '.' and '..'
```
tobsan: Perhaps this TODO should be removed (since it checks for ".json" later anyway)

agent/src/softwarecontaineragent.cpp
```diff
@@ -31,6 +59,7 @@ SoftwareContainerAgent::~SoftwareContainerAgent()
 
 bool SoftwareContainerAgent::triggerPreload()
 {
+    //log_debug() << "triggerPreload " << m_preloadCount - m_preloadedContainers.size();
```
tobsan: Remove this or have it uncommented.

agent/src/softwarecontaineragent.cpp
```diff
@@ -261,8 +303,38 @@ void SoftwareContainerAgent::setGatewayConfigs(const uint32_t &containerID, cons
...
+
+
+bool SoftwareContainerAgent::setCapabilities(const uint32_t &containerID,
+                                             const std::vector<std::string> &capabilities)
 {
+    for (unsigned int i = 0; i != capabilities.size(); i++) {
+        std::string capId = capabilities[i];
+        GatewayConfiguration newConfigs = m_configStore.readConfigs(capId);
+
+        setGatewayConfigs(containerID,newConfigs);
```
tobsan: Space here, please.

agent/src/softwarecontaineragent.cpp
```diff
@@ -261,8 +303,38 @@ void SoftwareContainerAgent::setGatewayConfigs(const uint32_t &containerID, cons
...
+    }
```
tobsan, line 425: This loop is actually a nice solution so that one doesn't have to merge the configs manually.

agent/src/softwarecontaineragent.cpp
```diff
@@ -261,8 +303,38 @@ void SoftwareContainerAgent::setGatewayConfigs(const uint32_t &containerID, cons
     }
 }
 
-bool SoftwareContainerAgent::setCapabilities(const uint32_t &containerID, const std::vector<std::string> &capabilities)
+void SoftwareContainerAgent::setGatewayConfigs(const uint32_t &containerID,
+                                               const std::map<std::string, std::string> &configs)
+{
+    GatewayConfiguration parsedConfigs;
+
+    for (auto it=configs.begin(); it!=configs.end(); ++it) {
```
tobsan: You could use a c++11 style loop here as well.

agent/src/softwarecontaineragent.cpp
```diff
@@ -261,8 +303,38 @@ void SoftwareContainerAgent::setGatewayConfigs(const uint32_t &containerID, cons
...
+        parsedConfigs[gwID] = jConfigs;
+    }
+    setGatewayConfigs(containerID, parsedConfigs);
+}
+
+
+bool SoftwareContainerAgent::setCapabilities(const uint32_t &containerID,
+                                             const std::vector<std::string> &capabilities)
 {
+    for (unsigned int i = 0; i != capabilities.size(); i++) {
```
tobsan: A range based loop would be nice here as well.

agent/src/softwarecontaineragent.h
```diff
@@ -235,5 +248,6 @@ class SoftwareContainerAgent : protected softwarecontainer::JSONParser
     SignalConnectionsHandler m_connections;
     bool m_shutdownContainers = true;
     std::vector<ContainerID> m_containerIdPool;
+    ConfigStore m_configStore = ConfigStore("");
```
tobsan: This is not very nice, create the m_configStore in the constructor instead.

agent/src/softwarecontaineragentadaptor.cpp
```diff
@@ -9,7 +9,14 @@ softwarecontainer::SoftwareContainerAgentAdaptor::SoftwareContainerAgentAdaptor(
...
 
-uint32_t softwarecontainer::SoftwareContainerAgentAdaptor::LaunchCommand(const uint32_t &containerID, const uint32_t &userID, const std::string &commandLine, const std::string &workingDirectory, const std::string &outputFile, const std::map<std::string, std::string> &env)
+uint32_t softwarecontainer::SoftwareContainerAgentAdaptor::LaunchCommand(
+    const uint32_t &containerID,
+    const uint32_t &userID,
+    const std::string &commandLine,
+    const std::string &workingDirectory,
+    const std::string &outputFile,
+    const std::map<std::string,
+    std::string> &env)
```
tobsan: Don't put one argument on two lines, please.

agent/unit-test/softwarecontaineragent_unittest.cpp
```diff
@@ -82,37 +82,45 @@ TEST_F(SoftwareContainerAgentTest, CreateContainerWithConf) {
...
+    // How to get the configurations in configStore
+    ContainerID id = sca->createContainer("confStoreTest-", "");
+    std::vector<std::string> caps;
+    caps.push_back("com.acme.A");
+    caps.push_back("com.acme.B");
+    caps.push_back("com.acme.C");
 
-    ASSERT_TRUE(retval == true);
-    ASSERT_TRUE(workspace->m_enableWriteBuffer == false);
+    // TODO (thenor): Most likely gives a false positive. Find a way to fix this.
```
tobsan: Yes, please.

tobsan (changes requested):

tobsan:
Also please rebase against master

agent/src/capability/configstore.cpp
thenor: Sounds very reasonable.

agent/src/capability/configstore.cpp
thenor: Forgot this when I made it into a help method.

agent/src/capability/configstore.h
```diff
@@ -0,0 +1,79 @@
...
+
+        ReturnCode parseCapabilities(json_t *capabilities);
+        ReturnCode parseGateways(std::string capName, json_t *gateways);
+
+        std::vector<std::string> fileList(const std::string &filePath);
+        bool is_json(const std::string &filename);
+
+        std::map<std::string, GatewayConfiguration> m_configStore;
+    };
+}
```
breakreturn, line 120: We usually don't indent the namespace, unindent the class level.

agent/src/capability/configstore.h
```diff
@@ -0,0 +1,79 @@
...
+         * @return ReturnCode::SUCCESS if successful
+         * @return ReturnCode::FAILURE if parsing of files is unsuccessful
+         *
+         */
+        ReturnCode readCapsFromDir(const std::string &dirPath);
+
+        /**
+         * @brief Returns all Gateways for a certain Capability.
+         */
+        GatewayConfiguration readConfigs(const std::string capID);
```
breakreturn: I think the above comments should be rewritten to not use the word "gateway" like it is currently. A "Gateway" is a specific thing, a "gateway configuration" is different specific thing.

agent/src/main.cpp
```diff
@@ -10,12 +29,13 @@ void usage(const char *argv0)
...
     printf("SoftwareContainer agent, v.%s\n", PACKAGE_VERSION);
     printf("Usage: %s [-p or --preload <num>] [-u or --user <uid>]", argv0);
-    printf("[-s or --shutdown <bool>] [-t or --timeout <seconds>]\n");
+    printf("[-s or --shutdown <bool>] [-t or --timeout <seconds>] [-c or --configPath <path>]\n");
     printf("\n");
     printf("--preload <num>     : Number of containers to preload, defaults to 0\n");
     printf("--user <uid>        : Default user id to be used when starting processes in the container, defaults to 0\n");
     printf("--shutdown <bool>   : If false, containers will not be shutdown on exit. Useful for debugging. Defaults to true\n");
     printf("--timeout <seconds> : Timeout in seconds to wait for containers to shutdown, defaults to 2\n");
+    printf("--configPath <path> : Path to a file or directory where Service Manifests exist, defaults to \"\"");
```
breakreturn: We might very well add a "SoftwareContainer configuration" concept later, so I think we should use an option name that is more specific for this. Like, service manifest path or something like that.

agent/src/softwarecontaineragent.cpp
```diff
@@ -1,14 +1,34 @@
...
-, m_preloadCount(preloadCount)
-, m_shutdownContainers(shutdownContainers)
+    Glib::RefPtr<Glib::MainContext> mainLoopContext,
+    int preloadCount,
+    bool shutdownContainers,
+    int shutdownTimeout,
+    const std::string &configPath)
+    : m_mainLoopContext(mainLoopContext)
+    , m_preloadCount(preloadCount)
+    , m_shutdownContainers(shutdownContainers)
```
breakreturn, line 31: This indentation makes it difficult to see the difference between the parameters and the initialization of members.

agent/src/softwarecontaineragent.cpp
```diff
@@ -261,8 +303,38 @@ void SoftwareContainerAgent::setGatewayConfigs(const uint32_t &containerID, cons
     }
 }
 
-bool SoftwareContainerAgent::setCapabilities(const uint32_t &containerID, const std::vector<std::string> &capabilities)
+void SoftwareContainerAgent::setGatewayConfigs(const uint32_t &containerID,
+                                               const std::map<std::string, std::string> &configs)
```
breakreturn, line 378: Some line break inconsistency here.

agent/src/capability/configstore.cpp
```diff
@@ -0,0 +1,237 @@
...
+        log_debug() << "File is not a json file: \n" << filePath;
+        return ReturnCode::FAILURE;
+    }
+
+    json_error_t error;
+    json_t *fileroot = json_load_file(filePath.c_str(), 0, &error);
+
+    if (fileroot == nullptr) {
+        log_error() << "Could not parse Service Manifest: \n" <<
+            filePath << ":" << error.line <<"\n" << error.text;
```
DunderRoffe: The `<<` operator should be before `filePath` not last on the line above.

agent/src/softwarecontaineragent.cpp
```diff
@@ -261,8 +303,38 @@ void SoftwareContainerAgent::setGatewayConfigs(const uint32_t &containerID, cons
...
+{
+    GatewayConfiguration parsedConfigs;
+
+    for (auto it=configs.begin(); it!=configs.end(); ++it) {
+        json_error_t error;
+        std::string gwID = it->first;
+        std::string configStr = it->second;
+
+        json_t *jConfigs = json_loads(configStr.c_str(),0,&error);
```
breakreturn: Put spaces between args

agent/src/softwarecontaineragent.cpp
```diff
@@ -261,8 +303,38 @@ void SoftwareContainerAgent::setGatewayConfigs(const uint32_t &containerID, cons
     }
 }
 
-bool SoftwareContainerAgent::setCapabilities(const uint32_t &containerID, const std::vector<std::string> &capabilities)
+void SoftwareContainerAgent::setGatewayConfigs(const uint32_t &containerID,
```
breakreturn: Perhaps we should return a bool from this as we do with setCapabilities?

agent/src/capability/configstore.cpp
```diff
@@ -0,0 +1,237 @@
...
+        }
+        if (!json_is_array(confs)) {
+            log_error() << "Gateway Config is not an array";
+            return ReturnCode::FAILURE;
+        }
+        gwConfs[gwID] = confs;
+
+    }
+
+    GatewayConfiguration oldGwConfs = m_configStore[capName];
```
DunderRoffe, line 201: Maybe return `ReturnCode::FAILURE` if there is no conf for `capName` in `m_configStore`

agent/src/capability/configstore.cpp
```diff
@@ -0,0 +1,237 @@
...
+    return ReturnCode::SUCCESS;
+}
+
+std::vector<std::string> ConfigStore::fileList(const std::string &filePath)
+{
+    std::vector<std::string> files;
+
+    DIR *dirFile = opendir(filePath.c_str());
+    if (dirFile == nullptr) {
+        log_error() << "Could not open directory: " << filePath << "\n";
```
DunderRoffe: Return empty `std::vector< ... >` here. Then we do not need the else clause and can lower the indentation.

agent/unit-test/configstore_unittest.cpp
```diff
@@ -0,0 +1,119 @@
...
+    std::string capNameB = "com.pelagicore.temperatureservice.settemperature";
+    std::string capNameC = "dummyCapC";
+    std::string capNameD = "dummyCapD";
+
+    /* The Service Manifest (absolute) file paths */
+    std::string homeDir       = "/home/vagrant/softwarecontainer/agent/unit-test/";
+    std::string sm_path       = "CS_unittest_ServiceManifest.json";
+    std::string short_sm_path = "CS_unittest_short_ServiceManifest.json";
+    std::string evil_sm       = "CS_unittest_parseError.json";
+    std::string inputDir      = "/home/vagrant/";
```
breakreturn: We need to have some way of not having the paths absolute to the vagrant machine.

agent/unit-test/configstore_unittest.cpp
```diff
@@ -0,0 +1,119 @@
...
+TEST_F(ConfigStoreTest, readConfigFetchEvilCap) {
+    ASSERT_FALSE(isError(cs->readCapsFromFile(homeDir+sm_path)));
+
+    GatewayConfiguration retGWs = cs->readConfigs("EvilCapName");
+    ASSERT_TRUE(retGWs.empty());
+}
+
+TEST_F(ConfigStoreTest, readEvilConfig) {
+    ASSERT_TRUE(isError(cs->readCapsFromFile(homeDir+evil_sm)));
+}
```
breakreturn, line 168: In general: add comments of what is tested, i.e. why would some calls be false etc. Also add spaces around operators like '+'

agent/src/capability/configstore.cpp
thenor: Will fix. Update in next upload.

agent/CMakeLists.txt
thenor: I know. Kept partly to start discussion on if I should use it. If no reason pops up for using it, I will remove it.

agent/src/softwarecontaineragent.cpp
```diff
@@ -261,8 +303,38 @@ void SoftwareContainerAgent::setGatewayConfigs(const uint32_t &containerID, cons
...
+{
+    GatewayConfiguration parsedConfigs;
+
+    for (auto it=configs.begin(); it!=configs.end(); ++it) {
+        json_error_t error;
+        std::string gwID = it->first;
+        std::string configStr = it->second;
+
+        json_t *jConfigs = json_loads(configStr.c_str(),0,&error);
+        if (!jConfigs) {
```
DunderRoffe, line 389: You alternate between using `!ptr` and `ptr == nullptr` please choose one style.

agent/unit-test/configstore_unittest.cpp
```diff
@@ -0,0 +1,119 @@
...
+#include <gmock/gmock.h>
+#include <gtest/gtest.h>
+#include <unistd.h>
+
+class ConfigStoreTest: public ::testing::Test
+{
+public:
+    LOG_DECLARE_CLASS_CONTEXT("TEST", "Tester");
+    ConfigStoreTest() { }
+    ConfigStore *cs = new ConfigStore("");
```
DunderRoffe: This does not seem to be deleted anywhere.

common/jsonparser.cpp
```diff
@@ -24,7 +24,8 @@ namespace softwarecontainer {
 bool JSONParser::read(const json_t *element, const char *key, std::string &result)
 {
     json_t *value = json_object_get(element, key);
-    if (!value) {
+    if (value == nullptr) {
```
breakreturn: If this is a better check, should the other similar checks below be changed too?

DunderRoffe (changes requested):
Overall good. Found a few things, a bit picky but still worth considering.

agent/src/softwarecontaineragent.cpp
```diff
@@ -252,7 +293,8 @@ std::string SoftwareContainerAgent::bindMountFolderInContainer(const uint32_t co
     return "";
 }
 
-void SoftwareContainerAgent::setGatewayConfigs(const uint32_t &containerID, const std::map<std::string, std::string> &configs)
+void SoftwareContainerAgent::setGatewayConfigs(const uint32_t &containerID,
+                                               const GatewayConfiguration &configs)
```
breakreturn: As far as I can see one method "setGatewayConfigs" is used by the user and the other overloaded version is not intended to be used by the user, rather by the other method. In this case one of them should be private, and I suggest we change the name and not only overload. Unless we plan for the user to be able to use both? I wouldn't be a fan of that either unless there's a good reason though.

agent/src/capability/configstore.h
```diff
@@ -0,0 +1,79 @@
...
+
+    private:
+
+        ReturnCode parseCapabilities(json_t *capabilities);
+        ReturnCode parseGateways(std::string capName, json_t *gateways);
```
breakreturn: We parse "gateway configs" and not "gateways", or am I missing the point?

agent/src/capability/configstore.h
```diff
@@ -0,0 +1,79 @@
...
+         * @brief Reads a directory and finds all json Service Manifest files
+         * and adds the capabilities to the internal storage.
+         * If a capability with the same name already exists in the storage
+         * the gateways will be appended to the previously stored gatway list.
+         *
+         * @return ReturnCode::SUCCESS if successful
+         * @return ReturnCode::FAILURE if parsing of files is unsuccessful
+         *
+         */
+        ReturnCode readCapsFromDir(const std::string &dirPath);
```
breakreturn: What's the intention for the user of this class? In this design, should anything apart from getting the actual gateway configs by cap ID be public?

agent/src/capability/configstore.h
breakreturn: We don't really do any "reading" at this stage, perhaps this method name should be changed so it doesn't accidentally imply implementation details.

common/softwarecontainer-common.h
```diff
@@ -40,7 +42,7 @@ static constexpr ContainerID INVALID_CONTAINER_ID = -1;
 static constexpr pid_t INVALID_PID = -1;
 static constexpr int INVALID_FD = -1;
 
-typedef std::map<std::string, std::string> GatewayConfiguration;
+typedef std::map<std::string, json_t *> GatewayConfiguration;
```
breakreturn, line 45: I still think this should be a class so we can hide raw types from external dependencies more easily.

agent/src/capability/configstore.cpp
thenor: Done.

agent/src/capability/configstore.cpp
thenor: Done.

agent/src/capability/configstore.cpp
thenor: Done.

agent/src/capability/configstore.cpp
thenor: Done.

agent/src/capability/configstore.cpp
thenor: Done.

agent/src/capability/configstore.cpp
thenor: Done.

agent/src/capability/configstore.cpp
thenor: Done.

agent/src/capability/configstore.cpp
thenor: This method is also used when parsing a file from the constructor, when the m_configstore is empty, so that would mean that it fails as soon as we try to add a new capability.

tobsan: Isn't the point here to **not fail** even if there are no old configs. The [] operator on std::map constructs an object of the value type if there is none available. I think it can be kept as is.

If there is a nice way to make it explicit if there are any old configs or not without making it too complicated, that could also be nice of course.

agent/unit-test/CS_unittest_ServiceManifest.json
```diff
@@ -0,0 +1,80 @@
+{
```
tobsan: Could we place this in some test-data directory instead?

agent/CMakeLists.txt
thenor: Removed in next upload.

agent/src/capability/capability.cpp
```diff
@@ -0,0 +1,32 @@
+
```
thenor: Removed.

agent/src/capability/capability.h
```diff
@@ -0,0 +1,44 @@
+
```
thenor: Removed.

agent/src/capability/configstore.cpp
thenor: Looks magical. Done.

agent/src/capability/configstore.cpp
thenor: Cool. Done.

agent/src/capability/configstore.cpp
thenor: This was in part a reminder about looking in to how to handle ".." or ".". How do you think they should be handled?
I think other checks will make this fail or not fail anyway. Was just thinking if I should handle this as "/" where I don't allow it.

agent/src/capability/configstore.h
thenor: The reason I made reading from file/dir public was that I thought perhaps users would want to start up the container and agent first, then read in the configuration. Perhaps not a valid use case?

agent/src/capability/configstore.h
thenor: Oh yes. Of course. Done.

thenor: Done.

agent/src/capability/configstore.h
thenor: This method actually parses the gateways in the json file, so in this case it is actually a "gateways" and not gateway configurations.

thenor: I will change the name.

agent/src/capability/configstore.h
thenor: Done.

agent/src/main.cpp
thenor: Done.

agent/src/capability/configstore.cpp
tobsan: Currently there is a function that checks if the file name ends with ".json". The special files "." and ".." does not end with ".json", so this shouldn't be a problem.

agent/src/softwarecontaineragent.cpp
thenor: Agreed. Done.

agent/src/softwarecontaineragent.cpp
thenor: Done. Mistake during pull from upstream.

agent/src/softwarecontaineragent.cpp
thenor: Yes, I wasn't sure about this, but I agree. What does @tobsan say about this change?

agent/src/softwarecontaineragent.cpp
thenor: Yes, I had a thought about this also. Does @tobsan agree?

agent/src/softwarecontaineragent.cpp
thenor: Actually not since it's a map of strings.

agent/src/softwarecontaineragent.cpp
thenor: Yes. Done.

agent/src/softwarecontaineragent.cpp
tobsan: Yes, that makes sense.

agent/src/softwarecontaineragent.cpp
thenor: Is there any case where !ptr could be a bad syntax? In my opinion it's cleaner since it's shorter. Any opinion to the contrary, @DunderRoffe ?

agent/src/softwarecontaineragent.cpp
tobsan: I agree, make the internal one private (the other one is used by the DBus API)

agent/src/softwarecontaineragent.cpp
thenor: I have done this, but feel like it may require even more changes to the softwarecontainer class, so let's see if this is sufficient. Changes in next upload.

agent/src/softwarecontaineragent.cpp
thenor: Done.

agent/src/softwarecontaineragent.cpp
thenor: Done.

agent/src/softwarecontaineragent.cpp
thenor: Done.

agent/src/softwarecontaineragent.cpp
thenor: Done.

agent/src/softwarecontaineragent.h
thenor: I agree, but couldn't get it to compile then. I will grab you to see if you can see a solution.

agent/src/softwarecontaineragentadaptor.cpp
thenor: Oops. Done.

agent/unit-test/CS_unittest_ServiceManifest.json
thenor: Done.

agent/src/softwarecontaineragent.h
thenor: Done.

agent/unit-test/configstore_unittest.cpp
thenor: I thought about this also, but I wasn't sure how. If I had tested this from e.g. the Agent it would have felt more intuitive to test the default container paths also. More tests are needed.

agent/src/softwarecontaineragent.cpp
DunderRoffe: No there are no functional difference between the two as far as I am aware.

agent/unit-test/configstore_unittest.cpp
thenor: Done.

agent/unit-test/configstore_unittest.cpp
thenor: Done.

common/jsonparser.cpp
thenor: DunderRoffe also commented on this. I will change the "ptr == nullptr" to "!ptr".

agent/unit-test/softwarecontaineragent_unittest.cpp
thenor: Done. Hopefully....

common/softwarecontainer-common.h
thenor: I will add that to my todo list. I think it's a good idea.

agent/src/softwarecontaineragent.cpp
thenor: I will use "!ptr". Looks nicer.

agent/src/capability/configstore.cpp
```diff
@@ -0,0 +1,242 @@
...
+
+    std::vector<std::string> files;
+    files = fileList(dirPath);
+
+    if (files.empty()) {
+        log_info() << "No configuration files found: " << dirPath;
+        return ReturnCode::SUCCESS;
+    }
+
+    for (unsigned int i = 0; i != files.size(); i++) {
```
tobsan: You can use a range-loop here.

agent/src/softwarecontaineragent.cpp
```diff
@@ -277,17 +315,54 @@ std::string SoftwareContainerAgent::bindMountFolderInContainer(const uint32_t co
...
 
-bool SoftwareContainerAgent::setCapabilities(const uint32_t &containerID, const std::vector<std::string> &capabilities)
+
+bool SoftwareContainerAgent::setCapabilities(const uint32_t &containerID,
+                                             const std::vector<std::string> &capabilities)
 {
+    for (std::string capId : capabilities) {
+        GatewayConfiguration newConfigs = m_configStore.getGatewayConfigs(capId);
+
+        if (!updateGatewayConfigs(containerID, newConfigs)) {
```
tobsan, line 420: An error log would be nice here.

agent/unit-test/configstore_unittest.cpp
```diff
@@ -0,0 +1,172 @@
...
+#include <gmock/gmock.h>
+#include <gtest/gtest.h>
+#include <unistd.h>
+
+class ConfigStoreTest: public ::testing::Test
+{
+public:
+    LOG_DECLARE_CLASS_CONTEXT("TEST", "Tester");
+    ConfigStoreTest() { }
+    ConfigStore *cs = new ConfigStore();
```
tobsan: Couldn't this be created on the stack instead, to avoid having to do delete in the teardown?

agent/unit-test/softwarecontaineragent_unittest.cpp
```diff
@@ -36,18 +35,20 @@ class SoftwareContainerAgentTest: public ::testing::Test
     bool m_shutdownContainers = true;
     int m_shutdownTimeout = 2;
     std::shared_ptr<Workspace> workspace;
+    std::string m_configPath =
+        "/home/vagrant/softwarecontainer/agent/unit-test/test-data/CS_unittest_ServiceManifest.json";
```
tobsan: Make this a relative path.

tobsan (changes requested):

agent/src/capability/configstore.cpp
thenor: Done.

agent/src/capability/configstore.cpp
thenor: I will add this as an improvement note, but unless you feel I need to change something now, I will not change this.

agent/src/softwarecontaineragent.cpp
thenor: Perhaps a "return" here may have been wrong.
Do we want to fail if we can't set the configuration for one capability?

agent/unit-test/configstore_unittest.cpp
thenor: Is this what @DunderRoffe did for the agent unittest? I can look at that code and do that here also.

agent/unit-test/softwarecontaineragent_unittest.cpp
thenor: I thought that would be the bad way to go. Why is a relative path better? So that I can move the test case and the test-data directory? I was more concerned that the tests could be started from "not here" and the relative path would point incorrectly.

agent/src/softwarecontaineragent.cpp
tobsan: We are doing it differently in different parts of the code actually. I think it should be a failure if we can't apply our configs.

agent/unit-test/softwarecontaineragent_unittest.cpp
tobsan: It will be equally bad if someone doesn't build softwarecontainer in vagrant, or in some other path in the VM. We can't assume everyone will build in a VM. You should be able to point to the tests relative to the test binary. It is reasonable to assume that test-data directory is available in the same place as the test binary.

thenor: Alright. Will change it.

agent/src/softwarecontaineragent.cpp
thenor: Done.

agent/src/capability/configstore.h
```diff
@@ -0,0 +1,111 @@
...
+class ConfigStore
+{
+    LOG_DECLARE_CLASS_CONTEXT("CS", "ConfigStore");
+
+public:
+    ConfigStore();
+    ConfigStore(const std::string &filePath);
+    ~ConfigStore();
+
+    /**
```
thenor, line 49: After discussions with @breakreturn I will make these two read methods private instead. They are not relevant to have public for any current use case.

agent/src/capability/configstore.cpp
```diff
@@ -0,0 +1,250 @@
...
+}
+
+ConfigStore::ConfigStore(const std::string &inputPath)
+{
+    ReturnCode retval = ReturnCode::SUCCESS;
+
+    if (inputPath.empty()) {
+        //File path is empty, this is ok
+        return;
+    } else if (isDirectory(inputPath)) {
```
DunderRoffe: In the cases were we return in an if statement, I think it is nice to not use an else clause. In cases such as this, consider using another if statement. This makes it separated and more clear that the function (or constructor in this case) terminates.

agent/src/capability/configstore.cpp
```diff
@@ -0,0 +1,250 @@
...
+        log_warning() << "Searching for configuration files from root dir not allowed";
+        return ReturnCode::FAILURE;
+    }
+    if (isDirectoryEmpty(dirPath)) {
+        log_warning() << "Path to configuration files is empty: " << dirPath;
+        return ReturnCode::FAILURE;
+    }
+
+    std::vector<std::string> files;
+    files = fileList(dirPath);
```
tobsan: This can be just one line instead.

agent/unit-test/configstore_unittest.cpp
DunderRoffe: What @tobsan means is that it should not be a pointer.

DunderRoffe (commented):
Looks good overall.
There are some cases where you need to change to use  "== nullptr" checks.

libsoftwarecontainer/src/softwarecontainer.cpp
```diff
@@ -163,8 +157,9 @@ void SoftwareContainer::setGatewayConfigs(const GatewayConfiguration &configs)
         std::string gatewayId = gateway->id();
 
         if (configs.count(gatewayId) != 0) {
-            std::string config = configs.at(gatewayId);
-            gateway->setConfig(config);
+            json_t *config = configs.at(gatewayId);
+            char *configStr = json_dumps(config,0);
```
DunderRoffe, line 161: All that is returned by json_dumps needs to be freed

agent/unit-test/configstore_unittest.cpp
thenor: Done.

libsoftwarecontainer/src/softwarecontainer.cpp
thenor: Done.

agent/src/capability/configstore.cpp
thenor: Done.

agent/src/capability/configstore.cpp
thenor: Done.

agent/unit-test/configstore_unittest.cpp
```diff
@@ -0,0 +1,172 @@
...
+    std::string capNameA = "com.pelagicore.temperatureservice.gettemperature";
+    std::string capNameB = "com.pelagicore.temperatureservice.settemperature";
+    std::string capNameC = "dummyCapC";
+    std::string capNameD = "dummyCapD";
+
+    /* The Service Manifest (absolute) file paths */
+    std::string homeDir       = "/home/vagrant/softwarecontainer/agent/unit-test/test-data/";
```
thenor: I need to make this into a relative path as discussed.

thenor: Done.

DunderRoffe (approved):

tobsan (approved):

thenor:
One new unit test case fails in Jenkins, but didn't fail in my vagrant, so I will try to reproduce this locally.

tobsan (approved):

PR #110, opened 2016-10-31
Separate targets for building API and User docs

DunderRoffe:
Solved by splitting the "doc" target into two separate targets named "api-doc" and "user-doc".


CMakeLists.txt
```diff
@@ -99,7 +99,8 @@ option(ENABLE_TEST "Enables unit testing" ON)
 option(ENABLE_PROFILING "Enables profiling support in the application" OFF)
 option(ENABLE_COVERAGE "Enable building and installation of examples" OFF)
 option(ENABLE_EXAMPLES "Enable building and installation of examples" OFF)
-option(ENABLE_DOC "Enable building and installation of documentation" OFF)
+option(ENABLE_USER_DOC "Enable building and installation of the user documentation" OFF)
+option(ENABLE_API_DOC "Enable building and installation of the api documentation" OFF)
```
tobsan, line 106: It would be nice to have an overall option still to build all docs. CMake has a module for setting dependencies between options.

https://cmake.org/cmake/help/v3.0/module/CMakeDependentOption.html

tobsan (commented):
Looks good in general.

CMakeLists.txt
tobsan: Actually, I would probably still want to guard the doc/ directory with ENABLE_DOC and then the USER_DOC and API_DOC can set that if they are set. If not, ENABLE_DOC should build both targets.

breakreturn (approved):
I agree with Tobias about an option that enables all docs, otherwise it looks good to me.

CMakeLists.txt
DunderRoffe: This would require that we bump the CMake version. Currently we use 2.6 which is really old and does not seem to support the `cmake_dependant_option` macro.

I agree that it would add readability by keeping the doc guard. I feel that we should maybe try to split the CMake files for the user doc and api doc as well. The problem with that is the doxygen will require a separate folder which might be a bit much for 2 files.

tobsan: - When was the dependent option first included in cmake? If we can't bump to that for legacy reasons, then perhaps we can set the USE_DOC manually if API_DOC or USER_DOC is set?
- You can still split into two files, just not two CMakeLists.txt. Call them doxygen.cmake and sphinx.cmake or something like that.

PR #113, opened 2016-10-31
docs: adds examples, integrating wayland

tobsan:
Tells the integrator how to integrate wayland. Uses wayland example to
illustrate.

### DIFF
--- a/doc/chapters/integration-guidelines/index.rst
+++ b/doc/chapters/integration-guidelines/index.rst
@@ -45,3 +45,34 @@ give access to graphics hardware. Not all applications require direct access to 
 hardware, see :ref:`Wayland example <wayland-example>`. A reasonable capability for a Wayland
 application would therefore include both the Wayland gateway and a configuration of the Device Node
 gateway for any graphics hardware access needed.
+
+Example
+-------
+Here is an example manifest defining wayland access::
+
+    {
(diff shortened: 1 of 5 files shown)


doc/chapters/integration-guidelines/index.rst
```diff
@@ -36,3 +36,12 @@ create such a bridge on the system if it is not already there. This can be chang
...
 The selection of whether or not to create the bridge is a compile-time options given to CMake.
+
+Wayland setup
+=============
+
+In order to have applications access wayland, one needs to enable the Wayland gateway, and possibly
+give access to graphics hardware. Not all applications require direct access to the graphics
+hardware, see :ref:`Wayland example <wayland-example>`. A reasonable capability for a Wayland
+application would therefore include both the Wayland gateway and a configuration of the Device Node
+gateway for any graphics hardware access needed.
```
breakreturn, line 47: What do you think about having an actual example of a complete service manifest here? A "Wayland Capability" is not really tied to a service on the system perhaps so even if it's a bit of a special case with Wayland, it would still take this paragraph from theoretical to more practical. Or do we want to only rely on the example  in this case?

tobsan: Yes, that would make sense. I will add one!

PR #114, opened 2016-11-01
Fixed faulty doxygen file tree generation

DunderRoffe:


doc/api_doc.cmake
```diff
@@ -17,41 +17,14 @@
 
 find_package(Doxygen REQUIRED)
 
-set(packageName "softwarecontainer")
+set(packageName "SoftwareContainer")
```
tobsan: You can use the PROJECT_NAME variable here instead, available in cmake.

tobsan (changes requested):

tobsan (approved):

PR #116, opened 2016-11-01
Error handling for createContainer in agent.

DunderRoffe:
createContainer now returns -1 when something fails.

### DIFF
--- a/agent/src/softwarecontaineragent.cpp
+++ b/agent/src/softwarecontaineragent.cpp
@@ -33,13 +33,13 @@ bool SoftwareContainerAgent::triggerPreload()
 {
     while (m_preloadedContainers.size() < m_preloadCount) {
         auto availableID = findSuitableId();
-        auto container = makeSoftwareContainer(availableID);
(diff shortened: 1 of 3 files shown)


DunderRoffe:
I  am thinking about changing the return type of `parseConfig` from `bool` to `ReturnCode`. Do you think that would be meaningful?

agent/src/softwarecontaineragent.cpp
```diff
@@ -162,9 +166,15 @@ ContainerID SoftwareContainerAgent::createContainer(const std::string &config)
 
     m_containers[availableID] = SoftwareContainerPtr(container);
     container->setMainLoopContext(m_mainLoopContext);
-    container->init();
 
-    triggerPreload();
+    if (isError(container->init()) {
+        log_error() << "Could not init the container.";
+        return -1;
```
tobsan, line 159: Magic values are not really nice. Can't you throw an exception instead?

tobsan (changes requested):

tobsan (approved):

PR #121, opened 2016-11-02
Changed method scope in gateway base class

breakreturn:
Made methods only used or implemented by inheriting
classes protected. Also updated/added some comments.

### DIFF
--- a/libsoftwarecontainer/include/gateway.h
+++ b/libsoftwarecontainer/include/gateway.h
@@ -19,16 +19,14 @@
  */
 
 
-#ifndef GATEWAY_H
-#define GATEWAY_H
+#pragma once
+
+#include <string>
 
 #include "containerabstractinterface.h"
 #include "jsonparser.h"
 
-#include <string>
-
 /**
- *
  * @brief Gateway base class for SoftwareContainer
  *
  * Gateways can be in one of three states:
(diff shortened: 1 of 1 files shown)


frznlogic (approved):

libsoftwarecontainer/include/gateway.h
```diff
@@ -107,30 +93,51 @@ class Gateway: protected softwarecontainer::JSONParser
...
+     * can access a ContainerAbstractInterface instance previously set by SoftwareContainer
+     * calling setContainer()
+     *
+     * @returns true if there is a ContainerAbstractInterface instance set, false if not.
      */
-    bool isActivated();
+    bool hasContainer();
 
+    /**
+     * @brief Get a reference to the associated container
```
DunderRoffe: I get why you do not write "shared pointer" but "reference" is a reserved word with a very specific meaning in C++. Perhaps a better word can be found?

libsoftwarecontainer/include/gateway.h
```diff
@@ -19,16 +19,14 @@
  */
 
 
-#ifndef GATEWAY_H
-#define GATEWAY_H
+#pragma once
+
+#include <string>
 
 #include "containerabstractinterface.h"
```
DunderRoffe, line 26: Would be nice to use forward declarations instead

tobsan (approved):

libsoftwarecontainer/include/gateway.h
breakreturn: "Get a handle.."?

PR #122, opened 2016-11-02
Checking if LXC ID already in use. Also cleanup

DunderRoffe:


libsoftwarecontainer/unit-test/softwarecontainerlib_unittest.cpp
```diff
@@ -88,6 +89,21 @@ TEST_F(SoftwareContainerApp, TestWayland) {
 
 }
 
+TEST_F(SoftwareContainerApp, DoubleIDCreatesError) {
```
frznlogic, line 95: We are bad at commenting/documenting our tests, but I think a few lines documentation would be nice here, why are we doign this test and what's the expected behavior.

libsoftwarecontainer/include/softwarecontainer.h
```diff
@@ -103,7 +103,7 @@ class SoftwareContainer :
 
     std::shared_ptr<ContainerAbstractInterface> m_container;
     pid_t m_pcPid = INVALID_PID;
-    std::vector<std::unique_ptr<Gateway> > m_gateways;
+    std::vector< std::unique_ptr<Gateway> > m_gateways;
```
breakreturn: I think we should not use spaces like this, the original space at the end was incorrect.

tobsan (changes requested):
1) It would make sense to set the state to some dead state in rollbackCreate, as to avoid doing the same error several times.
2) How does the agent handle this type of error?

libsoftwarecontainer/src/container.cpp
```diff
@@ -146,64 +145,67 @@ ReturnCode Container::create()
...
-        }
+    if (m_enableWriteBuffer) {
+        createDirectory(rootfspath_dst);
+        const std::string rootfspath_lower = rootfspath_dst + "-lower";
+        createDirectory(rootfspath_lower);
+        const std::string rootfspath_upper = rootfspath_dst + "-upper";
+        createDirectory(rootfspath_upper);
+        const std::string rootfspath_work = rootfspath_dst + "-work";
+        overlayMount(rootfspath_lower, rootfspath_upper, rootfspath_work, rootfspath_dst);
+        m_rootFSPath = rootfspath_dst;
```
breakreturn: It was like that before, but this is not the variable name style we use.

libsoftwarecontainer/unit-test/softwarecontainerlib_unittest.cpp
breakreturn, line 95: Document what the test is about

libsoftwarecontainer/include/softwarecontainer.h
DunderRoffe: ok

libsoftwarecontainer/src/container.cpp
```diff
@@ -146,64 +145,67 @@ ReturnCode Container::create()
...
 
-    return status;
+    m_state = ContainerState::CREATED;
+    log_debug() << "Container created. RootFS: " << m_rootFSPath;
+
+    return ReturnCode::SUCCESS;
+}
+
+ReturnCode Container::rollbackCreate() {
+    lxc_container_put(m_container);
```
frznlogic: Error handling of lxc_container_put? not sure if we can do any

frznlogic (changes requested):

DunderRoffe:
Answers to the questions asked by @tobsan :
1. I Added a null check in the rollback function. More than that is not really possible from inside the Container class or from the SoftwareContainer class since the ID is set **only** in their constructors.
2. The agent handles this by destroying containers that fail to init. This wont really be a problem though as the agent knows which IDs that are already taken.

breakreturn (approved):

tobsan (approved):

PR #123, opened 2016-11-02
Changed behavior of env vars precedence

breakreturn:
Env vars set through LaunchCommand will now
take precedence over varaiables set through
the env gateway config.

### DIFF
--- a/libsoftwarecontainer/src/container.cpp
+++ b/libsoftwarecontainer/src/container.cpp
@@ -280,8 +280,13 @@ int Container::executeInContainerEntryFunction(void *param)
     return (*function)();
 }
 
-ReturnCode Container::executeInContainer(ContainerFunction function, pid_t *pid, const EnvironmentVariables &variables, uid_t userID,
-                                         int stdin, int stdout, int stderr)
+ReturnCode Container::executeInContainer(ContainerFunction function,
+                                         pid_t *pid,
+                                         const EnvironmentVariables &variables,
+                                         uid_t userID,
+                                         int stdin,
+                                         int stdout,
+                                         int stderr)
 {
     if (pid == nullptr) {
         log_error() << "Supplied pid argument is nullptr";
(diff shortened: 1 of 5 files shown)


thenor (approved):

tobsan (approved):

libsoftwarecontainer/src/container.cpp
```diff
@@ -301,18 +306,26 @@ ReturnCode Container::executeInContainer(ContainerFunction function, pid_t *pid,
...
-                log_warning() << "Variable set by gateway overriding original variable value: "
-                              << var.first << ". values: " << var.second;
+                // Inform user that GW config will be overridden, it might be unintentionally done
+                log_info() << "Variable \""
+                           << var.first
+                           << "\" set by gateway will be overwritten with the value: \""
+                           << variables.at(var.first)
+                           << "\"";
             }
+            actualVariables[var.first] = variables.at(var.first);
```
DunderRoffe, line 324: Is there any specific reason why use two different styles of accessing values from a map?

libsoftwarecontainer/src/gateway/envgateway.cpp
```diff
@@ -62,6 +62,16 @@ ReturnCode EnvironmentGateway::readConfigElement(const json_t *element)
 
     if (m_variables.count(variableName) == 0) {
         m_variables[variableName] = variableValue;
+        if (appendMode) {
+            // The variable did not exist but the config intention was still to append,
+            // this might be a misconfiguration
+            log_info() << "Env variable \""
+                       << variableName
+                       << "\" was configured to be appended, but will instead be created. "
```
DunderRoffe: Please add the reason in the log output

DunderRoffe (commented):
Looks nice, but I have two questions:

libsoftwarecontainer/src/container.cpp
breakreturn: The one is const and the other not.

libsoftwarecontainer/src/gateway/envgateway.cpp
breakreturn: Will do.

PR #124, opened 2016-11-02
Makes it possible to bind-mount read-only

tobsan:
Bind-mounting read-only into the container now works, by re-mounting
from inside the container.

### DIFF
--- a/libsoftwarecontainer/src/container.cpp
+++ b/libsoftwarecontainer/src/container.cpp
@@ -588,6 +588,10 @@ ReturnCode Container::bindMountFileInContainer(const std::string &pathOnHost,
     }
 
     result = gatewaysDirInContainer() + "/" + pathInContainer;
+    if (readonly) {
+        return remountReadOnlyInContainer(result);
+    }
+
     return ReturnCode::SUCCESS;
 }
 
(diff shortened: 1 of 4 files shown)


thenor (approved):

libsoftwarecontainer/src/container.cpp
```diff
@@ -601,7 +605,31 @@ ReturnCode Container::bindMountFolderInContainer(const std::string &pathOnHost,
...
+        return remountReadOnlyInContainer(result);
+    } else {
+        return ReturnCode::SUCCESS;
+    }
+}
+
+ReturnCode Container::remountReadOnlyInContainer(const std::string &path)
+{
+    pid_t pid = INVALID_PID;
+    EnvironmentVariables envVars;
```
breakreturn: The env var 'variables' parameter in executeInContainer has a default value so if we don't plan to pass anything here I guess we can remove the env vars from this part of the code.

libsoftwarecontainer/src/container.cpp
```diff
@@ -574,7 +574,11 @@ ReturnCode Container::bindMountFileInContainer(const std::string &pathOnHost,
     }
 
     result = gatewaysDirInContainer() + "/" + pathInContainer;
-    return ReturnCode::SUCCESS;
+    if (readonly) {
+        return remountReadOnlyInContainer(result);
+    } else {
+        return ReturnCode::SUCCESS;
+    }
```
DunderRoffe, line 593: Could skip the else clause

libsoftwarecontainer/src/container.cpp
```diff
@@ -601,7 +605,31 @@ ReturnCode Container::bindMountFolderInContainer(const std::string &pathOnHost,
     }
 
     result = gatewaysDirInContainer() + "/" + pathInContainer;
-    return ReturnCode::SUCCESS;
+
+    if (readonly) {
+        return remountReadOnlyInContainer(result);
+    } else {
+        return ReturnCode::SUCCESS;
+    }
```
DunderRoffe, line 625: Could skip the else clause

libsoftwarecontainer/src/container.cpp
tobsan: Good point, I'll remove it.

libsoftwarecontainer/src/container.cpp
tobsan: Ok

libsoftwarecontainer/src/container.cpp
tobsan: Ok

PR #125, opened 2016-11-02
Improved integrator docs regarding cmake options

DunderRoffe:


tobsan (approved):

doc/chapters/configuration/index.rst
```diff
@@ -49,4 +49,33 @@ As it is indicated in :ref:`design chapter <design>`, each IPC is managed by a g
...
+    cmake -DENABLE_NETWORKGATEWAY=OFF ..
+    make
+    sudo make install
+
+A complete list of available gateways can be found at :ref:`Gateways <gateways>`.
+
+Documentation options
+---------------------
+These options allows specification of what documentation to generate when running make.
+
```
frznlogic: Would be nice with a list of the options even if they aren't properly documented on effects for now.

doc/chapters/configuration/index.rst
```diff
@@ -49,4 +49,33 @@ As it is indicated in :ref:`design chapter <design>`, each IPC is managed by a g
...
+Network
+-------
+Options governing how SoftwareContainer should react when establishing network access.
+
```
frznlogic: Would be nice with a list of the options even if they aren't properly documented on effects for now.

doc/chapters/configuration/index.rst
```diff
@@ -49,4 +49,33 @@ As it is indicated in :ref:`design chapter <design>`, each IPC is managed by a g
...
+Testing
+-------
+Options used when running tests, such as timing and coverage requirements.
```
frznlogic: Would be nice with a list of the options even if they aren't properly documented on effects for now.

doc/chapters/configuration/index.rst
```diff
@@ -49,4 +49,33 @@ As it is indicated in :ref:`design chapter <design>`, each IPC is managed by a g
...
+----------------
+If a gateway is disabled this means that the gateway will not be apart of SoftwareContainer at compile time. Hence this will disable the IPC mechanism that the gateway in question handles.
+As an example, the following snippet will install SoftwareContainer without any network support::
+
+    cmake -DENABLE_NETWORKGATEWAY=OFF ..
+    make
+    sudo make install
+
+A complete list of available gateways can be found at :ref:`Gateways <gateways>`.
+
```
frznlogic: Would be nice with a list of the options even if they aren't properly documented on effects for now.

frznlogic (changes requested):

doc/chapters/configuration/index.rst
```diff
@@ -49,4 +49,33 @@ As it is indicated in :ref:`design chapter <design>`, each IPC is managed by a g
 
 CMake Options
 =============
-CMake options can be used to configure softwareContainer. Desired gateways can be enabled/disabled by passing flags to cmake. (Currently following gateways are able for enable/disable option; network, pulse, device node, dbus, cgroup :ref:`Gateways <gateways>`).
+
+CMake options can be used to configure softwareContainer. The CMake options can be categorized into four groups:
```
breakreturn: Captital S in SoftwareContainer

doc/chapters/configuration/index.rst
breakreturn: We discussed this a bit yesterday. The risk of getting slightly skewed docs increase the more duplication we have. Maybe it can be OK with referencing the root README.md and some other part of the user-docs that shows an examples to allow the reader to put the pieces together. E.g. reference the README early, describe what sort of things are configured with CMake and then also reference a more complete example. Maybe it's possible to do in a way that the reader can put the two pieces together and figure out how to specify the different options. 

Or we go the other way and duplicate in such a way that it's easily grep'able.

My basic idea of this is essentially that the user should reference the root README.md for configuring and building SC itself, and the user-docs will explain other configuration options, e.g. the configuration files, and how it all ties together. One is totally configure-and-build centric, the other use-case centric.

breakreturn (approved):

doc/chapters/configuration/index.rst
frznlogic: Have we considered scripting a generated documentation in that case, cmake -LAH? Feels almost like a Gothenburgian wrote that ena la. Anyway, the documentation here just feels way too short and useless without the options. I understand not wanting duplicate stuff, but I think the docs are not of any use right now. My 2 cents.

breakreturn (approved):

frznlogic (approved):

PR #126, opened 2016-11-03
tests for common no longer fail with memory issues

tobsan:
The unit tests for the common library fails with invalid pointer
or double free (depending on run). This fixes that issue.


tobsan:
Don't merge until Jenkins has run this properly. The failures have been different in Jenkins and on my local dev VM.

PR #129, opened 2016-11-03
Fix iptables domain name lookup problem

kursatkobya:
* Applies policies after applying rules
* Applies rules after setting up network properly


DunderRoffe (approved):
Oh Yoda typing cool :smiling_imp:

tobsan (approved):

PR #134, opened 2016-11-04
Rewording of Filesystems/Standard Filesystem

DunderRoffe:


doc/chapters/filesystem/index.rst
```diff
@@ -5,64 +5,65 @@ Filesystems
...
 Standard filesystem
 ===================
 
 Each container has a rootfs that is bind mounted as the containers root
 filesystem of said container. The creation and bind mounting of the root
-filesystem is automated by LXC based on a configuration file and a shell script
+filesystem is automated based on a configuration file and a shell script
 provided at startup. The process of creating the filesystem and bind mounting
 different parts is spread out over several involved parties because of this.
+Following are a rough description of the steps taken when setting up the
```
breakreturn: Should be "is a" ?

breakreturn (approved):
I think it looks good overall given the scope of the changes. I think we will eventually need to revise and separate abstract descriptions from details due to maintenance burden, and also to see how we should relate to some more use-case centric configuration chapter. But I'd rather see these incremental additions to the docs than try to resolve everything too early.

tobsan (approved):
Looks good. Perhaps we should use some CMake vars to point out where things will be installed (but then we need to show how SC was configured when building the docs, so perhaps it's a bad idea).

PR #137, opened 2016-11-07
Add unit tests for IPTableEntry

kursatkobya:
unit tests for only testing IPTableEntry functions

### DIFF
--- a/libsoftwarecontainer/src/gateway/iptableentry.h
+++ b/libsoftwarecontainer/src/gateway/iptableentry.h
@@ -33,7 +33,11 @@ public:
      * @brief container for port filtering options. Used internally in a Rule.
      */
     struct portFilter {
-        portFilter() : any{0}, multiport{0}, ports{""} {};
+        portFilter(bool _any=0, bool _multiport=0, std::string _ports="") :
+            any(_any),
+            multiport(_multiport),
+            ports(_ports)
+            {};
         bool any;
         bool multiport;
         std::string ports;
@@ -67,6 +71,23 @@ public:
      */
     ReturnCode applyRules();
 
+    /**
+     * @brief Interprets a rule to iptables applicable string
+     * @return string indicating interpreted rule
+     */
+    std::string interpretRule(Rule rule);
+
+    /**
+     * @brief This function Interprets defaultTarget rule to iptables applicable policy string
+     *
+     * defaultTarget indicates what happens to packets if they don't match to any rules.
+     * iptables apply this functionality with setting policy. The role of this function is
+     * converting defaultTarget configuration value to iptables policy
+     * @return string indicating interpreted policy
+     * @return empty string when interfered incompatible m_defaultTarget
+     */
+    std::string interpretPolicy(void);
+
     unsigned int m_priority;
     std::string m_type;
     std::vector<Rule> m_rules;
@@ -78,19 +99,11 @@ private:
      */
     std::string convertTarget (Target& t);
 
-
     /**
      * @brief Inserts a rule to iptables
      * @return true  Upon success
      * @return false Upon failure
      */
-    ReturnCode insertRule(Rule rule);
-
-    /**
-    * @brief set the default policy on the m_type chain to m_defaultTarget
-    * @return true  Upon success
-    * @return false Upon failure
-    */
-    ReturnCode setPolicy();
+    ReturnCode insertCommand(std::string command);
 };
 
--- a/libsoftwarecontainer/unit-test/iptableentry_unittest.cpp
+++ b/libsoftwarecontainer/unit-test/iptableentry_unittest.cpp
@@ -1,0 +1,144 @@
+
+/*
+ * Copyright (C) 2016 Pelagicore AB
+ *
+ * Permission to use, copy, modify, and/or distribute this software for
+ * any purpose with or without fee is hereby granted, provided that the
+ * above copyright notice and this permission notice appear in all copies.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL
+ * WARRANTIES WITH REGARD TO THIS SOFTWARE INCLUDING ALL IMPLIED
+ * WARRANTIES OF MERCHANTABILITY AND FITNESS. IN NO EVENT SHALL THE AUTHOR
+ * BE LIABLE FOR ANY SPECIAL, DIRECT, INDIRECT, OR CONSEQUENTIAL DAMAGES
+ * OR ANY DAMAGES WHATSOEVER RESULTING FROM LOSS OF USE, DATA OR PROFITS,
+ * WHETHER IN AN ACTION OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION,
+ * ARISING OUT OF OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS
+ * SOFTWARE.
+ *
+ * For further information see LICENSE
+ */
+
+
+#include "gateway/iptableentry.h"
+#include "gmock/gmock.h"
+#include "gtest/gtest.h"
+
+class MockIPTableEntry :
+    public IPTableEntry
+{
+public:
+    MockIPTableEntry():IPTableEntry()
+    {
+        m_type = "INPUT";
+        m_defaultTarget = Target::ACCEPT;
+    }
+};
+
+class IPTableEntryTest : public ::testing::Test
+{
+protected:
+    ::testing::NiceMock<MockIPTableEntry> ipTable;
+};
+
+/*
+ * @brief Tests if ACCEPT policy matched with default targets
+ * */
+TEST_F(IPTableEntryTest, TestIPTableEntryAcceptPolicy) {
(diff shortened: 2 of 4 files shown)


libsoftwarecontainer/unit-test/iptableentry_unittest.cpp
```diff
@@ -0,0 +1,96 @@
...
+{
+protected:
+    ::testing::NiceMock<MockIPTableEntry> ipTable;
+};
+
+//test policies
+TEST_F(IPTableEntryTest, TestIPTableEntryPolicies) {
+    IPTableEntry::Rule r;
+    r.host = "127.0.0.1/16";
+    r.ports.any = true; r.ports.multiport = true; r.ports.ports = "80,8080";
```
tobsan: Don't put multiple statements on one line. You can set all of these when declaring the rule, right?

libsoftwarecontainer/unit-test/iptableentry_unittest.cpp
```diff
@@ -0,0 +1,96 @@
...
+    //Test wrong default Policy
+    ipTable.m_defaultTarget = IPTableEntry::Target::REJECT;
+    ASSERT_EQ("", ipTable.interpretPolicy());
+}
+
+
+TEST_F(IPTableEntryTest, TestIPTableEntryRules) {
+    IPTableEntry::Rule r;
+    r.host = "127.0.0.1/16";
+    r.ports.any = true; r.ports.multiport = true; r.ports.ports = "80,8080";
```
tobsan: Don't put multiple statements on one line.

libsoftwarecontainer/unit-test/iptableentry_unittest.cpp
```diff
@@ -0,0 +1,96 @@
...
+    r.target = IPTableEntry::Target::ACCEPT;
+    //Test INPUT chain
+    //multiport list test
+    ASSERT_EQ("iptables -A INPUT -s 127.0.0.1/16 -p tcp --match multiport --sports 80,8080"
```
tobsan, line 76: You could put 127.0.0.1/16 as a variable and check it. Not sure it is worth the effort though, just a thought.

tobsan (changes requested):
Looks good in general, mostly cosmetic comments from me.

libsoftwarecontainer/unit-test/iptableentry_unittest.cpp
kursatkobya: You are right. Let me change constructor for ports structure

libsoftwarecontainer/src/gateway/iptableentry.h
```diff
@@ -67,6 +67,18 @@ class IPTableEntry
...
 
+    /**
+     * @brief Interprets a rule to iptables applicable string
+     * @return string indicating interpreted rule
+     */
+    std::string interpretRule(Rule rule);
+
+    /**
+     * @brief Interprets defaultTarget rule to iptables applicable string
+     * @return string indicating interpreted policy
```
DunderRoffe, line 86: Maybe we should mention that empty string is returned on failure

libsoftwarecontainer/unit-test/iptableentry_unittest.cpp
```diff
@@ -0,0 +1,100 @@
...
+    r.target = IPTableEntry::Target::ACCEPT;
+    ASSERT_EQ("iptables -P INPUT ACCEPT", ipTable.interpretPolicy());
+
+    ipTable.m_defaultTarget = IPTableEntry::Target::DROP;
+    ASSERT_EQ("iptables -P INPUT DROP", ipTable.interpretPolicy());
+
+    //Test wrong default Policy
+    ipTable.m_defaultTarget = IPTableEntry::Target::REJECT;
+    ASSERT_EQ("", ipTable.interpretPolicy());
+}
```
breakreturn, line 65: Does these tests need to be run in a sequence? Otherwise split them into multiple tests.

libsoftwarecontainer/unit-test/iptableentry_unittest.cpp
```diff
@@ -0,0 +1,100 @@
...
+    }
+};
+
+class IPTableEntryTest : public ::testing::Test
+{
+protected:
+    ::testing::NiceMock<MockIPTableEntry> ipTable;
+};
+
+//test policies
```
breakreturn: I think it would be good with a more explanatory comment.

libsoftwarecontainer/unit-test/iptableentry_unittest.cpp
```diff
@@ -0,0 +1,100 @@
...
+    //multiport range test
+    r.ports.ports = "80:85";
+    ASSERT_EQ("iptables -A OUTPUT -d 127.0.0.1/16 -p tcp --match multiport --dports 80:85"
+            " -j ACCEPT", ipTable.interpretRule(r));
+
+    //sinlgeport test
+    r.ports.multiport = false; r.ports.ports = "80";
+    ASSERT_EQ("iptables -A OUTPUT -d 127.0.0.1/16 -p tcp --dport 80"
+            " -j ACCEPT", ipTable.interpretRule(r));
+}
```
breakreturn, line 144: Does these tests need to be run in a sequence? Otherwise split them into multiple tests.

libsoftwarecontainer/unit-test/iptableentry_unittest.cpp
```diff
@@ -0,0 +1,100 @@
...
+    ipTable.m_defaultTarget = IPTableEntry::Target::DROP;
+    ASSERT_EQ("iptables -P INPUT DROP", ipTable.interpretPolicy());
+
+    //Test wrong default Policy
+    ipTable.m_defaultTarget = IPTableEntry::Target::REJECT;
+    ASSERT_EQ("", ipTable.interpretPolicy());
+}
+
+
+TEST_F(IPTableEntryTest, TestIPTableEntryRules) {
```
breakreturn: A descriptive comment would be nice here.

libsoftwarecontainer/unit-test/iptableentry_unittest.cpp
kursatkobya: The aim of test is to test whether IPTableEntry can procude valid rule with given rule. So i believe it is insignificant to parse rules to match string. But it is always doable on demand.

libsoftwarecontainer/src/gateway/iptableentry.h
```diff
@@ -67,6 +67,18 @@ class IPTableEntry
...
     ReturnCode applyRules();
 
+    /**
+     * @brief Interprets a rule to iptables applicable string
+     * @return string indicating interpreted rule
+     */
+    std::string interpretRule(Rule rule);
+
+    /**
+     * @brief Interprets defaultTarget rule to iptables applicable string
```
breakreturn: Does this comment make sense in the iptables domain? E.g. what's 'defaultTarget'?

breakreturn (changes requested):

libsoftwarecontainer/src/gateway/iptableentry.h
kursatkobya: Actually this check is not supposed to be interfered on real scenarios. 
Normally NetworkGateway parser eliminates the possibility. 
But in case for maybe future use of it in another gateway or class it should be beneficial.

libsoftwarecontainer/unit-test/iptableentry_unittest.cpp
kursatkobya: They do not have to be run in sequence. Yet i believe they have a meaningful unity.

libsoftwarecontainer/src/gateway/iptableentry.h
breakreturn: If it's a public member of the class then I think it's good to make the assumption that the user of this class might do erroneous things.

libsoftwarecontainer/src/gateway/iptableentry.h
DunderRoffe: defaultTarget is something that we named in our syntax. It has a maps to policies in iptables i belive

kursatkobya: defaultTarget indicates what happens to packets if they don't match to any rules. 
iptables apply this functionality with setting policy. So defaultTarget here represents default behaviour of packets in NetworkGateway configuration.

breakreturn: What is "our syntax"? The GW configs? In that case, I'm not sure I like a config implementation detail to sneak its way to here. Perhaps it would be better to express it in more general terms so we don't have to remember updating this comment in case the config syntax changes slightly?

breakreturn: Kursat - that explanation makes for a better comment then I think, put the explanation there.

breakreturn (approved):

libsoftwarecontainer/unit-test/iptableentry_unittest.cpp
kursatkobya: Done

tobsan (approved):

DunderRoffe (approved):

PR #139, opened 2016-11-07
service-tests: bind mounted dirs are not read-only

tobsan:
This was not a problem before since read-only didn't work. Now it works,
so we need r/w access in order to write inside the container.


DunderRoffe (approved):

breakreturn:
I see it's merged now, but shouldn't the readonly flag be in the "data" dict passed?

PR #140, opened 2016-11-08
Minor fix in service-test framework

breakreturn:
The "data" dict passed by tests to the Container
helper now carries the readonly flag as well.

### DIFF
--- a/servicetest/testframework/lib.py
+++ b/servicetest/testframework/lib.py
@@ -50,6 +50,7 @@ class Container():
     CONFIG = "config"
     BIND_MOUNT_DIR = "bind-mount-dir"
     HOST_PATH = "host-path"
+    READONLY = False
 
     def __init__(self):
         bus = dbus.SystemBus()
@@ -109,7 +110,8 @@ class Container():
             {
                 Container.CONFIG: "{enableWriteBuffer: false}",
                 Container.HOST_PATH: my_path_string_variable,
-                Container.BIND_MOUNT_DIR: "app"
+                Container.BIND_MOUNT_DIR: "app",
+                Container.READONLY: True
             }
 
             The values in the dict are passed as arguments to the D-Bus methods on SoftwareContainerAgent.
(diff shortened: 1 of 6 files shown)


servicetest/testframework/lib.py
```diff
@@ -109,19 +110,25 @@ def start(self, data):
...
             The values in the dict are passed as arguments to the D-Bus methods on SoftwareContainerAgent.
             The dict is mapped like so:
             Container.CONFIG - argument to SoftwareContainerAgent::CreateContainer
             Container.HOST_PATH - second argument to SoftwareContainerAgent::BindMountFolderInContainer
             Container.BIND_MOUNT_DIR - third argument to SoftwareContainerAgent::BindMountFolderInContainer
+            Container.READONLY - fourth argument to SoftwareContainerAgent::BindMountFolderInContainer
         """
+        # Set some defaults if the user did not pass all values in the data dict
+        if Container.READONLY not in data:
+            data[Container.READONLY] = False
```
DunderRoffe: This is a weird way of setting default values. Either set it in the function or make it mandatory. This will easily be missed by users of the lib.

DunderRoffe (changes requested):

servicetest/testframework/lib.py
breakreturn: There is no function to set the default in. We'd need to check if the key is in the dict anyway and set some value so it doesn't really work like it was written originally unless the value is mandatory. And since almost all the tests will want to have this as 'False', I'm not sure we should have it mandatory for convenience reasons. 

What do you consider the issue being easily missed? The dict and it's usage is documented, it's just that we don't break if this value is not specified.

An option is of course to have the framework provide a method to set these values and skip the passing of a dict from the tests. The the defaults could be added on the method signature.

breakreturn: Perhaps the best option here is to make the value in the dict mandatory (in the sense that the framework won't work unless the key is in the dict). It creates one more line of boiler plate in the test modules, but provides explicitness.

DunderRoffe (approved):

PR #141, opened 2016-11-08
Create NetworkGatewayParser class with unit tests

kursatkobya:
This patch create NetworkGatewayParser class and uses it from NetworkGateway
Also contains tests of parsing configuration into rule structure

### DIFF
--- a/libsoftwarecontainer/src/gateway/networkgateway.cpp
+++ b/libsoftwarecontainer/src/gateway/networkgateway.cpp
@@ -23,6 +23,7 @@
 #include "unistd.h"
 #include "networkgateway.h"
 #include "generators.h"
+#include "networkgatewayparser.h"
 
 NetworkGateway::NetworkGateway() :
     Gateway(ID),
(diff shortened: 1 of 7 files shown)


libsoftwarecontainer/src/gateway/networkgateway.cpp
```diff
@@ -38,71 +39,8 @@ NetworkGateway::~NetworkGateway()
...
-    e.m_defaultTarget = parseTarget(readTarget);
-    if (e.m_defaultTarget == IPTableEntry::Target::INVALID_TARGET
-        || e.m_defaultTarget == IPTableEntry::Target::REJECT) {
-        log_error() << "Default target '" << readTarget << "' is not a supported target";
-        return ReturnCode::FAILURE;
-    }
+    NetworkGatewayParser configParser;
+    ReturnCode returnValue = configParser.parseNetworkGatewayConfiguration(element, e);
 
     m_entries.push_back(e);
```
DunderRoffe: Does this mean that we push entries even if the fail to parse?

DunderRoffe (changes requested):

libsoftwarecontainer/src/gateway/networkgateway.cpp
kursatkobya: Ups, fixed

libsoftwarecontainer/src/gateway/networkgateway.cpp
```diff
@@ -38,172 +39,24 @@ NetworkGateway::~NetworkGateway()
...
-        log_error() << "No rules specified";
-        return ReturnCode::FAILURE;
-    }
+    NetworkGatewayParser configParser;
+    ReturnCode returnValue = configParser.parseNetworkGatewayConfiguration(element, e);
 
-    if (!json_is_array(rules)) {
-        log_error() << "Rules not specified as an array";
-        return ReturnCode::FAILURE;
+    if (ReturnCode::SUCCESS == returnValue) {
```
tobsan: Please use isSuccess instead here.

libsoftwarecontainer/src/gateway/networkgateway.cpp
```diff
@@ -38,172 +39,24 @@ NetworkGateway::~NetworkGateway()
...
+        m_entries.push_back(e);
     }
```
tobsan, line 47: In case of error, shouldn't we print something high-level here like, "could not parse, sorry about that"

libsoftwarecontainer/unit-test/networkgatewayparser_unittest.cpp
```diff
@@ -0,0 +1,171 @@
...
+    public NetworkGatewayParser
+{
+public:
+    MockNetworkGatewayParser() { };
+};
+
+class NetworkGatewayParserTest : public ::testing::Test
+{
+protected:
+    ::testing::NiceMock<MockNetworkGatewayParser> networkParser;
```
tobsan: This mock class doesn't mock anything, so you could use the NetworkGatewayParser directly and remove the Mock class, I think.

tobsan (changes requested):
This looks good, but I have some changes I'd like to see.

libsoftwarecontainer/src/gateway/networkgateway.cpp
kursatkobya: Actually in case of error, parser says some blues sentences and the function returns error to parseConfig which also outputs error. So imho it would be redundant info.

libsoftwarecontainer/src/gateway/networkgateway.cpp
kursatkobya: done

libsoftwarecontainer/unit-test/networkgatewayparser_unittest.cpp
kursatkobya: done

tobsan (approved):
Looks nice!

DunderRoffe (approved):

PR #145, opened 2016-11-08
Using GNUInstallDirs to set install paths

DunderRoffe:


CMakeLists.txt
```diff
@@ -60,7 +62,7 @@ add_definitions(${DBusCpp_CFLAGS_OTHER})
 add_definitions(${Glibmm_CFLAGS_OTHER})
 add_definitions(${LXC_CFLAGS_OTHER})
 add_definitions(${Jansson_CFLAGS_OTHER})
-add_definitions(-DINSTALL_PREFIX="${CMAKE_INSTALL_PREFIX}")
+add_definitions(-DINSTALL_PREFIX="${CMAKE_INSTALL_FULL_BINDIR}")
```
tobsan: If this is already the bindir, why do we add /bin to the setup_softwarecontainer.sh script? Also, it feels a bit misleading to say install prefix, when it is actually the bindir we're adding here.

tobsan (changes requested):
Just one concrete comment. I see this as a good start for handling the issue on running the agent on the session bus (since that's what you get if installing to a non-/usr prefix.

CMakeLists.txt
tobsan: Ok my bad (read it as being /bin/bin), but the name is still bad :)

tobsan (approved):
This looks nice!

tobsan (approved):

tobsan:
Just noticing that the troubleshooting chapter in the user docs still mentions that one has to set prefix to /usr, perhaps that should be removed as part of this PR?

tobsan (approved):

tobsan:
Just noticing that the troubleshooting chapter in the user docs still mentions that one has to set prefix to /usr, perhaps that should be removed as part of this PR?

breakreturn (approved):

PR #152, opened 2016-11-10
Add service tests for NetworkGateway

kursatkobya:


servicetest/networkgateway/test_network_gateway.py
```diff
@@ -0,0 +1,233 @@
...
+            10 for expected output discconnectivity
+            15 for expected input connectivity
+
+        these durations will be shorten with -i option when possible    
+    """
+    def test_network_policy_drop(self):
+        """
+            Test setting network default target to drop and tests the corresponding behavior
+            which is being not possible to ping a hostname
+        """
```
breakreturn, line 100: In all of these, pull up the first text line so it's part of the opening quotes, like you've done in the class comment above.

breakreturn (changes requested):
A minor comment on style, but in general I think this looks great.

DunderRoffe (approved):
Looks really good. It would be cool if we could use some python lib instead of the ping binary though. That would almost entirely remove our need for busybox

breakreturn (approved):

PR #158, opened 2016-11-11
Use the JSONParser statically, don't inherit.

tobsan:
All functions in the JSONParser are static, now all uses of it
is also static, removing the inheritance of the class from various
places.


DunderRoffe (changes requested):
Awesome sauce!
Would be nice if the JSONParser class were made final as well.

thenor (approved):

DunderRoffe (approved):

PR #161, opened 2016-11-11
Introduced an option to disable the WaylandGateway

DunderRoffe:
Since there is an "ENABLE" option for all other Gateways,  I added an option for the Wayland gateway as well.

@tobsan is there any reason why this was not done before or is it just a miss?

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,6 +53,9 @@ To disable support for various gateways at compile time, set
 * -DENABLE_DEVICENODEGATEWAY=OFF (for device node gateway)
 * -DENABLE_DBUSGATEWAY=OFF (for dbus)
 * -DENABLE_CGROUPSGATEWAY=OFF (for cgroups)
+* -DENABLE_WAYLANDGATEWAY=OFF (for wayland)
+* -DENABLE_FILEGATEWAY=OFF (for mounting files)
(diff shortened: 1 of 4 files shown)


tobsan (changes requested):
There are no options for FileGateway and EnvGateway. If we want to have options for all of them, those two should be included in that as well (and should default to ON also).

README.md
```diff
@@ -53,6 +53,9 @@ To disable support for various gateways at compile time, set
 * -DENABLE_DEVICENODEGATEWAY=OFF (for device node gateway)
 * -DENABLE_DBUSGATEWAY=OFF (for dbus)
 * -DENABLE_CGROUPSGATEWAY=OFF (for cgroups)
+* -DENABLE_WAYLANDGATEWAY=OFF (for wayland)
+* -DENABLE_FILEGATEWAY=OFF (for mounting files)
+* -DENABLE_ENVGATEWAY=OFF (for manipulateing environment variables)
```
tobsan: manipulating*

tobsan (approved):
Just a typo, looks good apart from that.

README.md
DunderRoffe: fixed

PR #163, opened 2016-11-14
Network service-tests use framework helper

breakreturn:


servicetest/networkgateway/test_network_gateway.py
```diff
@@ -226,8 +250,5 @@ def test_ip_diversity(self):
             assert 0 == testhelper.compare_ips(CURRENT_DIR, "f1", "f2")
```
kursatkobya, line 252: This can create problems later.

tobsan (changes requested):
Can we please remove google.com from these tests. If we want to use something else than example.com, we can use example.org or example.net, see https://tools.ietf.org/html/rfc2606#section-3

servicetest/testframework/testhelper.py
```diff
@@ -150,14 +198,48 @@ def __file_name(self):
...
-        h = EnvironmentHelper(file_base_path)
+        env_vars_file_base_path = env_vars_file_base_path_parsed_value.pop()
+        h = EnvironmentHelper(env_vars_file_base_path)
         # Get the environment from inside the container
         env_vars = h.get_env_vars()
         # Dump the information for the helper to read back later in the tests
         h.write_result(env_vars)
+
+    do_ping_file_base_path_parsed_value = getattr(args, DO_PING_OPTION)
+    if do_ping_file_base_path_parsed_value is not None:
```
DunderRoffe: really hard to read as the variables due to the combination: long and similar. Consider dropping the prefix and add a comment instead.

DunderRoffe (changes requested):

servicetest/testframework/testhelper.py
breakreturn: I agree. I think this structure will need to change anyway, when and if we can make use of this generic helper design. But point taken, I will try to find some better names or use whitespace for an easier time for the eye.

servicetest/networkgateway/test_network_gateway.py
breakreturn: Added a comment that this is broken at the moment.

DunderRoffe (approved):

tobsan (approved):

PR #168, opened 2016-11-17
Unique IP address

kursatkobya:
Generation of unique IP address for each SoftwareContainer
Updates related service test environment


libsoftwarecontainer/src/gateway/networkgateway.h
```diff
@@ -98,39 +92,40 @@ class NetworkGateway :
...
      */
-    virtual bool isBridgeAvailable();
+    virtual ReturnCode isBridgeAvailable();
 
-    std::string m_ip;
+    struct in_addr m_ip;
+    uint32_t m_netmask;
     std::string m_gateway;
 
+
```
DunderRoffe: why the extra whitespace

libsoftwarecontainer/src/softwarecontainer.cpp
```diff
@@ -55,9 +54,10 @@
 
 namespace softwarecontainer {
 
-SoftwareContainer::SoftwareContainer(std::shared_ptr<Workspace> workspace, const std::string &containerID) :
+SoftwareContainer::SoftwareContainer(std::shared_ptr<Workspace> workspace, const ContainerID id) :
```
DunderRoffe, line 57: Would be nice to have the string prefix as well here or maybe the ContainerID type should be changed to a std::string.

DunderRoffe (changes requested):

libsoftwarecontainer/src/softwarecontainer.cpp
kursatkobya: I couldn't reason it, i think agent shouldn't deal with LXC related id. 
The idea of changing id to integer was spin-off from stand-up to get rid of comparision in gateways.

DunderRoffe (approved):

breakreturn (approved):

PR #169, opened 2016-11-18
Add gateway ip and netmask

kursatkobya:
Adding gateway ip and network mask bits input to NetworkGateway constructor

### DIFF
--- a/libsoftwarecontainer/src/gateway/networkgateway.cpp
+++ b/libsoftwarecontainer/src/gateway/networkgateway.cpp
@@ -24,16 +24,30 @@
 #include "networkgateway.h"
 #include "networkgatewayparser.h"
 
-NetworkGateway::NetworkGateway(const int32_t id) :
+NetworkGateway::NetworkGateway(const int32_t id,
+                               const std::string gateway,
+                               const uint8_t maskBits) :
     Gateway(ID),
-    m_netmask(0xFF),
-    m_gateway("10.0.3.1"),
+    m_gateway(gateway),
     m_interfaceInitialized(false),
-    m_containerID(id) { }
-
-NetworkGateway::~NetworkGateway()
+    m_containerID(id)
 {
+    /*
+     * Netmask are used for determining range of ip adress assignment. maskBits represent bit-count
+     * for creating ip range starting from least significant bit of m_gateway. Since an ipv4 address
+     * consist of 32 bits, maskBits shall not be greater than 32. And since bit 0 and bit 31 cannot
+     * give a range but a single exact value, those will not be accepted as a maskBits.
+     */
+    if (maskBits > 31 || maskBits < 1) {
+        log_error() << "inappropriate netmask : " << maskBits;
+        throw ReturnCode::FAILURE;
+    }
+
+    // value of m_netmask interprets maskBits to a mask integer to calculate range.
+    m_netmask = (1L<<(32-maskBits));
 }
+
+NetworkGateway::~NetworkGateway() { }
 
 ReturnCode NetworkGateway::readConfigElement(const json_t *element)
 {
--- a/libsoftwarecontainer/src/softwarecontainer.cpp
+++ b/libsoftwarecontainer/src/softwarecontainer.cpp
@@ -121,7 +121,13 @@ ReturnCode SoftwareContainer::init()
     }
 
 #ifdef ENABLE_NETWORKGATEWAY
-    addGateway(new NetworkGateway(m_containerID));
+    try {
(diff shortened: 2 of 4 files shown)


libsoftwarecontainer/src/gateway/networkgateway.cpp
```diff
@@ -24,17 +24,25 @@
...
-NetworkGateway::~NetworkGateway()
+    m_containerID(id)
 {
+    if (maskBits > 31 || maskBits < 1) {
+        log_error() << "inappropriate netmask : " << maskBits;
+        m_netmask = 0xFF;
+        return;
+    }
+
+    m_netmask = (1L<<(32-maskBits));
```
breakreturn, line 47: I would appreciate more comments around these parts.

libsoftwarecontainer/unit-test/networkgateway_unittest.cpp
```diff
@@ -27,7 +27,7 @@ class MockNetworkGateway :
     public NetworkGateway
 {
 public:
-    MockNetworkGateway():NetworkGateway(11) {}
+    MockNetworkGateway():NetworkGateway(11,"10.0.3.1", 0) {}
```
breakreturn: Spaces here and above. :D

libsoftwarecontainer/src/gateway/networkgateway.cpp
DunderRoffe: I think you should write the magic numbers in hex or binary when dealing with bits

DunderRoffe (approved):

libsoftwarecontainer/src/gateway/networkgateway.cpp
```diff
@@ -24,17 +24,33 @@
...
+     * for creating ip range starting from least significant bit of m_gateway. Since an ipv4 address
+     * consist of 32 bits, maskBits shall not be greater the 32. And since bit 0 and bit 31 cannot
+     * give a range but a single exact value, those cannot be accepted as a maskBits.
+     *
+     * value of m_netmask interprets maskBits to a mask integer to calculate range.
+     */
+    if (maskBits > 31 || maskBits < 1) {
+        log_error() << "inappropriate netmask : " << maskBits;
+        m_netmask = 0xFF;
+        return;
```
tobsan: I would like to see an exception being thrown here actually.

tobsan (changes requested):

libsoftwarecontainer/src/gateway/networkgateway.cpp
kursatkobya: Good idea, consider it done.

libsoftwarecontainer/src/gateway/networkgateway.cpp
```diff
@@ -24,17 +24,32 @@
...
+     * give a range but a single exact value, those cannot be accepted as a maskBits.
+     *
+     * value of m_netmask interprets maskBits to a mask integer to calculate range.
+     */
+    if (maskBits > 31 || maskBits < 1) {
+        log_error() << "inappropriate netmask : " << maskBits;
+        throw ReturnCode::FAILURE;
+    }
+
+    m_netmask = (1L<<(32-maskBits));
```
tobsan, line 47: A comment about this would be helpful.

libsoftwarecontainer/src/softwarecontainer.cpp
```diff
@@ -121,7 +121,12 @@ ReturnCode SoftwareContainer::init()
     }
 
 #ifdef ENABLE_NETWORKGATEWAY
-    addGateway(new NetworkGateway(m_containerID));
+    try {
+        addGateway(new NetworkGateway(m_containerID, "10.0.3.1", 16));
+    } catch (ReturnCode failure) {
+        log_error() << "inappropriate netmask";
```
tobsan: This error message could be made clearer.

breakreturn (approved):

libsoftwarecontainer/src/gateway/networkgateway.cpp
kursatkobya: There is a huge comment block explaining the snippet. I just didn't want over-commenting. But surely i can add a comment if you don't think the comment above is not sufficient.

breakreturn: I suggest you move the part from the block comment that is about the last line so it reads nicer totally, i.e split the comment up to be aligned with the organization of the code.

PR #176, opened 2016-11-23
Add status return to all D-Bus calls

kursatkobya:
agent code is updated
some unit tests are updated
all service tests are updated
api document is updated

### DIFF
--- a/agent/src/softwarecontaineragent.cpp
+++ b/agent/src/softwarecontaineragent.cpp
@@ -172,14 +172,15 @@ SoftwareContainerAgent::SoftwareContainerPtr SoftwareContainerAgent::makeSoftwar
     return container;
 }
 
-ContainerID SoftwareContainerAgent::createContainer(const std::string &config)
+bool SoftwareContainerAgent::createContainer(const std::string &config, ContainerID &containerID)
 {
     profilepoint("createContainerStart");
     profilefunction("createContainerFunction");
 
     if (!parseConfig(config)) {
         log_error() << "The provided config could not be successfully parsed";
-        return -1;
+        containerID = INVALID_CONTAINER_ID;
+        return false;
     }
 
     std::pair<ContainerID, SoftwareContainerPtr> pair = std::move(getContainerPair());
@@ -187,7 +188,8 @@ ContainerID SoftwareContainerAgent::createContainer(const std::string &config)
 
     if (isError(pair.second->init())) {
         log_error() << "Could not init the container";
-        return -1;
+        containerID = INVALID_CONTAINER_ID;
+        return false;
     }
 
     // TODO: Would be nice if this could be done async.
@@ -197,8 +199,8 @@ ContainerID SoftwareContainerAgent::createContainer(const std::string &config)
 
     m_containers.insert(std::move(pair));
 
-    // Return the ContainerID for the container
-    return pair.first;
+    containerID = pair.first;
+    return true;
 }
 
 std::pair<ContainerID, SoftwareContainerAgent::SoftwareContainerPtr> SoftwareContainerAgent::getContainerPair()
@@ -242,19 +244,21 @@ bool SoftwareContainerAgent::writeToStdIn(pid_t pid, const std::vector<uint8_t> 
     return true;
 }
 
-pid_t SoftwareContainerAgent::launchCommand(ContainerID containerID,
+bool SoftwareContainerAgent::launchCommand(ContainerID containerID,
                                             uid_t userID,
                                             const std::string &cmdLine,
                                             const std::string &workingDirectory,
                                             const std::string &outputFile,
                                             const EnvironmentVariables &env,
+                                            int32_t &pid,
                                             std::function<void (pid_t, int)> listener)
 {
     profilefunction("launchCommandFunction");
     SoftwareContainer *container = nullptr;
     if (!checkContainer(containerID, container)) {
         log_error() << "Invalid container ID " << containerID;
-        return INVALID_PID;
+        pid = INVALID_PID;
+        return false;
     }
 
     // Set up a CommandJob for this run in the container
@@ -270,7 +274,8 @@ pid_t SoftwareContainerAgent::launchCommand(ContainerID containerID,
     // Start it
     if (isError(job->start())) {
         log_error() << "Could not start job";
-        return INVALID_PID;
+        pid = INVALID_PID;
+        return false;
     }
 
     // If things went well, do what we need when it exits
@@ -283,7 +288,8 @@ pid_t SoftwareContainerAgent::launchCommand(ContainerID containerID,
 
     profilepoint("launchCommandEnd");
 
-    return job->pid();
+    pid = job->pid();
+    return true;
 }
 
 bool SoftwareContainerAgent::setContainerName(ContainerID containerID, const std::string &name)
@@ -352,25 +358,29 @@ bool SoftwareContainerAgent::resumeContainer(ContainerID containerID)
     return isSuccess(container->resume());
 }
 
-std::string SoftwareContainerAgent::bindMountFolderInContainer(const ContainerID containerID,
+bool SoftwareContainerAgent::bindMountFolderInContainer(const ContainerID containerID,
                                                                const std::string &pathInHost,
                                                                const std::string &pathInContainer,
-                                                               bool readOnly)
+                                                               bool readOnly,
+                                                               std::string &returnPath)
 {
     profilefunction("bindMountFolderInContainerFunction");
     SoftwareContainer *container = nullptr;
     if (!checkContainer(containerID, container)) {
         log_error() << "Invalid container ID " << containerID;
-        return "";
+        returnPath = "";
+        return false;
     }
 
     ReturnCode result = container->getContainer()->bindMountFolderInContainer(pathInHost, pathInContainer, readOnly);
     if (isError(result)) {
         log_error() << "Unable to bind mount folder " << pathInHost << " to " << pathInContainer;
-        return "";
+        returnPath = "";
+        return false;
     }
 
-    return pathInContainer;
+    returnPath = pathInContainer;
+    return true;
 }
 
 bool SoftwareContainerAgent::setGatewayConfigs(const ContainerID &containerID,
--- a/agent/src/softwarecontaineragent.h
+++ b/agent/src/softwarecontaineragent.h
@@ -116,13 +116,11 @@ public:
      * @brief Create a new container
      *
      * @param config container-wide configuration string
+     * @param a reference to ContainerID
      *
-     * @return If the container is successfully created a positive ContainerID
-     *         representing the newly created container will be returned.
-     *         Otherwise, -1 will be returned.
-     *
+     * @return If the container is successfully created true otherwise false
      */
-    ContainerID createContainer(const std::string &config);
+    bool createContainer(const std::string &config, ContainerID &containerID);
 
 
     /**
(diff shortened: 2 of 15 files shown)


agent/src/softwarecontaineragent.cpp
```diff
@@ -172,22 +172,24 @@ SoftwareContainerAgent::SoftwareContainerPtr SoftwareContainerAgent::makeSoftwar
...
+        return false;
     }
 
     std::pair<ContainerID, SoftwareContainerPtr> pair = std::move(getContainerPair());
     pair.second->setMainLoopContext(m_mainLoopContext);
 
     if (isError(pair.second->init())) {
         log_error() << "Could not init the container";
-        return -1;
+        containerID = -1;
```
DunderRoffe: We should add a DEFINE or something for INVALID_CONTAINER_ID instead of using magic numbers

agent/src/softwarecontaineragentadaptor.cpp
```diff
@@ -55,23 +58,26 @@ bool SoftwareContainerAgentAdaptor::ShutDownContainerWithTimeout(
...
+    const bool &readOnly,
+    std::string& pathInContainer,
+    bool& success)
 {
-    return m_agent.bindMountFolderInContainer(containerID,
-                                              pathInHost,
-                                              subPathInContainer,
-                                              readOnly);
+    success = m_agent.bindMountFolderInContainer(containerID,
+                                                         pathInHost,
```
DunderRoffe: Fix indentation

DunderRoffe (changes requested):
Looks good, only small things to fix

agent/src/softwarecontaineragent.cpp
tobsan: Actually there is already a defined INVALID_CONTAINER_ID in softwarecontainer-common.h

agent/src/softwarecontaineragent.cpp
```diff
@@ -172,22 +172,24 @@ SoftwareContainerAgent::SoftwareContainerPtr SoftwareContainerAgent::makeSoftwar
     return container;
 }
 
-ContainerID SoftwareContainerAgent::createContainer(const std::string &config)
+bool SoftwareContainerAgent::createContainer(const std::string &config, ContainerID& containerID)
```
tobsan: Style issue: should be ContainerID &containerID (move the &)

agent/src/softwarecontaineragent.cpp
```diff
@@ -242,19 +244,21 @@ bool SoftwareContainerAgent::writeToStdIn(pid_t pid, const std::vector<uint8_t>
...
 }
 
-pid_t SoftwareContainerAgent::launchCommand(ContainerID containerID,
+bool SoftwareContainerAgent::launchCommand(ContainerID containerID,
                                             uid_t userID,
                                             const std::string &cmdLine,
                                             const std::string &workingDirectory,
                                             const std::string &outputFile,
                                             const EnvironmentVariables &env,
+                                            int32_t& pid,
```
tobsan: Move the & here also

agent/src/softwarecontaineragent.cpp
tobsan: Move the &

agent/src/softwarecontaineragent.cpp
```diff
@@ -352,26 +358,30 @@ bool SoftwareContainerAgent::resumeContainer(ContainerID containerID)
     return isSuccess(container->resume());
 }
 
-std::string SoftwareContainerAgent::bindMountFolderInContainer(const ContainerID containerID,
+bool SoftwareContainerAgent::bindMountFolderInContainer(const ContainerID containerID,
                                                                const std::string &pathInHost,
                                                                const std::string &subPathInContainer,
-                                                               bool readOnly)
+                                                               bool readOnly,
+                                                               std::string& pathInContainer)
```
tobsan: Move the &

agent/src/softwarecontaineragent.h
```diff
@@ -116,13 +116,11 @@ class SoftwareContainerAgent
...
+     * @param a reference to ContainerID
      *
-     * @return If the container is successfully created a positive ContainerID
-     *         representing the newly created container will be returned.
-     *         Otherwise, -1 will be returned.
-     *
+     * @return If the container is successfully created true otherwise false
      */
-    ContainerID createContainer(const std::string &config);
+    bool createContainer(const std::string &config, ContainerID& containerID);
```
tobsan: Move the &

agent/src/softwarecontaineragent.h
```diff
@@ -142,14 +140,17 @@ class SoftwareContainerAgent
...
+     * @return true or false indicating whether or not the operation was successful.
      */
-    pid_t launchCommand(ContainerID containerID,
+    bool launchCommand(ContainerID containerID,
                         uid_t userID,
                         const std::string &cmdLine,
                         const std::string &workingDirectory,
                         const std::string &outputFile,
                         const EnvironmentVariables &env,
+                        int32_t& pid,
```
tobsan: Move the &

agent/src/softwarecontaineragent.h
```diff
@@ -196,13 +197,15 @@ class SoftwareContainerAgent
...
-     * @return the full path of the mounted folder in the container
+     * @return true or false indicating whether or not the operation was successful.
      */
-    std::string bindMountFolderInContainer(const ContainerID containerID,
+    bool bindMountFolderInContainer(const ContainerID containerID,
                                                 const std::string &pathInHost,
                                                 const std::string &subPathInContainer,
-                                                bool readOnly);
+                                                bool readOnly,
+                                                std::string& pathInContainer);
```
tobsan: Move the &

agent/src/softwarecontaineragentadaptor.cpp
```diff
@@ -11,21 +11,24 @@ SoftwareContainerAgentAdaptor::SoftwareContainerAgentAdaptor(SoftwareContainerAg
...
-int32_t SoftwareContainerAgentAdaptor::LaunchCommand(
+void SoftwareContainerAgentAdaptor::LaunchCommand(
     const int32_t &containerID,
     const uint32_t &userID,
     const std::string &commandLine,
     const std::string &workingDirectory,
     const std::string &outputFile,
-    const std::map<std::string, std::string> &env)
+    const std::map<std::string, std::string> &env,
+    int32_t& pid,
```
tobsan: Move the &

agent/src/softwarecontaineragentadaptor.cpp
```diff
@@ -55,23 +58,26 @@ bool SoftwareContainerAgentAdaptor::ShutDownContainerWithTimeout(
...
 }
 
-std::string SoftwareContainerAgentAdaptor::BindMountFolderInContainer(
+void SoftwareContainerAgentAdaptor::BindMountFolderInContainer(
     const int32_t &containerID,
     const std::string &pathInHost,
     const std::string &subPathInContainer,
-    const bool &readOnly)
+    const bool &readOnly,
+    std::string& pathInContainer,
```
tobsan: Move the &

agent/src/softwarecontaineragentadaptor.cpp
```diff
@@ -81,9 +87,11 @@ bool SoftwareContainerAgentAdaptor::SetCapabilities(
     return m_agent.setCapabilities(containerID, capabilities);
 }
 
-int32_t SoftwareContainerAgentAdaptor::CreateContainer(const std::string &config)
+void SoftwareContainerAgentAdaptor::CreateContainer(const std::string &config,
+                                                    int32_t& containerID,
+                                                    bool& success)
```
tobsan: Move both &'s

agent/src/softwarecontaineragentadaptor.h
```diff
@@ -33,34 +33,39 @@ class SoftwareContainerAgentAdaptor :
...
+    void LaunchCommand(const int32_t &containerID,
                           const uint32_t &userID,
                           const std::string &commandLine,
                           const std::string &workingDirectory,
                           const std::string &outputFile,
-                          const std::map<std::string,
-                          std::string> &env);
+                          const std::map<std::string, std::string> &env,
+                          int32_t& pid,
+                          bool& success);
```
tobsan: Move both &'s

agent/src/softwarecontaineragentadaptor.h
```diff
@@ -33,34 +33,39 @@ class SoftwareContainerAgentAdaptor :
...
-    std::string BindMountFolderInContainer(const int32_t &containerID,
-                                           const std::string &pathInHost,
-                                           const std::string &subPathInContainer,
-                                           const bool &readOnly) override;
+    void BindMountFolderInContainer(const int32_t &containerID,
+                                    const std::string &pathInHost,
+                                    const std::string &subPathInContainer,
+                                    const bool &readOnly,
+                                    std::string& pathInContainer,
+                                    bool& success) override;
```
tobsan: Move both &'s

agent/src/softwarecontaineragentadaptor.h
```diff
@@ -33,34 +33,39 @@ class SoftwareContainerAgentAdaptor :
...
+    bool SetGatewayConfigs(const int32_t &containerID,
                            const std::map<std::string, std::string> &configs) override;
 
     bool SetCapabilities(const int32_t &containerID,
                          const std::vector<std::string> &capabilities) override;
 
-    int32_t CreateContainer(const std::string &config) override;
+    void CreateContainer(const std::string &config,
+                         int32_t& containerID,
+                         bool& success) override;
```
tobsan: Move both &'s

doc/chapters/api/index.rst
```diff
@@ -33,8 +33,8 @@ Controls availability of com.pelagicore.SoftwareContainerAgent interface
 
 CreateContainer
 ---------------
-Creates a new container and returns created container id.
-If the call fails -1 will be returned.
+Creates a new container and returns created container id with success.
+If the call fails success parameter will be returned false.
```
breakreturn: Revise for language please, the sentences doesn't read nicely IMHO.

servicetest/networkgateway/test_network_gateway.py
```diff
@@ -110,13 +110,18 @@ def test_network_policy_drop(self):
...
-            sc.start(DATA)
-            sc.set_gateway_config("network", GW_CONFIG_POLICY_DROP)
-            sc.launch_command("python " +
+            success = sc.start(DATA)
+            assert success is True
+            
+            success = sc.set_gateway_config("network", GW_CONFIG_POLICY_DROP)
+            assert success is True
+
+            pid, success = sc.launch_command("python " +
```
breakreturn: In all places where "pid" will not be used, replace it with a "_" (an underscore).

DunderRoffe (approved):

servicetest/networkgateway/test_network_gateway.py
```diff
@@ -110,13 +110,18 @@ def test_network_policy_drop(self):
...
                               sc.get_bind_dir() +
                               "/testhelper.py" +
                               " --test-dir " + sc.get_bind_dir() +
                               " --do-ping example.com")
+            assert success is True
```
breakreturn: I don't like all these asserts on the success value. That's not what the tests are supposed to test in most cases (except for the current caps test perhaps). Is the purpose to make it more clear that something has gone wrong in the test setup? Could that be made clear in some other way that doesn't clutter the tests?

breakreturn (changes requested):
See individual comments

servicetest/capabilities/test_caps.py
```diff
@@ -37,6 +37,7 @@ def logfile_path():
 DATA = {
     Container.CONFIG: '[{"enableWriteBuffer": false}]',
     Container.BIND_MOUNT_DIR: "app",
+    Container.BIND_MOUNT_DIR: "/gateways/app",
```
tobsan, line 39: This looks like a conflict, since you set BIND_MOUNT_DIR twice.

servicetest/networkgateway/test_network_gateway.py
tobsan: If it was possible, it would be nice to do all these asserts as part of the Container class in the test framework. In the service tests, things aren't usually supposed to fail, are they?

Or, we could just ignore the return value if we feel confident about the test setup. But I agree with that it looks bad having them clutter all tests.

tobsan (changes requested):
See comments.

servicetest/networkgateway/test_network_gateway.py
kursatkobya: yes but following snippet seems even worse 
`launched = sc.launch_command("true")`
`assert launched == -1`

So lets keep success return but don't assert them when its not mandatory

tobsan (approved):

PR #180, opened 2016-11-23
Added test for not being able to double mount

DunderRoffe:


tobsan (changes requested):
This is nice, but I would also like to see tests for trying to mount over things like / and /lib (that are already mounted by LXC)

tobsan (approved):

PR #182, opened 2016-11-24
Uses glibmm for cmdline option handling

tobsan:

### DIFF
--- a/agent/src/main.cpp
+++ b/agent/src/main.cpp
@@ -18,25 +18,12 @@
  */
 
 #include "softwarecontaineragentadaptor.h"
+#include <glibmm/optioncontext.h>
 
 LOG_DEFINE_APP_IDS("SCAG", "SoftwareContainer agent");
 
 namespace softwarecontainer {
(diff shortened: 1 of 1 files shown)


breakreturn (approved):

agent/src/main.cpp
```diff
@@ -155,7 +145,7 @@ int main(int argc, char **argv)
     dbusDispatcher.attach(mainContext->gobj());
 
     try {
-        DBus::Connection connection = getBusConnection(useSystemBus);
+        DBus::Connection connection = getBusConnection(!useSessionBus);
```
kursatkobya: it may be better if we use this as `getBusConnection(useSessionBus);`
and change  `DBus::Connection getBusConnection(bool useSystemBus)` to get  an argument as `bool useSessionBus`

kursatkobya (approved):

agent/src/main.cpp
tobsan: Yes, that would be nicer. I will update to use that instead.

PR #184, opened 2016-11-24
Proposal for SC general config

breakreturn:
This is becoming fairly complete. Left to do is to figure out if and when we want mandatory config options, how to support config files in the service-tests, moving all LXC configs (except the template) to the config, and some other minor things. Not sure all of that should go into this change though.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -163,6 +163,12 @@ if(ENABLE_COVERAGE)
     add_coverage()
 endif(ENABLE_COVERAGE)
 
+# Set and define the path to the general config file, used by the agent
+# as default location if nothing else is set on command line when starting
+# the agent.
+set(SC_CONFIG_FILE ${CMAKE_INSTALL_FULL_DATAROOTDIR}/softwarecontainer/softwarecontainer-config)
+add_definitions(-DSC_CONFIG_FILE="${SC_CONFIG_FILE}")
+
 #
 # Network related options
 #
@@ -211,13 +217,20 @@ endif()
 #
 # General configuration file
 #
-configure_file(softwarecontainer-config.h.in softwarecontainer-config.h @ONLY)
-install(FILES ${CMAKE_CURRENT_BINARY_DIR}/softwarecontainer-config.h DESTINATION ${CMAKE_INSTALL_FULL_LIBDIR}/softwarecontainer)
+set(SHARED_MOUNTS_DIR "/tmp/container/")
+set(LXC_CONFIG_PATH "${CMAKE_INSTALL_FULL_SYSCONFDIR}/softwarecontainer.conf")
+set(SERVICE_MANIFEST_DIR "${CMAKE_INSTALL_FULL_SYSCONFDIR}/softwarecontainer/service-manifest.d/")
+set(DEFAULT_SERVICE_MANIFEST_DIR "${CMAKE_INSTALL_FULL_SYSCONFDIR}/softwarecontainer/service-manifest.default.d/")
+
+# NOTE: These definitons are only set because the "middle layer" tests currently need to create a Workspace
+#       with valid values, see the setup code of SoftwareContainerTest test class.
+add_definitions(-DSHARED_MOUNTS_DIR_TESTING="${SHARED_MOUNTS_DIR}")
+add_definitions(-DLXC_CONFIG_PATH_TESTING="${LXC_CONFIG_PATH}")
+
 
 #
 # Add all sub-projects
 #
-
 add_subdirectory(common)
 add_subdirectory(libsoftwarecontainer)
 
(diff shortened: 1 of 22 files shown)


common/config.h
```diff
@@ -0,0 +1,118 @@
...
+
+#include <string>
+#include <exception>
+#include <glibmm.h>
+
+#include "softwarecontainer-common.h"
+
+
+/*
+ * Propose we have this inherit a general SoftwareContainer exception instead
```
tobsan: Why is this not in the softwarecontainer namespace instead?

common/unit-test/config_unittest.cpp
```diff
@@ -0,0 +1,196 @@
...
+
+/*
+ * ConfigTest suite
+ *
+ * This suite tests the Config class. It uses a stubbed loader, the StringConfigLoader, to
+ * provide the config source, and does asserts on the way Config parses the config provided.
+ */
+class ConfigTest : public ::testing::Test
+{
+};
```
tobsan: This fixture class isn't very useful. Consider moving your instantiation of the loader and config objects into the fixture class.

common/unit-test/config_unittest.cpp
```diff
@@ -0,0 +1,196 @@
...
+TEST_F(ConfigTest, IncorrectKeyThrows) {
+    // Well formed config
+    const std::string configString = "[SoftwareContainer]\n"
+                                     "sc.foo=bar\n"
+                                     "sc.preload=2\n";
+
+    std::unique_ptr<ConfigLoaderAbstractInterface> loader(new StringConfigLoader(configString));
+    Config config(std::move(loader));
+
+    ASSERT_THROW(config.getStringValue("SoftwareContainer", "does-not-exist"), SCException::ConfigError);
```
tobsan: You are testing similar things here. You could consider having the test being value parameterized on group and key if there are many of these tests later on.

CMakeLists.txt
```diff
@@ -217,18 +223,20 @@ install(FILES ${CMAKE_CURRENT_BINARY_DIR}/softwarecontainer.pc DESTINATION ${CMA
...
+
 #
 # Add all sub-projects
 #
-
 add_subdirectory(common)
 add_subdirectory(libsoftwarecontainer)
 
 add_subdirectory(agent)
-add_subdirectory(libsoftwarecontaineragent)
```
DunderRoffe: This could be a separate small pr

agent/src/main.cpp
```diff
@@ -105,17 +118,24 @@ int main(int argc, char **argv)
...
+    /* Default values need to be somehting that should not be set explicitly
+     * by the user.
+     */
+    Glib::ustring configPath = "";
+    int preloadCount = -1;
     int userID = 0;
-    bool shutdownContainers = true;
-    int timeout = 1;
+    bool keepContainersAlive = false;
+    int timeout = -1;
```
DunderRoffe: Is hanging indefinitely really a sensible default?

libsoftwarecontainer/include/workspace.h
```diff
@@ -43,11 +43,11 @@ class Workspace :
...
      * @param containerShutdownTimeout The timeout time for the softwarecontainer
+     *
+     * TODO: Remove the abouve param docs if the resulting workspace design
+     *       means we use the default constructor.
      */
-    Workspace(bool enableWriteBuffer = false,
-              const std::string &containerRootDir = SOFTWARECONTAINER_DEFAULT_WORKSPACE,
-              const std::string &containerConfigPath = SOFTWARECONTAINER_DEFAULT_CONFIG,
-              unsigned int containerShutdownTimeout = 1);
+    Workspace();
```
DunderRoffe, line 50: I think Workspace should be removed entirely

CMakeLists.txt
breakreturn: Yes I know. It's just that I have to fix tests and build errors etc., and I know we are to remove it. I guess this will go away during a rebase soon.

agent/src/main.cpp
breakreturn: No it's not. But what would be a sensible default? Some of the configs I think will require us to do sanity checks after the priority and logic of choosing source is done. I.e. this default should only be used to check if the user did something explicit but never propagate downwards.

libsoftwarecontainer/include/workspace.h
breakreturn: I was hoping we could do that separately, there's so much scope creep in this change anyway ;)

CMakeLists.txt
DunderRoffe: Update: Done in another pr by @tobsan

libsoftwarecontainer/include/workspace.h
DunderRoffe: sure

tobsan (changes requested):
1) Fix conflicts
2) Bump version

Looks good though.

DunderRoffe (approved):

PR #186, opened 2016-11-25
Support for default capabilities

DunderRoffe:


breakreturn (approved):

agent/src/capability/filteredconfigstore.cpp
```diff
@@ -0,0 +1,51 @@
...
+
+FilteredConfigStore::~FilteredConfigStore()
+{
+}
+
+GatewayConfiguration FilteredConfigStore::configsByID(const std::vector<std::string> &capIDs) const
+{
+    GatewayConfiguration conf;
+    for (auto &id : capIDs) {
+        conf.append(configByID(id));
```
tobsan, line 34: Check the return value here before appending, since configByID can fail?

agent/src/capability/filteredconfigstore.h
```diff
@@ -0,0 +1,49 @@
...
+     * @param capIDs vector of strings representating Capability IDs
+     * @return A vector of GatewayConfigurations matching the given Capability ids.
+     */
+    GatewayConfiguration configsByID(const std::vector<std::string> &capIDs) const;
+
+    /**
+     * @brief Returns all Gateway configurations for a certain Capability.
+     *
+     * @param capID a string representation of the Capability ID
+     * @return GatewayConfiguration maped by the given ID.
```
tobsan: mapped*

agent/src/softwarecontaineragent.cpp
```diff
@@ -423,14 +424,15 @@ bool SoftwareContainerAgent::updateGatewayConfigs(const ContainerID &containerID
...
-                        << std::to_string(containerID) << " does not exist";
-            return false;
-        }
+    auto gatewayConfigs = m_defaultConfigStore->configs();
+    auto filteredConfigs = m_filteredConfigStore->configsByID(capabilities);
+    gatewayConfigs.append(filteredConfigs);
+
+    // Update container gateway configuration
+    if (!updateGatewayConfigs(containerID, gatewayConfigs)) {
+        log_error() << "Could noteset gateway configuration for capability"
```
tobsan: typo

agent/src/softwarecontaineragent.cpp
```diff
@@ -423,14 +424,15 @@ bool SoftwareContainerAgent::updateGatewayConfigs(const ContainerID &containerID
...
+                    << std::to_string(containerID) << " does not exist";
```
tobsan: Is this the only case where it would fail?

common/gatewayconfig.cpp
```diff
@@ -0,0 +1,83 @@
...
+
+#include "gatewayconfig.h"
+
+GatewayConfiguration::GatewayConfiguration()
+{
+}
+
+GatewayConfiguration::~GatewayConfiguration()
+{
+    for (auto &i : m_configMap) {
```
tobsan: i is a bad variable name imho

doc/chapters/integration-guidelines/index.rst
```diff
@@ -19,8 +19,9 @@ Service manifests
 For details about format and content of service manifests, see :ref:`Service manifests <service-manifests>`
 and :ref:`Gateways <gateways>`.
 
-Service manifests should be installed in ``/etc/softwarecontainer/caps.d/<service-name>`` and are
-expected to be of type JSON, including the "json" file extension.
+Service manifests should be installed in either ``/etc/softwarecontainer/caps.d/<service-name>`` or
+``/etc/softwarecontainer/caps.default.d/<service-name>`` depending on if it should be applied by default
+or not. Each manifest are expected to be of type JSON, including the "json" file extension.
```
tobsan, line 24: Shouldn't this be ${FULL_SYSCONFDIR}/softwarecontainer/...? We can do substitutions from cmake to the sphinx config file for things like this, please do.

tobsan (changes requested):

agent/src/capability/filteredconfigstore.cpp
DunderRoffe: Is it not better to return all found things instead of empty when something fails?

agent/src/capability/filteredconfigstore.h
DunderRoffe: Fixed

agent/src/softwarecontaineragent.cpp
DunderRoffe: fixed

agent/src/softwarecontaineragent.cpp
DunderRoffe: probably not, removing extra row

common/gatewayconfig.cpp
DunderRoffe: Changing to "it" then since. It is an iterator after all

agent/src/capability/filteredconfigstore.cpp
tobsan: I think we should aim for failing early in case of errors, so I would go for that - wanting an unexisting cap is an error imho.

PR #188, opened 2016-11-30
Major changes to the DBus API

tobsan:
This removes some unused/obsolete methods, renames a few to more
logical names, and adds one new method to list containers.


kursatkobya (approved):
looks good

DunderRoffe (changes requested):
Please update the agent as well

tobsan:
@DunderRoffe Done, thanks for pointing it out.

DunderRoffe (approved):

agent/src/softwarecontaineragentadaptor.h
```diff
@@ -33,45 +33,37 @@ class SoftwareContainerAgentAdaptor :
...
-    void CreateContainer(const std::string &config,
-                         int32_t &containerID,
-                         bool &success) override;
-
-    bool SetContainerName(const int32_t &containerID, const std::string &name) override;
-
-    void Ping() override;
-
-    bool WriteToStdIn(const uint32_t &processID, const std::vector<uint8_t> &bytes) override;
+    void Create(const std::string &config,
```
breakreturn: It's a bit inconsistent when the lines are broken around here.

servicetest/testframework/lib.py
```diff
@@ -118,50 +118,51 @@ def start(self, data):
...
             print "Failed to create container"
             return False
         
-        self.__bind_dir, success = self.__bindmount_folder_in_container(data[Container.HOST_PATH],
-                                                               data[Container.BIND_MOUNT_DIR],
-                                                               data[Container.READONLY])
+        success = self.__bindmount_folder_in_container(data[Container.HOST_PATH],
+                                                       data[Container.BIND_MOUNT_DIR],
+                                                       data[Container.READONLY])
         if False == success:
```
breakreturn: I think it's more typical to Python to check this like "if success is False". There's no risk of assignment then either and it doesn't have to be set up backwards looking.

servicetest/testframework/lib.py
```diff
@@ -118,50 +118,51 @@ def start(self, data):
...
             print "Failed to retrieve binded folder"
             return False
-
-        return True
+        else:
+            self.__bind_dir = data[Container.BIND_MOUNT_DIR]
```
breakreturn, line 138: This is not necessarily the same as the original code w.r.t. the value in __bind_dir is it? Previously it was set by the external component which means we cover some actual behavior there, now we are reusing the value the user set. This should of course be the same, but this is a test after all.

servicetest/testframework/lib.py
DunderRoffe: the pythonic way would be "if not success"

servicetest/testframework/lib.py
```diff
@@ -118,50 +118,51 @@ def start(self, data):
...
         """
         success = self.__create_container(data[Container.CONFIG])
         if False == success:
             print "Failed to create container"
             return False
         
-        self.__bind_dir, success = self.__bindmount_folder_in_container(data[Container.HOST_PATH],
-                                                               data[Container.BIND_MOUNT_DIR],
-                                                               data[Container.READONLY])
+        success = self.__bindmount_folder_in_container(data[Container.HOST_PATH],
```
DunderRoffe: This should be renamed as well to remove the trace of the old name.

servicetest/testframework/lib.py
breakreturn: Ooooh a Pythonic battle. My version is better than yours.

servicetest/testframework/lib.py
tobsan: Since you need to provide an absolute path to which to mount, we don't have that echoed back by the agent, so if the mounting was successful it was indeed mounted to where we asked it to. That's why it is only happening if mounting was successful.

tobsan:
Updated according to discussion, except for the bindmount value in the testframework python lib.

servicetest/testframework/lib.py
breakreturn: Ah I see! Well that sounds good. Maybe a comment here about this would be helpful for the reader?

breakreturn (approved):

PR #189, opened 2016-11-30
Add whitelisting test for NetworkGateway

kursatkobya:


tobsan (changes requested):
It would be nice to see a test that sets the same host but with different protocols after each other, so that merging of configs are properly handled.

kursatkobya:
Ping cannot be occur without dns lookup thats why tcp and udp protocols on port 53 are needed for any ping operations. 

So if the iana.org is pingable after adding that configuration that means it is perfectly merged with previous configuration which adds port 53 on tcp and udp protocols to allow lost. 

Therefore i believe this test fully covers about merging and white-listing.

tobsan:
@kursatkobya Ah yes, looking at the whole file made it clearer. Could you add some comments on why exactly port 53 is used, so that it doesn't look random for someone not familiar with DNS? After that I'm fully satisfied :)

tobsan (approved):

servicetest/networkgateway/test_network_gateway.py
```diff
@@ -187,6 +203,53 @@ def test_allow_ping_only_for_specific_domain(self):
         finally:
             sc.terminate()
 
+    def test_whitelist(self):
+        """ Tests the whitelisting feauture.
```
breakreturn: "Feature"

servicetest/networkgateway/test_network_gateway.py
```diff
@@ -187,6 +203,53 @@ def test_allow_ping_only_for_specific_domain(self):
...
+            is_pingable = helper.ping_result()
+            assert is_pingable is True
+            helper.remove_file()
+
+            sc.launch_command("python " +
+                              sc.get_bind_dir() +
+                              "/testhelper.py" +
+                              " --test-dir " + sc.get_bind_dir() +
+                              " --do-ping pelagicore.com")
+            time.sleep(10)
```
breakreturn, line 245: Could you add some comments describing when we do long-ish sleeps? I think it should be motivated so anyone maintaining this doesn't need to think too much about it.

kursatkobya: It was already commented in the service test class TestNetworkRules
```

    """ This suite tests that whether NetworkGateway can parse given configuration and applies it to be used
        in SoftwareContainer with expected results.

        timeouts are:
             2 seconds for expected connectivity
            10 for expected output discconnectivity
            15 for expected input connectivity

        these durations will be shorten with -i option when possible
    """
```

PR #190, opened 2016-12-01
Envgateway supports prepend+append 

tobsan:
Some changes in filegateway were done as a consequence of this.

### DIFF
--- a/doc/chapters/gateways/index.rst
+++ b/doc/chapters/gateways/index.rst
@@ -271,10 +271,15 @@ a log warning.
 
 The configuration may also, optionally, specify the following parameters:
 
-- ``append`` (bool) If the environment variable is already defined by the gateway, append the new value to the value already defined. Defaults to false.
+- ``mode``: A string that can be either:
+    - ``set``: set the variable (this is the default, so one does not usually need to set it)
+    - ``append``: append the value given to the previous value for the given variable.
+    - ``prepend``: prepend the value given to the previous value for the given variable.
+- ``separator``: A string that will be squeezed in between the old and new value if one wants to
+                 prepend or append to a variable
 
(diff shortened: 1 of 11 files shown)


doc/chapters/gateways/index.rst
```diff
@@ -271,10 +271,15 @@ a log warning.
 
 The configuration may also, optionally, specify the following parameters:
 
-- ``append`` (bool) If the environment variable is already defined by the gateway, append the new value to the value already defined. Defaults to false.
+- ``mode``: A string that can be either:
+    - ``set``: set the variable (this is the default, so one does not usually need to set it)
+    - ``append``: append the value given to the previous value for the given variable.
+    - ``prepend``: prepend the value given to the previous value for the given variable.
+- ``separator``: A string that will be squeezed in between the old and new value if one wants to
```
breakreturn, line 278: Squeezed? Could use more boring but "correct" language :)

doc/chapters/gateways/index.rst
```diff
@@ -297,8 +302,9 @@ There is also the possibility to append to an already defined variable::
     [
         {
             "name": "SOME_ENVIRONMENT_VARIABLE",
-            "value": ":/some/path",
-            "append": true
+            "value": "/some/path",
+            "mode": "append"
```
DunderRoffe, line 306: Make "append" big letters or non case sensitive

DunderRoffe (changes requested):

doc/chapters/gateways/index.rst
tobsan: Making it case insensitive seems reasonable, I'm doing that.

DunderRoffe (approved):
Awesomesauce

PR #191, opened 2016-12-01
Removes the agent library

tobsan:
As a consequence of this, the pc file was moved to be located
inside the libsoftwarecontainer directory, and also depend
on the common lib.


breakreturn:
Does this deserve some bump in version? Not sure what could be considered API here...

tobsan:
I think it does, since this is a not-backwards compatible change. I can update the version field.

DunderRoffe (approved):

tobsan:
Version bumped.

PR #192, opened 2016-12-01
adds initial CONTRIBUTING.md

tobsan:

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,24 @@
+How to contribute
+=================
+We are happy to accept contributions. To accept a contribution, we require two things:
+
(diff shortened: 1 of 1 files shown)


tobsan:
I'd love some comments on this.

DunderRoffe:
Maybe we should add why we do this and that it is a temporary solution.

jeremiah:
I think this matches the vision we had in our internal discussions. I think the implementation is great, my only reservation is the language of the document but that we can iterate on.

jeremiah:
(previous comments were for something else.) 

@DunderRoffe I agree -- we should note that this is just a temporary solution.

breakreturn (approved):

jeremiah (approved):
Looks good. :+1:

PR #193, opened 2016-12-02
Correcting cmake files for Agent unit tests

thenor:
SoftwareContainerAgent and configstore unit tests were not being run
because of a (silent) fault in the cmake file.
When enabled the unit tests for configstore segfaulted, so also fixed
an incorrect free (for json_t).


agent/unit-test/configstore_unittest.cpp
```diff
@@ -96,12 +85,12 @@ TEST_F(ConfigStoreTest, constructorDir) {
 /* Constructing a FilteredConfigStore with a directory path,
  * even if all files can not be parsed, should not throw an exception.
  */
-/* This test fails intermittently, not clear why
-TEST_F(FilteredConfigStoreTest, constructorDir2) {
+//This test fails intermittently, not clear why
+TEST_F(ConfigStoreTest, constructorDir2) {
     // No config files, but ok dir
     ASSERT_NO_THROW(FilteredConfigStore("/home/vagrant/softwarecontainer/"));
```
breakreturn: How come an absolute path works here? If it's not actually used, perhaps it's better to change it to something more obviously irrelevant. If we rely on it, it will fail if run in another environment.

breakreturn (changes requested):

agent/src/capability/baseconfigstore.cpp
```diff
@@ -204,11 +204,19 @@ std::vector<std::string> BaseConfigStore::fileList(const std::string &dirPath)
...
+
+    std::string prefix = ".json";
+    size_t lastFivePos = filename.size()-5;
+
+    std::string lastFiveLetters = filename.substr(lastFivePos);
+
+    if (prefix.compare(lastFiveLetters) != 0) {
+        return false;
+    }
+
```
tobsan: This is a very roundabout way to check if the end is ".json". How about this:
`return filename.substr(filename.size()-5) == ".json"`
I guess the check for string length is reasonable though.

tobsan (commented):
Looks good in general, see my comment about the file ending check.

agent/unit-test/configstore_unittest.cpp
thenor: I agree. We shouldn't have an absolute path here. As the description suggests the purpose of the test was to use a directory we had the rights to read from and which contains many files, preferably no json files.

tobsan (approved):

breakreturn (approved):

PR #194, opened 2016-12-02
NetworkGW: test that no rules means DROP

tobsan:


kursatkobya (approved):
Although this is tested partly in white-listing test, it is always nice to be straightforward. Cool

DunderRoffe (approved):

PR #198, opened 2016-12-05
Removes the env var options from filegateway

tobsan:
Since we mount with absolute paths, there is no longer a need
for env variables that hold the actual path. If such behaviour
is desired, one can combine usage of the file gateway with the
env gateway instead.

Also reformatted the gateways doc chapter to cap lines to 100 chars.

### DIFF
--- a/libsoftwarecontainer/unit-test/softwarecontainerlib_unittest.cpp
+++ b/libsoftwarecontainer/unit-test/softwarecontainerlib_unittest.cpp
@@ -199,53 +199,27 @@ TEST_F(SoftwareContainerApp, FileGatewayReadOnly) {
     job0.start();
     ASSERT_EQ(job0.wait(), EXISTENT);
 
-    // Create two temporary files
+    // Create a temporary file, and verify that we could indeed create it.
     char tempFilename1[] = "/tmp/fileGatewayXXXXXX";
-    char tempFilename2[] = "/tmp/fileGatewayXXXXXX";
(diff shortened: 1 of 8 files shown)


breakreturn (approved):

libsoftwarecontainer/unit-test/softwarecontainerlib_unittest.cpp
```diff
@@ -199,53 +199,27 @@ TEST_F(SoftwareContainerApp, FileGatewayReadOnly) {
     job0.start();
     ASSERT_EQ(job0.wait(), EXISTENT);
 
-    // Create two temporary files
+    // Create two temporary files, and verify that we could indeed create them.
```
thenor: Do you in fact create 2 files?

libsoftwarecontainer/unit-test/softwarecontainerlib_unittest.cpp
```diff
@@ -324,47 +274,24 @@ TEST_F(SoftwareContainerApp, FileGatewayReadWrite) {
 
     // Create two temporary files
```
thenor: There seems to be only one file here, as above.

libsoftwarecontainer/unit-test/softwarecontainerlib_unittest.cpp
tobsan: Well spotted, we don't! Updated.

thenor (approved):

PR #200, opened 2016-12-05
Adds the temperature service/client example

tobsan:
This is a somewhat fully-featured example utilizing the dbus gateway.


examples/temperature/temperature-service-console-client/temperatureserviceinterface.cpp
```diff
@@ -0,0 +1,58 @@
...
+
+double TemperatureServiceInterface::getTemperature()
+{
+    return GetTemperature();
+}
+
+bool TemperatureServiceInterface::setTemperature(const double &temperature)
+{
+    SetTemperature(temperature);
+    return true; // Always succesfull
```
DunderRoffe: What is the purpose of returning then?

examples/temperature/temperature-service/configs/temperature-service.json
```diff
@@ -0,0 +1,86 @@
+{
```
DunderRoffe, line 1: Are there no license for the service manifests?
At least there should be copyright.

DunderRoffe (commented):
Seems generally fine. However, is this maintained elsewhere as well? If so it might be better to use a submodule. Furthermore, this project does not follow our coding guideline and tools. Might be nice to make it use ivi-logging maybe?

examples/temperature/temperature-service/configs/temperature-service.json
tobsan: Good point, will add that.

tobsan:
@DunderRoffe It used to be a closed-source example to showcase softwarecontainer. My view is that we should kill off the old internal repo and only keep the example here instead.

DunderRoffe:
Ok sounds nice. But then I would argue even more that it should use the same codestyle as the rest of SC.

examples/temperature/temperature-service/configs/temperature-service.json
tobsan: Actually, it turns out you can't have comments in json files.

examples/temperature/temperature-service-console-client/main.cpp
```diff
@@ -0,0 +1,55 @@
...
+    // Setup DBus
+    DBus::BusDispatcher dispatcher;
+    DBus::default_dispatcher = &dispatcher;
+    DBus::Connection bus = DBus::Connection::SystemBus();
+
+    // The interface class needs a path to a logfile because we
+    // can't access stdout.
+    std::string homeDir = getenv("HOME");
+    if (homeDir.empty()) {
+        std::cout << "No $HOME set, can't run" << std::endl;
```
thenor, line 37: Do we need to write to std::cout?

examples/temperature/temperature-service-console-client/main.cpp
```diff
@@ -0,0 +1,55 @@
...
+
+int main()
+{
+    // Setup DBus
+    DBus::BusDispatcher dispatcher;
+    DBus::default_dispatcher = &dispatcher;
+    DBus::Connection bus = DBus::Connection::SystemBus();
+
+    // The interface class needs a path to a logfile because we
+    // can't access stdout.
```
thenor: Is this something that is still true?

examples/temperature/temperature-service/src/temperatureservice.h
```diff
@@ -0,0 +1,58 @@
...
+class TemperatureServiceToDBusAdapter;
+
+/*!
+ * The TemperatureService class is where you implement
+ * your service, this is the work horse.
+ */
+
+class TemperatureService {
+public:
+    /*! Constructor
```
thenor: As has been commented early, it would be good to adapt to the SC coding format, also for the comments.

examples/temperature/temperature-service/configs/temperature-service.json
DunderRoffe: Oh yeah, forgot about that :/ Never mind then

examples/temperature/temperature-service-console-client/temperatureserviceinterface.cpp
tobsan: Changed this to void.

examples/temperature/temperature-service-console-client/main.cpp
tobsan: We don't need to write anywhere really, but to minimize dependencies (I guess) we don't pull ivi-logging in here.

PR #204, opened 2016-12-06
implements whitelisting in filegateway

tobsan:


libsoftwarecontainer/unit-test/filegatewayparser_unittest.cpp
```diff
@@ -139,6 +140,109 @@ TEST_F(FileGatewayParserTest, TestEmptyPathInContainer) {
...
+    ASSERT_EQ(ReturnCode::SUCCESS, parser.parseConfigElement(configJSON, result));
+    ASSERT_EQ(ReturnCode::SUCCESS, parser.matchEntry(result, settings));
+    ASSERT_EQ(1u, settings.size());
+}
+
+/*
+ * Using the same paths but with different values on read-only is fine, since we
+ * will use the most permissive setting.
+ */
+TEST_F(FileGatewayParserTest, TestSameConfigReadWrite) {
```
kursatkobya: test name is sounds like misleading. Maybe it should be something like 
`TEST_F(FileGatewayParserTest, ParameterReadonlyOverload)`
or something like that

doc/chapters/gateways/index.rst
```diff
@@ -341,6 +341,12 @@ Configuration
 The paths inside the container has to be absolute. There is a check for not mounting over already
 existing mount paths.
 
+It is possible to supply the same host path in several configuration snippets to get the same file
+mounted onto several locations in the container. It is however not possible to supply the same
```
DunderRoffe: "It is howerver ..." -> "However, it is ..."

scripts/run-tests.sh
```diff
@@ -81,6 +81,12 @@ retval=$(($retval+$?))
     --gtest_output=xml:softwarecontainerlib_unittest_result.xml
 retval=$(($retval+$?))
 
+if [ $retval -ne 0 ]; then
+	echo "********************************"
+	echo "*** ONE OR MORE TESTS FAILED ***"
+	echo "********************************"
```
DunderRoffe: Should this be included? Seems awfully verbose.

DunderRoffe (changes requested):

libsoftwarecontainer/unit-test/filegatewayparser_unittest.cpp
```diff
@@ -139,6 +140,109 @@ TEST_F(FileGatewayParserTest, TestEmptyPathInContainer) {
...
+    ASSERT_EQ(ReturnCode::SUCCESS, parser.matchEntry(result, settings));
+    ASSERT_EQ(1u, settings.size());
+    ASSERT_EQ(settings.at(0).readOnly, false);
+}
+
+/*
+ * Testing same host path but different container paths is fine, we will
+ * mount the same file to two places, no problems.
+ */
+TEST_F(FileGatewayParserTest, TestSameHostPathDifferentContainerPath) {
```
kursatkobya: I think TestSame template is not good for those. ParameterHostPathOverload could be better fit imho.
And also since the test fixture is already mentioning this is a test maybe we can remove test word from test names. I mean wouldn't be better to filter like `FileGatewayParserTest.SameHostPathDifferentContainerPath` instead of `FileGatewayParserTest.TestSameHostPathDifferentContainerPath`

scripts/run-tests.sh
tobsan: The point of having this is that since we have many tests in different test suites, there is no way to get output about that some early test failed unless you check the exit status of the test runner. This is pure developer convenience imo.

libsoftwarecontainer/unit-test/filegatewayparser_unittest.cpp
tobsan: Actually, I agree with that.

tobsan:
Updated after review comments.

scripts/run-tests.sh
DunderRoffe: OK. But in that case it should be done properly. Would be nice to print out the name of each failed test suite and have the text in red. :)

kursatkobya (approved):
looks good now

DunderRoffe (approved):

PR #207, opened 2016-12-07
Implement white-listing to WaylandGateway

kursatkobya:


libsoftwarecontainer/src/gateway/waylandgateway.cpp
```diff
@@ -40,6 +41,11 @@ WaylandGateway::~WaylandGateway()
 
 ReturnCode WaylandGateway::readConfigElement(const json_t *element)
 {
+    if (m_enabled == true) {
+        log_debug() << "WaylanGateway is already enabled";
+        return ReturnCode::SUCCESS;
```
tobsan: This is imho not the desired behaviour. With this, one can supply bad configurations once there is one valid configuration sent with enabled to true. We should ALWAYS read the config to make sure it is well-formed. In this case it makes sense to read it into a temporary var first which one can then or with `m_enabled`.

tobsan (changes requested):

tobsan (approved):

DunderRoffe (approved):

doc/chapters/gateways/index.rst
```diff
@@ -468,3 +468,5 @@ Example configuration enabling Wayland::
         { "enabled": true }
     ]
 
+The configuration of Wayland gateway can be set several times, but once it has been set to ``true``
```
DunderRoffe: "Wayland gateway" -> "the Wayland gateway"

libsoftwarecontainer/unit-test/softwarecontainerlib_unittest.cpp
```diff
@@ -54,14 +54,18 @@ class SoftwareContainerApp : public SoftwareContainerTest
     }
 };
 
-TEST_F(SoftwareContainerApp, TestWayland) {
+TEST_F(SoftwareContainerApp, TestWaylandWhitelist) {
 
     GatewayConfiguration config;
     std::string configStr = "[ { \"enabled\" : true } ]";
+    std::string configStr2 = "[ { \"enabled\" : false } ]";
```
thenor: I think an even more explanatory name would be e.g. "configStrFalse".

PR #209, opened 2016-12-07
Default capabilities

thenor:
Default capabilities and service tests for configstore/Service Manifests/capabilities.

### DIFF
--- a/agent/src/main.cpp
+++ b/agent/src/main.cpp
@@ -107,17 +107,21 @@ int main(int argc, char **argv)
     timeoutOpt.set_arg_description("<seconds>");
     timeoutOpt.set_description("Timeout in seconds to wait for containers to shutdown, defaults to 1");
 
-    Glib::OptionEntry capsDirOpt;
-    capsDirOpt.set_long_name("caps-dir");
-    capsDirOpt.set_short_name('c');
-    capsDirOpt.set_arg_description("<filepath>");
-    capsDirOpt.set_description("Path to a file or directory where service manifest(s) exist, defaults to" + std::string(CAPABILITIES_DIR));
+    Glib::OptionEntry serviceManifestDirOpt;
+    serviceManifestDirOpt.set_long_name("manifest-dir");
+    serviceManifestDirOpt.set_short_name('m');
+    serviceManifestDirOpt.set_arg_description("<filepath>");
+    serviceManifestDirOpt.set_description("Path to a file or directory where service manifest(s) "
+                                          "exist, defaults to " +
+                                          std::string(SERVICE_MANIFEST_DIR_OPT));
 
-    Glib::OptionEntry defaultCapsDirOpt;
-    defaultCapsDirOpt.set_long_name("default-caps-dir");
-    defaultCapsDirOpt.set_short_name('d');
-    defaultCapsDirOpt.set_arg_description("<filepath>");
-    defaultCapsDirOpt.set_description("Path to a file or directory where default service manifest(s) exist, defaults to " + std::string(DEFAULT_CAPABILITIES_DIR));
+    Glib::OptionEntry defaultServiceManifestDirOpt;
+    defaultServiceManifestDirOpt.set_long_name("default-manifest-dir");
+    defaultServiceManifestDirOpt.set_short_name('d');
+    defaultServiceManifestDirOpt.set_arg_description("<filepath>");
+    defaultServiceManifestDirOpt.set_description("Path to a file or directory where default "
+                                                 "service manifest(s) exist, defaults to " +
+                                                 std::string(DEFAULT_SERVICE_MANIFEST_DIR_OPT));
 
     Glib::OptionEntry sessionBusOpt;
     sessionBusOpt.set_long_name("session-bus");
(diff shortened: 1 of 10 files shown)


agent/src/main.cpp
```diff
@@ -177,9 +188,13 @@ int main(int argc, char **argv)
      * over the static configuration.
      */
     std::map<std::string, std::string> stringOptions;
-    if (servicemanifest != "") {
+    if (manifestDir != "") {
```
thenor: serviceManifestDir or manifestDir?
If I use the first option and re-use the same for default it will be a very long variable name.

servicetest/capabilities/test_caps.py
```diff
@@ -41,31 +49,67 @@ def logfile_path():
...
-        the gatways or their configs, which is done in more specific test suites, but rather
-        to make sure the capabilities result in the correct gatway configs being applied.
+        The tests use the Agent D-Bus interface to enable gateway configurations,
+        for a specified list of capabilities. Default gateway configurations will
+        always be applied.
 
-        TODO: Above is not currently true, these tests are only for the currently stubbed
-              implementation in SC.
-        TODO: How to piggy-back on the other tests that already do these things?
+        The purpose is not to test the gatways or their configs, which is done in
```
thenor: Typo "gatways" fixed.

servicetest/capabilities/test_caps.py
```diff
@@ -41,31 +49,67 @@ def logfile_path():
...
+        more specific test suites, but rather to make sure the capabilities result
+        in the correct gatway configs being applied.
```
thenor: Another "gatway".

agent/src/main.cpp
```diff
@@ -107,11 +107,20 @@ int main(int argc, char **argv)
...
-    manifestOpt.set_long_name("manifest");
-    manifestOpt.set_short_name('m');
-    manifestOpt.set_arg_description("<filepath>");
-    manifestOpt.set_description("Path to a file or directory where service manifest(s) exist, defaults to nothing");
+    Glib::OptionEntry manifestDirOpt;
+    manifestDirOpt.set_long_name("manifest-dir");
+    manifestDirOpt.set_short_name('m');
+    manifestDirOpt.set_arg_description("<filepath>");
+    manifestDirOpt.set_description("Path to a file or directory where service manifest(s) exist, "
+                                   "defaults to " + std::string(SERVICE_MANIFEST_DIR_TESTING));
```
breakreturn: These defines should not be here, they are explicitly exported for testing purposes. When the next config commit comes in there will be defines for these, but if they are needed now please add another definition, i.e. the some one but without "_TESTING".

agent/src/main.cpp
```diff
@@ -127,7 +136,8 @@ int main(int argc, char **argv)
     int userID = 0;
     bool keepContainersAlive = false;
     int timeout = -2;
-    Glib::ustring servicemanifest = "";
+    Glib::ustring manifestDir = SERVICE_MANIFEST_DIR_TESTING;
+    Glib::ustring defaultManifestDir = DEFAULT_SERVICE_MANIFEST_DIR_TESTING;
```
breakreturn: Same goes here.

servicetest/testframework/lib.py
```diff
@@ -171,20 +172,29 @@ class SoftwareContainerAgentHandler():
...
         self.__rec.start()
         self.__rec.wait_until_setup_is_done()
 
         # Starting softwarecontainer-agent
         # TODO: This doesn't work if the user pass 'None' as log_file_path
         assert log_file_path is not None
-        self.__agent = subprocess.Popen("softwarecontainer-agent", stdout=self.__log_file, stderr=self.__log_file)
+        print(cmd)
+        self.__agent = subprocess.Popen(cmd, stdout=self.__log_file, stderr=self.__log_file)
+        print(self.__agent)
```
tobsan: Is there a good reason for printing these?

servicetest/networkgateway/.gitignore
```diff
@@ -1 +1,2 @@
 test.log
+env_vars_dump.json
```
tobsan: Don't we have a separate story somewhere for handling this?

tobsan (commented):
I think this looks good, just had some comments.

servicetest/testframework/lib.py
thenor: Looks like perhaps @DunderRoffe added some debug printouts, guessing they shouldn't be kept.

breakreturn (changes requested):

servicetest/testframework/lib.py
DunderRoffe: Sorry that is debug stuff. Please remove it

agent/src/main.cpp
thenor: Oh I think I misunderstood. I changed it. Hopefully I have understood it now.

servicetest/networkgateway/.gitignore
DunderRoffe: I added this because I was tired of it getting generated

servicetest/testframework/lib.py
thenor: Done.

servicetest/networkgateway/.gitignore
thenor: Removed.

agent/src/main.cpp
tobsan: Descriptive variable name > short variable name. I would go for serviceManifestDir and defaultServiceManifestDir, even though they are long. The directories and their names and functions are complicated enough as it is.

tobsan (commented):
Just one small comment on variable names.

DunderRoffe (approved):

PR #210, opened 2016-12-07
Uses Glib::getenv which provides nicer error check

tobsan:
Glibmm provides a higher-level function to check environment variables,
which also can set a bool if they are set or not, letting us get rid of
all those nullptr or NULL comparisons.


breakreturn (approved):

DunderRoffe (approved):
Although I am not a fan of the C way of setting the bools in this case, I must admit it is more tidy

PR #212, opened 2016-12-08
Change test names for EnvironmentGateway

kursatkobya:
- small changes for consistency

### DIFF
--- a/libsoftwarecontainer/src/gateway/environment/envgateway.h
+++ b/libsoftwarecontainer/src/gateway/environment/envgateway.h
@@ -39,6 +39,5 @@ public:
     bool teardownGateway();
 
 private:
-    bool m_enabled = false;
     EnvironmentVariables m_variables;
 };
--- a/libsoftwarecontainer/unit-test/envgatewayparser_unittest.cpp
+++ b/libsoftwarecontainer/unit-test/envgatewayparser_unittest.cpp
@@ -48,7 +48,7 @@ public:
 /*
  * Test that not supplying a name fails
  */
-TEST_F(EnvGatewayParserTest, TestNoName) {
+TEST_F(EnvGatewayParserTest, NoName) {
     const std::string config = "{ \"value\": \"" + value + "\" }";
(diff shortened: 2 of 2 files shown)


libsoftwarecontainer/src/gateway/environment/envgateway.cpp
```diff
@@ -26,7 +26,8 @@
 #include "envgatewayparser.h"
 
 EnvironmentGateway::EnvironmentGateway() :
-    Gateway(ID)
+    Gateway(ID),
+    m_enabled(false)
```
tobsan: This variable is never used.

libsoftwarecontainer/src/gateway/environment/envgateway.h
```diff
@@ -39,6 +39,6 @@ class EnvironmentGateway :
     bool teardownGateway();
 
 private:
-    bool m_enabled = false;
+    bool m_enabled;
```
tobsan: This variable is never used.

tobsan (changes requested):
Good change of test names, just remove the unused variable and I'm satisfied.

tobsan (approved):

libsoftwarecontainer/unit-test/envgatewayparser_unittest.cpp
```diff
@@ -99,7 +99,7 @@ TEST_F(EnvGatewayParserTest, TestValidConfModeSet) {
     ASSERT_EQ(result.second, value);
```
thenor, line 99: Is there any risk that these re-used variables (result and store) are not cleared as they should and can affect other test cases? I understand that it looks a bit cleaner to not create the variables in every test case, but this type of coding style makes me a bit nervous, since the variables are never cleaned/cleared before the test case finishes.

thenor: Internet says "no risk". :)

thenor (approved):

PR #213, opened 2016-12-08
WIP: Introduces a feature summary printout from cmake

tobsan:
Also does a bit of cmake cleanup

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -23,8 +23,10 @@ project(softwarecontainer)
 
 set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} ${CMAKE_CURRENT_SOURCE_DIR}/cmake_modules)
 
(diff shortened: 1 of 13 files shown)


CMakeLists.txt
```diff
@@ -23,8 +23,9 @@ project(softwarecontainer)
 
 set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} ${CMAKE_CURRENT_SOURCE_DIR}/cmake_modules)
 
-# load GNUInstallDirs module
+# load some cmake modules
```
breakreturn: I think this comment should explain why we use these modules, the fact that we import them is obvious.

breakreturn (changes requested):
Had a minor comment only.

DunderRoffe (approved):

tobsan:
Apparently I forgot some settings from the config file. Will update!

PR #219, opened 2016-12-12
Adds ListCapabilities on the D-BUS API

tobsan:
Both unit tests, service tests and user docs are updated as part of
this.

### DIFF
--- a/agent/src/softwarecontaineragentadaptor.cpp
+++ b/agent/src/softwarecontaineragentadaptor.cpp
@@ -73,6 +73,13 @@ bool SoftwareContainerAgentAdaptor::SetGatewayConfigs(
     return m_agent.setGatewayConfigs(containerID, configs);
 }
 
+void SoftwareContainerAgentAdaptor::ListCapabilities(std::vector<std::string> &capabilities,
+                                                     bool &success)
+{
+    capabilities = m_agent.listCapabilities();
+    success = true; // This operation can not fail
+}
+
 bool SoftwareContainerAgentAdaptor::SetCapabilities(
     const int32_t &containerID,
     const std::vector<std::string> &capabilities)
--- a/agent/src/softwarecontaineragentadaptor.cpp
+++ b/agent/src/softwarecontaineragentadaptor.cpp
@@ -11,6 +11,16 @@ SoftwareContainerAgentAdaptor::SoftwareContainerAgentAdaptor(SoftwareContainerAg
 {
 }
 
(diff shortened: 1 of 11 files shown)


agent/src/softwarecontaineragentadaptor.cpp
```diff
@@ -73,6 +73,13 @@ bool SoftwareContainerAgentAdaptor::SetGatewayConfigs(
     return m_agent.setGatewayConfigs(containerID, configs);
 }
 
+void SoftwareContainerAgentAdaptor::ListCapabilities(std::vector<std::string> &capabilities,
+                                                     bool &success)
+{
+    capabilities = m_agent.listCapabilities();
+    success = true; // This operation can not fail
```
thenor: Why do you want/need a success flag if it can't fail?

thenor (approved):

agent/src/softwarecontaineragentadaptor.cpp
tobsan: Mostly for consistency in the API. Everywhere where we have some value returned, we also return whether or not the call was successful. This is to avoid relying on magic values.

agent/unit-test/configstore_unittest.cpp
```diff
@@ -118,6 +118,20 @@ TEST_F(ConfigStoreTest, readConfigFetchOneCap) {
     ASSERT_FALSE(retGWs.empty());
 }
 
+/* Getting all capability names after reading a service manifest should
+ * result in getting all those names that are listed in the manifest file.
+ */
+TEST_F(ConfigStoreTest, readConfigFetchAllCaps) {
+    FilteredConfigStore cs = FilteredConfigStore(testDataDir + sm_path);
```
breakreturn: Please take the opportunity to change the variables that are named differently than the others (and also not according to the style guide).

breakreturn (changes requested):

agent/src/softwarecontaineragentadaptor.cpp
DunderRoffe: That does not seem like a good reason to me.
I think there should always be a purpose to a return value otherwise users might end up getting false positives.

breakreturn (approved):

agent/src/softwarecontaineragentadaptor.cpp
tobsan: I've updated this PR where the success is removed both for ListCapabilities and for List (listing of containers). Neither change the state of anything, so this should be safe.

PR #221, opened 2016-12-13
Add whitelisting policy to Pulse Gateway

kursatkobya:


breakreturn (approved):

libsoftwarecontainer/src/gateway/pulsegateway.cpp
```diff
@@ -23,20 +23,33 @@
...
 }
 
 PulseGateway::~PulseGateway()
 {
 }
 
+bool PulseGateway::getConfValue(void)
+{
+    return m_enableAudio;
+}
```
tobsan: I don't like having this method only for being able to write tests. In that case it is better to break out activation into checking the value and calling another method for doing the work. Then, a test gateway could mock off the second method and could check that that second method is being called (because it will only be called if `m_enabled` is true)

tobsan (changes requested):

libsoftwarecontainer/src/gateway/pulsegateway.cpp
kursatkobya: Too much effort for just one variable, but unfortunately you are right :)
i will do it

tobsan (approved):

PR #223, opened 2016-12-13
bump cookbook + build in parallel in Jenkinsfile

tobsan:
Since the cookbook sets -j$(nproc) now, we should give the amount of cores/ram we have to it when building in Jenkinsfile.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -16,13 +16,14 @@ node {
         buildParams       += "-DENABLE_EXAMPLES=ON -DCMAKE_INSTALL_PREFIX=/usr"
 
         // Stages are subtasks that will be shown as subsections of the finiished build in Jenkins.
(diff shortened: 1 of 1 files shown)


thenor (approved):

Jenkinsfile
```diff
@@ -23,9 +23,26 @@ node {
             sh 'git submodule update --init'
 
         stage 'StartVM'
+            // Calculate available amount of RAM
```
DunderRoffe, line 27: Please change the stage variables to use block bodies{} as the alternative is deprecated

DunderRoffe (changes requested):

Jenkinsfile
tobsan: Done.

DunderRoffe (approved):

PR #227, opened 2016-12-14
Fix behavior of creating unrelated file

kursatkobya:
fix unrelated file creation on network service tests


tobsan (changes requested):
Please add a comment on why we're exiting. As I understand it, it is to avoid doing further checks, right?

tobsan (approved):

PR #229, opened 2016-12-14
servicetest: increase dbus timeout to 5 secs

tobsan:
For some reason, the dbus tests have started to run very slowly, so for
them not to fail, we increase the timeout. The timeout is set to 5 secs
instead of 1.

### DIFF
--- a/servicetest/dbus/dbusapp.py
+++ b/servicetest/dbus/dbusapp.py
@@ -65,13 +65,21 @@ class Server(threading.Thread):
         self.service = Service(bus)
         self.__loop.run()
 
-    def wait_until_requests(self, multiplier=1, timeout=1):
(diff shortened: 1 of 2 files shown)


servicetest/dbus/dbusapp.py
```diff
@@ -65,13 +65,21 @@ def run(self):
         self.service = Service(bus)
         self.__loop.run()
 
-    def wait_until_requests(self, multiplier=1, timeout=1):
+    def wait_until_requests(self, multiplier=1, timeout=5):
```
DunderRoffe, line 68: This will break previous work with lowering the execution time. There should be a real solution to this problem.

DunderRoffe (changes requested):

tobsan:
Closed because I for some reason pushed to another branch with the actual fix.

PR #231, opened 2016-12-15
Testfw

breakreturn:


libsoftwarecontainer/src/softwarecontainer.cpp
```diff
@@ -165,45 +165,91 @@ void SoftwareContainer::addGateway(Gateway *gateway)
...
-     * The below assumption is incorrect, we do not want to appy configs to
-     * activated gateways.
-     */
-    // Go through the active gateways and check if there is a configuration for it
-    // If there is, apply it.
+    ReturnCode result = ReturnCode::SUCCESS;
 
+    result = configureGateways(gwConfig);
+    if (isError(result)) {
+        return result;
```
DunderRoffe, line 174: return the literal "ReturnCode::FAILURE" instead as it is more explicit.

libsoftwarecontainer/src/softwarecontainer.cpp
```diff
@@ -165,45 +165,91 @@ void SoftwareContainer::addGateway(Gateway *gateway)
...
+    }
+
+    result = activateGateways();
+    if (isError(result)) {
+        return result;
```
DunderRoffe, line 179: return the literal "ReturnCode::FAILURE" instead as it is more explicit.

libsoftwarecontainer/src/softwarecontainer.cpp
```diff
@@ -165,45 +165,91 @@ void SoftwareContainer::addGateway(Gateway *gateway)
...
+            log_debug() << "Configuring gateway \""
+                        << gatewayId
+                        << "\" with config: "
+                        << configStr;
             try {
-                gateway->setConfig(std::string(configStr));
+                ReturnCode configurationResult = gateway->setConfig(std::string(configStr));
+                if (isError(configurationResult)) {
+                    log_error() << "Failed to apply gateway configuration: " << configStr;
+                    return configurationResult;
```
DunderRoffe, line 203: return the literal "ReturnCode::FAILURE" instead as it is more explicit.

libsoftwarecontainer/src/softwarecontainer.cpp
```diff
@@ -165,45 +165,91 @@ void SoftwareContainer::addGateway(Gateway *gateway)
...
-                 * Because we are currently changing the error handling and behavior of
-                 * gateways, we need to ignore any exceptions at this point. This should
-                 * not be ignored when the issue is addressed.
-                 *
-                 * We get an exception here because we somethimes call setConfig() on
-                 * already activated gateways.
+                 * Exceptions in gateways during configuration are fatal errors for the whole of SC
+                 * as it means one or more capabilities are broken.
                  */
+                log_debug() << "Fatal error when configuring gateway \""
```
DunderRoffe: I think this should be an error log instead of a debug log?

libsoftwarecontainer/src/softwarecontainer.cpp
```diff
@@ -165,45 +165,91 @@ void SoftwareContainer::addGateway(Gateway *gateway)
...
+        std::string gatewayId = gateway->id();
+
+        try {
+            if (gateway->isConfigured()) {
+                ReturnCode activationResult = gateway->activate();
+                if (isError(activationResult)) {
+                    log_error() << "Failed to activate gateway \""
+                                << gatewayId
+                                << "\"";
+                    return activationResult;
```
DunderRoffe, line 235: return the literal "ReturnCode::FAILURE" instead as it is more explicit.

libsoftwarecontainer/src/softwarecontainer.cpp
```diff
@@ -165,45 +165,91 @@ void SoftwareContainer::addGateway(Gateway *gateway)
...
+                }
+            }
+        } catch (GatewayError &error) {
+            /*
+             * Exceptions in gateways during activation are fatal errors for the whole of SC
+             * as it means one or more gateways will not be active in the way one or more
+             * capabilities implies, i.e. the application environment will be in a broken state.
+             */
+            log_debug() << "Fatal error when activating gateway \""
```
DunderRoffe: error I think this should be an error log instead of a debug log?

servicetest/conftest.py
```diff
@@ -59,14 +59,35 @@ def agent(request):
...
+                    os.makedirs(default_location)
+                service_manifest_file = default_location + manifest.name()
+            else:
+                standard_location = manifest.location() + "/service-manifest.d/"
+                if not os.path.exists(standard_location):
+                    os.makedirs(standard_location)
+                service_manifest_file = standard_location + manifest.name()
+
+            with open(service_manifest_file, "w") as fh:
+                fh.write(manifest.json_as_string())
```
DunderRoffe, line 81: This assumes that the file is always created with user read permissions set. Maybe we should make sure they are set?

servicetest/environment/test_environment.py
```diff
@@ -141,6 +143,59 @@ def create_testoutput_dir(scope="module"):
...
+                                test_cap_2,
+                                test_cap_3,
+                                test_cap_4,
+                                test_cap_5,
+                                test_cap_6
+                            ])
+
+
+def service_manifests():
+    return [manifest]
```
DunderRoffe, line 203: Would be nice to define manifest in this function scope, lowering the stuff in global scope.

DunderRoffe (changes requested):

libsoftwarecontainer/src/softwarecontainer.cpp
breakreturn: I considered that at first. But then I started to use isError() to decide if the return is in fact something we should consider an error. I assume this check is available like this in order to have an abstraction between the check and the type, e.g. for extending it later and separating the logic of decision. So if the check is done with isError() I think it is more consistent with the abstraction to not at that point decide it means ReturnCode::FAILURE specifically. I don't see the benefit in isError() if we are not using it consistently. E.g. using it to enable extensibility and separation of concerns will break anyway if we interpret "error" as ReturnCode::FAILURE in the middle of the code.

If we however have some design principle about at what level of abstraction we decide to interpret "error" to ReturnCode::FAILURE specifically, then I can see a reason to do that in specific places, e.g. at some "end point" beyond which we use different abstractions.

If we should do this more explicit (which has its pros and cons as well of course), then I think the if should check for ReturnCode::FAILURE explicitly as well, and we should re-consider the purpose of isError(), isSuccess() etc.

libsoftwarecontainer/src/softwarecontainer.cpp
breakreturn: Agree, will change it.

libsoftwarecontainer/src/softwarecontainer.cpp
DunderRoffe: OK. I see your point and agree. However, since this is not how it is done in elsewhere in the code we have some cleaning up to do.

agent/src/capability/baseconfigstore.cpp
```diff
@@ -67,8 +68,9 @@ ReturnCode BaseConfigStore::readCapsFromDir(const std::string &dirPath)
     for (std::string file : files) {
         std::string filePath = dirPath + file;
 
-        if (readCapsFromFile(filePath) != ReturnCode::SUCCESS) {
+        if (isError(readCapsFromFile(filePath))) {
             log_warning() << "Could not parse a file in directory: " << filePath;
+            return ReturnCode::FAILURE;
```
kursatkobya: why don't we `return readCapsFromFile(filePath);` instead ? `readCapsFromFile` function has error logs already on failure.

DunderRoffe (approved):

agent/src/softwarecontaineragent.cpp
```diff
@@ -440,8 +442,20 @@ std::vector<std::string> SoftwareContainerAgent::listCapabilities()
 bool SoftwareContainerAgent::setCapabilities(const ContainerID &containerID,
                                              const std::vector<std::string> &capabilities)
 {
+    if (capabilities.empty()) {
+        log_warning() << "Got empty list of capabilities";
+        return true;
```
tobsan, line 447: This means that calling setCapabilities() only to get the default capabilities applied will not work. I'm fine with that though, since we're going to make sure that the default caps are applied anyway.

libsoftwarecontainer/src/gateway/gateway.cpp
```diff
@@ -69,51 +70,45 @@ void Gateway::setConfigRollback(std::string message, json_t *element)
...
+        throw GatewayError("Gateway is not configured");
     }
 
     if (!hasContainer()) {
-        log_warning() << "Activate was called on a gateway which "
-                      << "has no associated container: "
-                      << id();
-        return false;
+        log_error() << "Activate was called on a gateway which has no associated container: " << id();
+        throw GatewayError("Gateway does not have any container instance");
```
tobsan: have **a** container instance

libsoftwarecontainer/src/softwarecontainer.cpp
breakreturn: Yes, we do. Let's fix this when the pending discussion about this has resulted in some conclusion.

libsoftwarecontainer/src/softwarecontainer.cpp
```diff
@@ -165,30 +165,92 @@ void SoftwareContainer::addGateway(Gateway *gateway)
     m_gateways.push_back(std::move(std::unique_ptr<Gateway>(gateway)));
 }
 
-void SoftwareContainer::setGatewayConfigs(const GatewayConfiguration &gwConfig)
+ReturnCode SoftwareContainer::setGatewayConfigs(const GatewayConfiguration &gwConfig)
```
tobsan, line 168: We should really come up with a better naming. Perhaps more of a style question, but having `setGatewayConfigs` call `configureGateways` and then `activateGateways` feels very non-intuitive. Maybe something like `setupGateways` or `startGateways` would be more fitting.

libsoftwarecontainer/src/softwarecontainer.cpp
```diff
@@ -165,30 +165,92 @@ void SoftwareContainer::addGateway(Gateway *gateway)
...
+}
+
+ReturnCode SoftwareContainer::configureGateways(const GatewayConfiguration &gwConfig)
+{
     for (auto &gateway : m_gateways) {
         std::string gatewayId = gateway->id();
 
         json_t *config = gwConfig.config(gatewayId);
         if (config != nullptr) {
             char *configStr = json_dumps(config,0);
```
tobsan: I'm guessing we don't ever want to use this as a `char *`, so you could implicitly convert it to a string by just using std::string as the type. It will be freed when the string goes out of scope anyway.

Then we'd get rid of the conversion below.

servicetest/testframework/lib.py
```diff
@@ -35,6 +35,113 @@
...
+            caps        A list of Capability objects.
+        """
+        self.__location = location
+        self.__name = name
+        self.__caps = caps
+        self.__manifest_content = dict()
+
+        self.__create_manifest_content()
+
+    def __create_manifest_content(self):
```
tobsan, line 100: Some comments in this file would be nice.

tobsan (changes requested):

tobsan (approved):

PR #238, opened 2016-12-16
Fixes spawn regression by only using one core

tobsan:
For some reason, when running with multiple cores and virtualbox,
performance degrades when stress testing by spawning many processes. An
article (http://www.mihaimatei.com/virtualbox-performance-issues-multiple-cpu-cores/)
suggests this is because processing with multicore in VBox requires the
physical cores it uses to be free (and locking more cpus take more
time).

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -36,18 +36,9 @@ node {
                 println "Will set VAGRANT_RAM to ${gigsram}"
             }
 
-            // And calculate number of CPUS available
-            String numcpus = sh (
-                script: 'nproc',
-                returnStdout: true
-            )
-
-            println "Will set VAGRANT_CPUS to ${numcpus}"
-
             // Start the machine (destroy it if present) and provision it
             sh "cd ${workspace} && vagrant destroy -f || true"
             withEnv(["VAGRANT_RAM=${gigsram}",
-                     "VAGRANT_CPUS=${numcpus}",
                      "APT_CACHE_SERVER=10.8.36.16"]) {
                 sh "cd ${workspace} && vagrant up"
             }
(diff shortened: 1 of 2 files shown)


Jenkinsfile
```diff
@@ -36,18 +36,9 @@ node {
...
-                script: 'nproc',
-                returnStdout: true
-            )
-
-            println "Will set VAGRANT_CPUS to ${numcpus}"
-
             // Start the machine (destroy it if present) and provision it
             sh "cd ${workspace} && vagrant destroy -f || true"
             withEnv(["VAGRANT_RAM=${gigsram}",
-                     "VAGRANT_CPUS=${numcpus}",
```
frznlogic, line 50: I would prefer that we dont remove this line, it might be nice to use sometimes.

servicetest/dbus/test_dbus.py
```diff
@@ -98,7 +98,6 @@ def test_query_in(self):
             finally:
                 ca.terminate()
 
-    @pytest.mark.skip(reason="See reported issue about this")
```
frznlogic, line 101: Ok, removed comment, but what report? Would be nice with a reference in the future.

frznlogic (changes requested):

servicetest/dbus/test_dbus.py
tobsan: The issue is in the internal bug tracker though, so it makes little sense to reference it here imho.

Jenkinsfile
tobsan: Then would you prefer it being set to 1?

frznlogic: Sorry about that, I failed to realize this is the jenkinsfile :D

servicetest/dbus/test_dbus.py
frznlogic: In that case I think we need somewhere better to reference, or add a better comment about what's actually wrong. Anyways, it was removed so no point debating it I guess ;)

frznlogic (approved):

PR #240, opened 2016-12-19
Fix file clean up problem

kursatkobya:


common/filetoolkitwithundo.cpp
```diff
@@ -226,19 +226,32 @@ ReturnCode FileToolkitWithUndo::createSharedMountPoint(const std::string &path)
...
     }
 
-    m_cleanupHandlers.push(new MountCleanUpHandler(path));
+    m_cleanupHandlers.push_back(new MountCleanUpHandler(path));
     log_debug() << "Created shared mount point at " << path;
 
     return ReturnCode::SUCCESS;
 }
 
+bool FileToolkitWithUndo::isPathExist(const std::string path)
```
tobsan: The name of this function isn't very obvious. hasPathAlready or something would make more sense.

common/filetoolkitwithundo.cpp
```diff
@@ -226,19 +226,32 @@ ReturnCode FileToolkitWithUndo::createSharedMountPoint(const std::string &path)
...
+
 ReturnCode FileToolkitWithUndo::writeToFile(const std::string &path, const std::string &content)
 {
     auto ret = softwarecontainer::writeToFile(path, content);
     if (isError(ret)) {
         return ret;
     }
-    m_cleanupHandlers.push(new FileCleanUpHandler(path));
+
+    if (!isPathExist(path)) {
```
tobsan: Shouldn't this type of check be done for any cleanup handler? I mean it's good that it is being done here, but shouldn't be done for mounts etc also, to make it more generic?

Perhaps this is a bit over-the-top though, since there is no common way to compare cleanup handlers :/

common/filetoolkitwithundo.h
```diff
@@ -112,12 +112,19 @@ class FileToolkitWithUndo
...
+     */
+    bool isPathExist(const std::string path);
 protected:
     /**
-     * @brief m_cleanupHandlers A stack of cleanupHandlers added during the lifetime of the
+     * @brief m_cleanupHandlers A vector of cleanupHandlers added during the lifetime of the
      *  FileToolKitWithUndo that will be run from the destructor.
      */
-    std::stack<CleanUpHandler *> m_cleanupHandlers;
+    std::vector<CleanUpHandler *> m_cleanupHandlers;
```
tobsan, line 125: Why is this changed from a stack to a vector? The desired behavior is LIFO as a stack, so why not stick with it?

tobsan (changes requested):

common/filetoolkitwithundo.h
kursatkobya: searching en element in stack does not meaningful since it does not support random access

common/filetoolkitwithundo.cpp
kursatkobya: make sense, will do

common/overlaysynccleanuphandler.cpp
```diff
@@ -32,3 +32,8 @@ ReturnCode OverlaySyncCleanupHandler::clean()
 {
     return RecursiveCopy::getInstance().copy(m_src, m_dst);
 }
+
+const std::string OverlaySyncCleanupHandler::queryName()
+{
+    return "";
```
tobsan, line 38: This does not make sense. I would say that the destination path would be what we want to check here? The reason for that would be that it is valid to mount the same location to many places (in theory), but not to mount several locations to the same place.

tobsan (changes requested):
Just one comment on the OverlaySyncCleanupHandler

common/overlaysynccleanuphandler.h
```diff
@@ -43,6 +43,15 @@ class OverlaySyncCleanupHandler : public CleanUpHandler
      * @return ReturnCode::FAILURE on failure
      */
     ReturnCode clean() override;
+
+    /**
+     * @brief this function is needed to query member name yet it is not needed for
+     * OverlaySyncCleanupHandler. Since OverlaySyncCleanupHandler uses DirectoryCleanUpHandler
+     * for removing directory, this function won't be called ever.
```
tobsan: 1. Only use one short sentence for @brief, write the longer description in a separate paragraph.
2. This function **is** called when we iterate over the cleanuphandlers, but the reasoning is good anyway.

tobsan (changes requested):

common/overlaysynccleanuphandler.h
```diff
@@ -43,6 +43,17 @@ class OverlaySyncCleanupHandler : public CleanUpHandler
...
+    /**
+     * @brief queryName is needed to query member name yet its full functionality is not needed for
+     * this class.
+     *
+     * Since OverlaySyncCleanupHandler uses DirectoryCleanUpHandler for removing directory, this
+     * function won't be called for the same reason.
+     *
+     * @return an empty string
+     */
+    const std::string queryName() override;
```
tobsan, line 57: Sorry, but I still don't like this description. OverlaySyncCleanupHandler doesn't use DirectoryCleanupHandler, it uses RecursiveCopy.

The reason why we're returning an empty string here is because we don't want any paths to compare equal to an overlay path, isn't it? Then we should write that.

kursatkobya: Actually we are returning empty string because no directory removal operation occurs within OverlaySyncCleanupHandler, so it is irrelevant when the code checks whether the directory is in the list or not since we are doing this for cleaning path. Maybe i should add this as a comment.

tobsan: Sure, but this check is done for anything we want to add a cleanup handler for, so it is a little more generic than just directories.

DunderRoffe (changes requested):
Sorry to be picky but there should really be a description attached to the commit message. This is in order to make it easier when searching through the commits at a later stage.

DunderRoffe (approved):

tobsan (approved):

PR #241, opened 2016-12-19
agent: set default caps if gateways not configured

tobsan:
The only way to have the gateways configured from the agent is through
setCapabilities, so if they are not configured, then that call has not
been made, and the default caps have not been applied.

### DIFF
--- a/agent/src/softwarecontaineragent.cpp
+++ b/agent/src/softwarecontaineragent.cpp
@@ -297,6 +297,24 @@ bool SoftwareContainerAgent::execute(ContainerID containerID,
         return false;
     }
 
+    /*
+     * We want to always apply any default capabilities we have. If the container is in READY state,
+     * that means that its gateways have been configured. The only way to configure the gateways
+     * from the agent is through setCapabilities - which sets the default caps also.
+     *
+     * If it has not been set, then we use a call without arguments to setCapabilities to set it up,
+     * since we then should get only the default ones.
+     */
+    if (container->getContainerState() != ContainerState::READY) {
+        log_info() << "Container not configured yet, configuring with default capabilities, if any";
+        GatewayConfiguration gatewayConfigs = m_defaultConfigStore->configs();
+        if (!updateGatewayConfigs(containerID, gatewayConfigs)) {
+            log_error() << "Could not set default capabilities on container " << containerID;
+            pid = INVALID_PID;
+            return false;
+        }
+    }
+
     // Set up a CommandJob for this run in the container
     auto job = new CommandJob(*container, cmdLine);
     job->setOutputFile(outputFile);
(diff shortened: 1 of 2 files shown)


agent/src/softwarecontaineragent.cpp
```diff
@@ -297,6 +297,22 @@ bool SoftwareContainerAgent::execute(ContainerID containerID,
         return false;
     }
 
+    // We want to always apply any default capabilities we have. If the container is in READY state,
+    // that means that its gateways have been configured. The only way to configure the gateways
+    // from the agent is through setCapabilities - which sets the default caps also.
+    //
+    // If it has not been set, then we use a call without arguments to setCapabilities to set it up,
+    // since we then should get only the default ones.
```
breakreturn: For multi line comments it would be more consistent with the rest of the code to use:
/*
 *
 */

breakreturn (changes requested):

agent/src/softwarecontaineragent.cpp
DunderRoffe: I think it is nice to not use that when writing something that should not end up in the API docs as it looks different from the doxygen comments. On the other hand, this might be useful in the API if tweaked a bit?

breakreturn: Do you mean you see a risk that it might be interpreted by someone as a doxygen comment and lead to confusion? AFAIK the most consistent choice now would be to use /* */ based on the rest of the code.

tobsan: FYI this will not be interpreted by doxygen.

servicetest/capabilities/test_caps.py
```diff
@@ -80,13 +83,34 @@ def logfile_path():
...
+ENV_GW_CONFIG = [{
+    "name": "TEST_ENV_VAR",
+    "value": "TEST_ENV_VALUE"
+}]
+
+""" The default capability, just the env gateway config """
+default_cap = Capability("test.cap.default.sample",
+                        [
+                            { "id": "env", "config": ENV_GW_CONFIG }
+                        ])
```
breakreturn: The list is missing one space indentation.

servicetest/capabilities/test_caps.py
```diff
@@ -186,5 +206,32 @@ def test_list_capabilities(self):
...
+        sc = Container()
+        try:
+            sc.start(DATA)
+
+            # Start without setting any capabilities
+            sc.launch_command("python " +
+                                  sc.get_bind_dir() +
+                                  "/testhelper.py --test-dir " +
+                                  sc.get_bind_dir() + "/testoutput" +
+                                  " --do-get-env-vars")
```
breakreturn: Indentation looks off

servicetest/capabilities/test_caps.py
```diff
@@ -186,5 +206,32 @@ def test_list_capabilities(self):
             caps = sc.list_capabilities()
             # All caps present in the service manifest(s) set up previously should be listed
             assert "test.cap" in caps and "test.cap.valid-dbus" in caps and "test.cap.broken-gw-config" in caps
+            # The default capabilities should not be visible in the list
+            assert "test.cap.default.sample" not in caps
+        finally:
+            sc.terminate()
+
+    def test_no_caps_gives_default_only(self):
```
breakreturn: This name is off, it should be something like "test_defaults_are_applied" or something, as it doesn't test any interaction between set caps and default caps, the test result would be the same even if there was some cap set.

breakreturn (approved):

PR #242, opened 2016-12-19
Make DBusGateway bus specific

DunderRoffe:
Rework of DBusGateway and DBusGatewayParser making them agnostic to
configuration for types they are not configured for.


thenor (approved):

libsoftwarecontainer/src/gateway/dbus/dbusgateway.cpp
```diff
@@ -33,8 +33,19 @@ DBusGateway::DBusGateway(ProxyType type
              + name
              + ".sock";
 
-    m_sessionBusConfig = json_array();
-    m_systemBusConfig = json_array();
+    // Write configuration
+    m_totConfig = json_object();
+    m_busConfig = json_array();
+
+    typeStr = m_type == SessionProxy ? SESSION_CONFIG : SYSTEM_CONFIG;
```
kursatkobya: kinda hard to read can you at least add parentheses for sides of equiality operator

libsoftwarecontainer/src/gateway/dbus/dbusgateway.cpp
```diff
@@ -44,12 +55,12 @@ DBusGateway::~DBusGateway()
     }
 }
 
-
 ReturnCode DBusGateway::readConfigElement(const json_t *element)
 {
     DBusGatewayParser parser;
 
-    if (isError(parser.parseDBusConfigElement(element, m_sessionBusConfig, m_systemBusConfig))) {
+    // TODO: This should really be done with exceptions instead and json_t* as return type.
```
breakreturn: There's an exception class in gateway.h used by the base class nowadays. Perhaps extend that and use here as you'd like. Or leave it to a separate commit.

libsoftwarecontainer/src/gateway/dbus/dbusgateway.cpp
```diff
@@ -33,8 +33,19 @@ DBusGateway::DBusGateway(ProxyType type
...
-    m_systemBusConfig = json_array();
+    // Write configuration
+    m_totConfig = json_object();
+    m_busConfig = json_array();
+
+    typeStr = m_type == SessionProxy ? SESSION_CONFIG : SYSTEM_CONFIG;
+    json_object_set(m_totConfig, typeStr, m_busConfig);
+
+    // Send an empty array as config for the irrelevant proxy type
+    const char* unusedTypeStr = m_type == SystemProxy ? SESSION_CONFIG : SYSTEM_CONFIG;
```
kursatkobya: also here

libsoftwarecontainer/src/gateway/dbus/dbusgateway.h
```diff
@@ -82,6 +85,12 @@ class DBusGateway :
     virtual bool teardownGateway() override;
 
 private:
+
+    // JSON stuff. These are members to allow continous appends to the config
+    json_t *m_totConfig;
```
kursatkobya: can you find better name for this variable i don't think its explicit enough. Suggestion : entireConfig

libsoftwarecontainer/src/gateway/dbus/dbusgatewayparser.cpp
```diff
@@ -19,45 +19,25 @@
...
-
-    if (!JSONParser::hasKey(element, SESSION_CONFIG)
-     && !JSONParser::hasKey(element, SYSTEM_CONFIG)) {
-        log_error() << "Neither system nor session configuration was provided";
-        return ReturnCode::FAILURE;
-    }
-
-    if (isError(parseDBusBusConfig(element, SESSION_CONFIG, session))) {
+    json_t *configExists = json_object_get(element, key);
+    if (configExists == nullptr) {
```
kursatkobya: its not that important but, (nullptr == configExists) would be better her imo.

kursatkobya (changes requested):
Generally looks good. I've got just minor comments.

libsoftwarecontainer/src/gateway/dbus/dbusgateway.cpp
DunderRoffe: I was thinking of doing that but hesitated in anticipation of the exception design discussion. But sure I can fix this if you like.

libsoftwarecontainer/src/gateway/dbus/dbusgateway.cpp
DunderRoffe: I'll do something about it

libsoftwarecontainer/src/gateway/dbus/dbusgatewayparser.cpp
tobsan: The name `configExists` is not very nice either, use `config` or something similar instead.

libsoftwarecontainer/src/gateway/dbus/dbusgateway.cpp
```diff
@@ -76,29 +87,26 @@ bool DBusGateway::activateGateway()
...
     }
 
-    // Write configuration
-    json_t *jsonConfig = json_object();
-    json_object_set(jsonConfig, DBusGatewayParser::SESSION_CONFIG, m_sessionBusConfig);
-    json_object_set(jsonConfig, DBusGatewayParser::SYSTEM_CONFIG, m_systemBusConfig);
-
-    char *config_c = json_dumps(jsonConfig, JSON_COMPACT);
+    // Dump to string and clean up
+    char *config_c = json_dumps(m_totConfig, JSON_COMPACT);
```
tobsan: You can use implicit conversion to std::string by declaring `std::string config` here directly, then you won't need to `to_string` it.

libsoftwarecontainer/src/gateway/dbus/dbusgateway.h
```diff
@@ -82,6 +85,12 @@ class DBusGateway :
     virtual bool teardownGateway() override;
 
 private:
+
+    // JSON stuff. These are members to allow continous appends to the config
```
tobsan: continuous*

tobsan (changes requested):
See inline comments, but I have two additional comments:

1) With this code, the proxy will still be started as long as one provides an array for the respective gateway (system or session).

2) This will still generate errors when one of the gateways are not configured properly, even if one only wants to configure, say, the session bus.

libsoftwarecontainer/src/gateway/dbus/dbusgateway.cpp
breakreturn: Do as you wish, it might be good to wait for that discussion as well if you feel like that.

libsoftwarecontainer/src/gateway/dbus/dbusgateway.cpp
```diff
@@ -18,191 +17,100 @@
...
+     } catch (GatewayError e) {
+         systemBusActivationResult = ReturnCode::FAILURE;
+     }
+
+     if (isError(sessionBusActivationResult) && isError(systemBusActivationResult)) {
+         std::string err = "Neither dbus session bus nor dbus system bus could be activated";
+         log_error() << err;
+         throw GatewayError(err);
+     }
+     return ReturnCode::SUCCESS;
```
tobsan, line 73: This is a bit backwards imho. Can't you instead check if the specific bus object was configured and then expect activation to work?

libsoftwarecontainer/src/gateway/dbus/dbusgateway.cpp
```diff
@@ -18,191 +17,100 @@
...
-    }
+     log_debug() << "Setting config for dbus system bus";
+     ReturnCode systemBusParseResult = systemBus.setConfig(config);
 
-    return ReturnCode::SUCCESS;
-}
+     if (isError(sessionBusParseResult) && isError(systemBusParseResult)) {
+         log_error() << "Neither session nor  system bus could use the given config";
+         return ReturnCode::FAILURE;
+     }
```
tobsan: I actually would rather see that instead of checking if we had two errors, did a bit of inspection on the config and expect config to work for the given keys. I realize that this would break the design of this container-style class, so it is perhaps not necessary.

libsoftwarecontainer/src/gateway/dbus/dbusgateway.cpp
```diff
@@ -18,191 +17,100 @@
...
+     } catch (GatewayError e) {
+         systemBusTeardownResult = ReturnCode::FAILURE;
+     }
+
+     if (isError(sessionBusTeardownResult) && isError(systemBusTeardownResult)) {
+         std::string err = "Neither dbus session bus nor dbus system bus could be activated";
+         log_error() << err;
+         throw GatewayError(err);
+     }
+     return ReturnCode::SUCCESS;
```
tobsan, line 99: And the same goes here. Perhaps we can check the gateway state here also before calling teardown so that we know if it is actually needed? That would still work with the container pattern since all gateways (or sub-objects) implements that method.

libsoftwarecontainer/src/gateway/dbus/dbusgateway.h
```diff
@@ -21,98 +20,50 @@
...
      */
-    virtual bool teardownGateway() override;
-
-private:
-    std::string socketName();
-    bool isSocketCreated() const;
-
-    // Socket used for exposing D-Bus in container
-    std::string m_socket;
+    virtual ReturnCode readConfigElement(const json_t *) override { return ReturnCode::SUCCESS; }
```
tobsan, line 48: Since you redefine readConfigElement in the cpp file anyway, why have the default implementation like this?

libsoftwarecontainer/src/gateway/dbus/dbusgatewayinstance.cpp
```diff
@@ -0,0 +1,225 @@
...
+    json_object_set(m_entireConfig, unusedTypeStr, emptyArray);
+
+
+}
+
+DBusGatewayInstance::~DBusGatewayInstance()
+{
+    if (isActivated()) {
+        teardown();
+    }
```
tobsan, line 62: I would prefer that members are decref:ed / deleted here, rather than in teardown()

tobsan (changes requested):

libsoftwarecontainer/src/gateway/dbus/dbusgateway.cpp
DunderRoffe: No since not being configured at activation will result in an exception being thrown.

libsoftwarecontainer/src/gateway/dbus/dbusgateway.h
DunderRoffe: I do not define it in the cpp file

libsoftwarecontainer/src/gateway/dbus/dbusgateway.cpp
DunderRoffe: Ill look in to it

libsoftwarecontainer/src/gateway/dbus/dbusgatewayinstance.cpp
DunderRoffe: That is a bad idea since the config will be saved or memory will be leaked when tearing down without deleting the Gateway.

libsoftwarecontainer/src/gateway/dbus/dbusgateway.cpp
DunderRoffe: This would be nice and could be done rather nice without breaking the design. However, it would require the setConfig() to take json_t* as a parameter to not force reimplementation / duplication of the string -> json_t* code.

libsoftwarecontainer/src/gateway/dbus/dbusgateway.h
tobsan: My bad, I saw the instance class, sorry.

libsoftwarecontainer/src/gateway/dbus/dbusgateway.cpp
```diff
@@ -18,191 +17,99 @@
...
-    } catch (const Glib::Error &ex) {
-        log_error() << "Failed to launch dbus-proxy";
-        return false;
-    }
-
-    log_debug() << "Started dbus-proxy: " << m_pid;
-
-    return true;
+     ReturnCode sessionBusTeardownResult;
+     if (sessionBus.isConfigured()) {
```
tobsan: I think you meant to use `isActivated()` instead.

tobsan (approved):

libsoftwarecontainer/src/gateway/dbus/dbusgatewayinstance.cpp
```diff
@@ -0,0 +1,225 @@
...
+
+
+#include "dbusgateway.h"
+#include "dbusgatewayparser.h"
+
+DBusGatewayInstance::DBusGatewayInstance(ProxyType type
+                       , const std::string &gatewayDir
+                       , const std::string &name)
+    : Gateway(ID)
+    , m_type(type)
```
kursatkobya, line 29: could you please correct initializer list format. I guess it is aggregriated as 
```
(...  , const std::string &name) : 
    Gateway(ID),
    m_type(type)
```

libsoftwarecontainer/src/gateway/dbus/dbusgatewayinstance.cpp
```diff
@@ -0,0 +1,225 @@
...
+}
+
+bool DBusGatewayInstance::testDBusConnection(const std::string &config)
+{
+    log_debug() << "Sending config " << config;
+    unsigned int count = sizeof(char) * config.length();
+    ssize_t written = write(m_infp, config.c_str(), count);
+
+    // writing didn't work at all
+    if (written == -1) {
```
kursatkobya, line 129: `-1 == written` would be better

libsoftwarecontainer/src/gateway/dbus/dbusgatewayinstance.cpp
```diff
@@ -0,0 +1,225 @@
...
+        Glib::spawn_async_with_pipes( "."
+                                    , commandVec
+                                    , envVec
+                                    , Glib::SPAWN_STDOUT_TO_DEV_NULL // Redirect stdout
+                                        | Glib::SPAWN_STDERR_TO_DEV_NULL // Redirect stderr
+                                        | Glib::SPAWN_SEARCH_PATH // Search $PATH
+                                        | Glib::SPAWN_DO_NOT_REAP_CHILD // Lets us do waitpid
+                                    , sigc::slot<void>() // child setup
+                                    , &m_pid
+                                    , &m_infp // stdin
```
breakreturn: We don't put commas like this anymore, they should be at the end of the lines.

kursatkobya (approved):
small comments, but looks good

breakreturn:
This looks good to me given that it has been reviewed while developing it. So apart from the discussion about making sure we catch errors when one of the "instances" can't be configured/activated properly, this is OK to merge for me.

PR #244, opened 2016-12-20
Debugging documentation

frznlogic:


tobsan (commented):
Looks good so far!

doc/chapters/debugging/index.rst
```diff
@@ -0,0 +1,197 @@
...
+so forth.
+
+Cleanup
+-------
+
+The internal filesystem will always be deleted and recreated when an app is
+restarted. This means that anything you want to save during debugging needs to
+be saved to a bind mounted directory before you shut down the container.
+
+`SoftwareContainer` will also delete all container instances upon startup,
```
tobsan: Maybe we should mention that this is done by the agent.

doc/chapters/debugging/index.rst
```diff
@@ -0,0 +1,197 @@
...
+*lxc-info* provides information about the running container, such as network
+interface used for t he bridge, the state of the container, what PID it has and
+so forth.
+
+Cleanup
+-------
+
+The internal filesystem will always be deleted and recreated when an app is
+restarted. This means that anything you want to save during debugging needs to
+be saved to a bind mounted directory before you shut down the container.
```
tobsan: Perhaps a bit misleading wording. Restarting an app is not necessarily the same as destroying the container and starting a new one, although that is probably what will happen most of the time.

doc/chapters/debugging/index.rst
```diff
@@ -0,0 +1,197 @@
...
+
+Running GDB inside container
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+If you run `gdb` from inside the container using `lxc-attach`, you should not
+need to change any paths etc to find the correct libraries etc. You will need
+to have the `PID` from inside the container available to connect to however. this
+can be found out by running::
+
+    sudo lxc-attach --name SC-X ps aux
```
tobsan: Why the sudo? You had a section above that one needs to be the right user anyway.

doc/chapters/debugging/index.rst
```diff
@@ -0,0 +1,197 @@
...
+Then running::
+
+    sudo lxc-attach --name SC-X -- gdb --pid YY
+
+Replace YY with the `PID` as seen inside the container. You can also run the
+application directly inside the container and start it with gdb directly::
+
+    sudo lxc-attach --name SC-X -- gdb --args /bin/ls
+
+TODO: Uncertain why, but this fails with unable to find executable.
```
tobsan, line 78: Could it be that in our containers ls is a shell builtin and not a binary? Not sure if that is the case.

doc/chapters/debugging/index.rst
```diff
@@ -0,0 +1,197 @@
...
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+If you run `gdb` from outside the container you will run into some problems with the
+paths as the application is executing inside a `chroot` and `gdb` is executing
+outside of it, this is fairly easily fixed however. First off you need to
+provide `gdb` with the path of the executable as the path provided by the system
+is inside the chroot::
+
+    # ps aux |grep main
+    # sudo gdb --pid XXX /tmp/container//SC-0/gateways/app/main
```
tobsan, line 90: This assumes that the app will always be in a directory that is reachable from the host. That is probably a good idea to suggest though, but it is not a requirement when running an app.

doc/chapters/debugging/index.rst
```diff
@@ -0,0 +1,197 @@
...
+available as normal without any changes. If you have dynamic libraries inside
+your container that the application are linking to, you may need to add the
+proper `solib-search-paths` to the libraries.
+
+Debugging using QtCreator inside a container
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Debugging applications running inside a container can be done, however, all
+processes are run as root so you need to start `QtCreator` as root to be
+able to connect to the process inside the container.
```
tobsan, line 107: Is there no way around this at all? That sucks a bit :/.

doc/chapters/debugging/index.rst
```diff
@@ -0,0 +1,197 @@
...
+7. Open *Debug -> Start Debugging -> Attach to Running Application...*
+8. Find the right `PID` in the list, and attach.
+9. The process will be halted and you can debug it from `QtCreator`.
+
+
+Core dumps
+----------
+
+Nothing special should be needed to debug a core file using for example
+`QtCreator` or `GDB`. Load the file as normal.
```
tobsan, line 125: When we've done work on generating core dumps, I will update this section I guess.

doc/chapters/debugging/index.rst
```diff
@@ -0,0 +1,197 @@
...
+
+Stack traces
+------------
+
+TODO: Do we actually receive backtraces from something? Not sure what this means
```
tobsan: Aren't those part of a core dump?

doc/chapters/debugging/index.rst
```diff
@@ -0,0 +1,197 @@
...
+exception that it runs as root, and hence all debugging tools also need to be
+run as root until such a time that `SoftwareContainerAgent` can be run as a
+non-root user.
+
+Logging
+-------
+
+All the logs are performed using `ivi-logging`, meaning that `ivi-logging`
+facilities can be used to filter logs. By default the logging is set to `DEBUG`
+as of this writing.
```
tobsan: We could add that one can use ivi-logging with DLT as backend.

doc/chapters/debugging/index.rst
```diff
@@ -0,0 +1,197 @@
...
+
+Core dumps
+----------
+
+Coredumps can be debugged normally, no special usage necessary.
+
+Stack traces
+------------
+
+TODO: Not sure what to write here?
```
tobsan: Part of core dumps, right?

tobsan (changes requested):
Have some inline comments, but something that I think should be added is the case where someone uses SC on an embedded system and works on an app on their workstation. A setup with gdb-server would be nice to be able to show. Perhaps it is super simple and it would just suffice to state "it is also possible to use gdb-server" or something.

doc/chapters/debugging/index.rst
frznlogic: The base filesystem is deleted AFAICR? Anyways, added a note that most of the filesystems should be bind mounted anyways.

doc/chapters/debugging/index.rst
frznlogic: Added.

doc/chapters/debugging/index.rst
frznlogic: Removed, keeping it in the gdb example though as I think it's not quite the same type of example.

doc/chapters/debugging/index.rst
frznlogic: I don't think so. I had the same issue with other applications as well... and gdb does not read shell builtins in that way anyways afaik.

doc/chapters/debugging/index.rst
frznlogic: Not sure that I understand this? you mean if the filesystem is bind mounted from somewhere on a network mount or...?

doc/chapters/debugging/index.rst
frznlogic: I dont think so, but I haven't looked into ways around it really. It's part of the security model of Linux imho, and being able to debug a root app from a non-root app would be strongly discouraged imho.

doc/chapters/debugging/index.rst
frznlogic: I did generate a core dump and debugged it... basically, it works as you would expect, nothing special needed :)

doc/chapters/debugging/index.rst
frznlogic: Ok :) So basically nothing extra required then. Removing it.

doc/chapters/debugging/index.rst
frznlogic: True =) I've added some blurb about it at least.

doc/chapters/debugging/index.rst
frznlogic: Removing it in that case.

doc/chapters/debugging/index.rst
tobsan: Yes, the rootfs is deleted IF we delete the container. My point was that deleting the container is not always necessary in order to restart an app (but one will probably do that anyway)

doc/chapters/debugging/index.rst
tobsan: I mean that one can bind-mount to any location in the container, in which case we first mount into /gateways, and then move the mount (to some place which the host can't easily access). Will the gdb chroot trick still work then?

doc/chapters/debugging/index.rst
frznlogic: Ah, I see your point. Rewriting it now to indicate container being destroyed/recreated.

doc/chapters/debugging/index.rst
frznlogic: It depends, if you do really funky shit with the chroot, it wont work... IMHO, I'm not sure it's possible for us to take every single variation into account though... I guess it's up to the user to solve those problems in that case... no?

tobsan: Yeah, I guess so.

tobsan (approved):
Looks good, but would be nice to get rid of those TODOs if possible.

PR #247, opened 2016-12-21
Jenkinsfile: avoid unreachable paths with coverage

tobsan:
We sometimes end up having (unreachable) paths when running in Jenkins, resulting in some unit test results being reported twice.


tobsan:
Alright, this seems to work out fine in Jenkins, feel free to merge.

breakreturn (approved):

PR #248, opened 2016-12-21
API and getting-started documentation

frznlogic:
Updated naming convention
Alphabetically sorted list of API's

### DIFF
--- a/doc/chapters/api/index.rst
+++ b/doc/chapters/api/index.rst
@@ -3,38 +3,41 @@
 API
 ***
 
-This chapter documents the D-Bus API for interacting with the SoftwareContainerAgent. This is
-mostly useful for integrators.
+This chapter documents the D-Bus API for interacting with the
+SoftwareContainerAgent. This is mostly useful for integrators.
 
 .. _dbus-api:
 
 D-Bus API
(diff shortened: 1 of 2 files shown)


DunderRoffe (approved):

doc/chapters/api/index.rst
```diff
@@ -3,38 +3,41 @@
...
+        :readOnly: ``bool`` indicates whether the directory is read-only or not.
 
-ListCapabilities
-----------------
-Lists all capabilities that the user can apply.
+|
 
-:Return Value:
-        :capabilities: ``array<string>`` all available capability names
+:Return Values:
```
thenor: This is only one value. It would be good to have consistency, either by always having plural or varying depending on if it is one or more.

frznlogic: I have not changed this text from before fyi, just moved it around, but I will fix it;).

thenor (approved):

doc/chapters/api/index.rst
thenor: I expected that that was the case. Thought it would be good to change it at this time. :)

PR #251, opened 2016-12-21
If the config file can't be found don't abort

thenor:
Softwarecontainer Agent should not abort if the config
file can't be found.


thenor:
I need to write some new tests for this, if there aren't any already. I pushed for review of the code only. Works when testing manually.

tobsan (approved):
Looks good. Could probably be easily tested by using the "-c" option to the agent, and use that through the service tests.

breakreturn (approved):

breakreturn:
Looks good.

For this type of change I'm OK with not testing it specifically, unless there is some smooth way I can't see at the moment. At some point we might want to test all sorts of permutations of configs and command line options anyway and I guess this would be covered by that. From my perspective, this is OK for a merge as it is.

kursatkobya (approved):

PR #253, opened 2016-12-22
Added support for mandatory configs

breakreturn:
It's now possible to specify that some configs
are mandatory. This is currently not used by the agent
but will be needed when we move to having certain configs
made mandatory due to how SC is configured and built, e.g.
for networking.

### DIFF
--- a/agent/src/config/config.cpp
+++ b/agent/src/config/config.cpp
@@ -21,6 +21,8 @@
 #include "config.h"
 
 
+namespace softwarecontainer {
+
 // Illegal values used as initial command line option values to check if user
 // set them or not
 const std::string Config::SC_CONFIG_PATH_INITIAL_VALUE = "";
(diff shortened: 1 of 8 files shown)


agent/src/config/config.cpp
```diff
@@ -45,24 +45,87 @@ const std::string Config::SERVICE_MANIFEST_DIR_KEY = "service-manifest-dir";
...
+    m_defaults(std::move(defaults)),
+    m_stringOptions(stringOptions),
+    m_intOptions(intOptions),
+    m_boolOptions(boolOptions)
+{
+    loadConfig();
+
+    /* Checking for mandatory config group is only relevant for the main config source
+     */
+    for (std::string group: mandatory.groups()) {
```
tobsan: space before ":"

agent/src/config/config.cpp
```diff
@@ -45,24 +45,87 @@ const std::string Config::SERVICE_MANIFEST_DIR_KEY = "service-manifest-dir";
...
+                case ConfigType::Integer:
+                    getIntegerValue(group, key);
+                    break;
+                case ConfigType::Boolean:
+                    getBooleanValue(group, key);
+                    break;
+                default:
+                    break;
+            }
+        } catch (softwarecontainer::ConfigError &error) {
```
tobsan: Are you not already in the softwarecontainer namespace? If not, you should be.

agent/src/config/mandatoryconfigs.cpp
```diff
@@ -0,0 +1,46 @@
...
+MandatoryConfigs::MandatoryConfigs():
+    m_configs(std::vector<std::tuple<std::string, std::string, ConfigType>>()),
+    m_groups(std::vector<std::string>())
+{
+}
+
+void MandatoryConfigs::addConfig(const std::string &group, const std::string &key, ConfigType type)
+{
+    log_debug() << "Adding mandatory config \"" << key << "\" with group \"" << group << "\"";
+    m_groups.push_back(group);
```
tobsan, line 33: This means that a mandatory key makes the group it is in mandatory as well, right? That should probably be documented at some stage.

tobsan (changes requested):

tobsan (approved):

PR #254, opened 2016-12-22
Split tests into unit tests + component tests

tobsan:
Some of the tests we used to call unit tests were actually component
tests. This splits them apart into different targets and it also makes
it easier to just run the actual unit tests, since they require no
special system setup


scripts/run-unit-tests.py
```diff
@@ -0,0 +1,64 @@
...
+    exitCode = os.system(command)
+    testSuites[name] = exitCode
+
+#
+# NOTE: Don't add spaces in the names, the names are used as parts of paths to save
+#       test results.
+#
+
+# These are proper unit tests
+runTestSuite("./agent/unit-test/configstore-unit-test", "ConfigStoreTest")
```
breakreturn: Since we will have one gtest binary per sub-component like this, the "ConfigStoreTest" should really be something like "AgentUnitTests" or whatever, since it will contain various <unit>_unittest.cpp and not just the ConfigStore related units.

scripts/run-unit-tests.py
```diff
@@ -0,0 +1,64 @@
...
+runTestSuite("./common/unit-test/softwarecontainercommon-unittest", "CommonTest")
+runTestSuite("./libsoftwarecontainer/unit-test/softwarecontainer-unit-test", "LibUnitTests")
+
+retval = 0
+for name,exitCode in testSuites.items():
```
breakreturn: Space after comma please.

scripts/run-unit-tests.py
tobsan: I agree, I'll rename it.

agent/component-test/main.cpp
```diff
@@ -0,0 +1,42 @@
...
+#include "gmock/gmock.h"
+#include "gtest/gtest.h"
+
+#include "ivi-logging-console.h"
+#include "softwarecontainer-common.h"
+
+LOG_DEFINE_APP_IDS("PCON", "SoftwareContainer Unit Test");
+LOG_DECLARE_CONTEXT(SoftwareContainer_DefaultLogContext, "PCON", "Main context");
+
+int main(int argc, char * *argv)
```
kursatkobya: **argv

libsoftwarecontainer/component-test/main.cpp
```diff
@@ -0,0 +1,42 @@
...
+#include "gmock/gmock.h"
+#include "gtest/gtest.h"
+
+#include "ivi-logging-console.h"
+#include "softwarecontainer-common.h"
+
+LOG_DEFINE_APP_IDS("PCON", "SoftwareContainer Unit Test");
+LOG_DECLARE_CONTEXT(SoftwareContainer_DefaultLogContext, "PCON", "Main context");
+
+int main(int argc, char * *argv)
```
kursatkobya: **argv

kursatkobya (approved):
snyggt

PR #255, opened 2016-12-22
Remove userID from Execute command

kursatkobya:
userID argument is removed from execute and all related places
service tests and documents are updated accordingly


breakreturn:
The version should be bumped with this change.

tobsan (approved):

PR #256, opened 2016-12-22
Fetch default values during agent help printout

thenor:
Don't print hard coded default values, instead fetch
the default values from cmake defines.

### DIFF
--- a/agent/src/main.cpp
+++ b/agent/src/main.cpp
@@ -91,7 +91,8 @@ int main(int argc, char **argv)
     preloadOpt.set_long_name("preload");
     preloadOpt.set_short_name('p');
     preloadOpt.set_arg_description("<number>");
-    preloadOpt.set_description("Number of containers to preload, defaults to 0");
+    preloadOpt.set_description("Number of containers to preload, "
+                               "defaults to " + std::to_string(PRELOAD_COUNT));
 
     Glib::OptionEntry userOpt;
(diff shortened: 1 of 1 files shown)


agent/src/main.cpp
```diff
@@ -103,36 +104,38 @@ int main(int argc, char **argv)
...
-                                                 std::string(DEFAULT_SERVICE_MANIFEST_DIR));
+                                                 "service manifest(s) exist, defaults to \""
+                                                  + std::string(DEFAULT_SERVICE_MANIFEST_DIR) + "\"");
 
     Glib::OptionEntry sessionBusOpt;
     sessionBusOpt.set_long_name("session-bus");
     sessionBusOpt.set_short_name('b');
-    sessionBusOpt.set_description("Use the session bus instead of the system bus");
+    sessionBusOpt.set_description("Use the session bus instead of the system bus, "
+                                  "defaults to " + std::string(USE_SESSION_BUS ? "true" : "false"));
```
DunderRoffe, line 138: There is no need to call the std::string constructor, it will work fine anyway :)

thenor: Du tänker zååå.
I will change it!

thenor: I tested it and it didn't compile, so I will keep it like this. :D

breakreturn (approved):

tobsan (approved):

PR #258, opened 2017-01-04
Use Giomm dbus instead of dbus-c++

DunderRoffe:
This pull-request attempts to remove dbus-c++ and replace it with giomm dbus bindings instead.
The stubs for the new bindings were generated with a giomm dbus binding generator. However, it was not complete so that some changes had to be made later. In order to make transition, if the generator is patched, to generating at build time, I have chosen to change as little as possible of the generated code. Due to this fact there might be some redundant code and I ask you to please have that in mind when reviewing the pr.

### DIFF
--- a/agent/src/main.cpp
+++ b/agent/src/main.cpp
@@ -44,39 +44,6 @@ int signalHandler(void *data) {
     return 0;
 }
 
-/*
- * Tries to register the agent on the given bus. If the system bus is given and fails,
- * this tries to register to the session bus instead.
- *
- * Throws a ReturnCode on failure
- */
-DBus::Connection getBusConnection(bool useSessionBus)
-{
-    static constexpr const char *AGENT_BUS_NAME = "com.pelagicore.SoftwareContainerAgent";
-    std::string busStr = useSessionBus ? "session" : "system";
-    try {
-        DBus::Connection bus = useSessionBus ? DBus::Connection::SessionBus()
-                                             : DBus::Connection::SystemBus();
-
-        // Check if the name is available or not
-        bool alreadyTaken = bus.has_name(AGENT_BUS_NAME);
-        if (alreadyTaken) {
-            log_error() << AGENT_BUS_NAME << " is already taken on the " << busStr << " bus.";
-            throw ReturnCode::FAILURE;
-        }
-
-        DBus::Connection connection(bus);
-        connection.request_name(AGENT_BUS_NAME);
-        log_debug() << "Registered " << AGENT_BUS_NAME << " on the " << busStr << " bus.";
-        return connection;
-
-    } catch (DBus::Error &err) {
-        log_warning() << "Could not register " << AGENT_BUS_NAME << " to the " << busStr << " bus";
-        throw ReturnCode::FAILURE;
-    }
-}
-
-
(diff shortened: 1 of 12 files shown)


agent/src/main.cpp
```diff
@@ -237,12 +203,8 @@ int main(int argc, char **argv)
         Config config(std::move(loader), std::move(defaults), stringOptions, intOptions, boolOptions);
 
         static constexpr const char *AGENT_OBJECT_PATH = "/com/pelagicore/SoftwareContainerAgent";
-        DBus::Connection connection = getBusConnection(useSessionBus);
-        SoftwareContainerAgent agent(mainContext, config);
-        std::unique_ptr<SoftwareContainerAgentAdaptor> adaptor =
-            std::unique_ptr<SoftwareContainerAgentAdaptor>(
-                new DBusCppAdaptor(connection, AGENT_OBJECT_PATH, agent)
-            );
+        ::softwarecontainer::SoftwareContainerAgent agent(mainContext, config);
```
tobsan, line 206: Do you really need the prefix :: here?

agent/src/softwarecontaineragent_dbus_stub.cpp
```diff
@@ -0,0 +1,322 @@
+#include "softwarecontaineragent_dbus_stub.h"
+
+Glib::ustring interfaceXml0 = R"xml(
```
tobsan: Does this really work? Using "xml" as delimiter should stop processing on the first line I think. Even if it happens to work, choose something that does not show up in the xml code, like "endOfXMLCode".

tobsan (changes requested):
Looks really good overall, see inline comments.

agent/src/main.cpp
DunderRoffe: yes, we use namespaces in a super wierd way that forces this.

agent/src/softwarecontaineragentadaptor.cpp
```diff
@@ -6,80 +6,96 @@ SoftwareContainerAgentAdaptor::~SoftwareContainerAgentAdaptor()
...
 
-SoftwareContainerAgentAdaptor::SoftwareContainerAgentAdaptor(SoftwareContainerAgent &agent) :
-    m_agent(agent)
+SoftwareContainerAgentAdaptor::SoftwareContainerAgentAdaptor(::softwarecontainer::SoftwareContainerAgent &agent, bool useSessionBus) :
+    com::pelagicore::SoftwareContainerAgent(), m_agent(agent)
 {
+    std::string agentBusName = "com.pelagicore.SoftwareContainerAgent";
+    Gio::DBus::BusType busType = useSessionBus ? Gio::DBus::BUS_TYPE_SESSION
+                                               : Gio::DBus::BUS_TYPE_SYSTEM;
+    connect(busType, agentBusName);
```
tobsan, line 15: are there any risks when running connect here in the constructor?

tobsan (approved):

agent/src/softwarecontaineragentadaptor.cpp
DunderRoffe: Not really. `connect` is supposed to handles possible exceptions internally.

PR #259, opened 2017-01-04
Add Whitelist policy to CGroups Gateway

kursatkobya:
Whitelisting feauture implemented on cgroupparser
	Related unittests are added
	User dics has updated

### DIFF
--- a/libsoftwarecontainer/src/gateway/cgroups/cgroupsparser.cpp
+++ b/libsoftwarecontainer/src/gateway/cgroups/cgroupsparser.cpp
@@ -22,7 +22,12 @@
 
 namespace softwarecontainer {
 
-ReturnCode CGroupsParser::parseCGroupsGatewayConfiguration(const json_t *element, CGroupsPair &result)
+CGroupsParser::CGroupsParser()
+    : m_settings()
+{
+}
+
+ReturnCode CGroupsParser::parseCGroupsGatewayConfiguration(const json_t *element)
 {
     std::string settingKey;
     std::string settingValue;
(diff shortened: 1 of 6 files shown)


libsoftwarecontainer/src/gateway/cgroups/cgroupsparser.cpp
```diff
@@ -17,7 +22,28 @@ ReturnCode CGroupsParser::parseCGroupsGatewayConfiguration(const json_t *element
         return ReturnCode::FAILURE;
     }
 
-    result.first = settingKey;
-    result.second = settingValue;
+    log_debug() << "key : " << settingKey << " value : " << settingValue;
+
+    if ("memory.limit_in_bytes" == settingKey) {
+        if (m_settings.count(settingKey)) {
+            // if the new value is lexicographical greater than old one set is as the value
```
tobsan: This is not lexicographically comparison :)

tobsan (approved):
Looks good!

libsoftwarecontainer/src/gateway/cgroups/cgroupsparser.cpp
kursatkobya: I just change it to integer comparision after deciding to apply to only one setting

libsoftwarecontainer/src/gateway/cgroups/cgroupsparser.cpp
```diff
@@ -17,7 +22,28 @@ ReturnCode CGroupsParser::parseCGroupsGatewayConfiguration(const json_t *element
...
+            if (std::stoi(m_settings[settingKey]) < std::stoi(settingValue)) {
+                m_settings[settingKey] = settingValue;
+            }
+        } else {
+            m_settings[settingKey] = settingValue;
+        }
+    } else {
+        log_warning() << "It is possible to apply this rule through control groups of working "
```
breakreturn: I think this should be more specific and clear, e.g. "cgroups gateway config \"" <<  config << "\" is not supported by the gateway"

breakreturn (changes requested):

libsoftwarecontainer/src/gateway/cgroups/cgroupsparser.cpp
kursatkobya: But kernel can support this setting so maybe it is not told that its not supported. But we can keep saying this setting is not guaranteed to work by cgroup gateway. But i agree we should mentioned about the key. Thus i propose to keep sentence but changing "this rule" part with the actual setting key.

DunderRoffe (approved):

breakreturn (approved):

PR #260, opened 2017-01-04
Enable and test the cgroup memory subsystem

kursatkobya:
Change Vagrantfile to update kernel options
Write related component tests

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -1,6 +1,8 @@
 # -*- mode: ruby -*-
 # vi: set ft=ruby :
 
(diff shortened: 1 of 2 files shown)


tobsan (changes requested):
While this tests the configuration and activation of the cgroups gateway, it does not actually test the cgroup memory subsystem. Perhaps that should be on a service test level though. What does @breakreturn think?

DunderRoffe (approved):

tobsan (approved):

PR #261, opened 2017-01-04
Add servicetest for list containers

thenor:


servicetest/general/test_general.py
```diff
@@ -0,0 +1,122 @@
...
+
+DBUS_GW_CONFIG = [{
+    "dbus-gateway-config-session": [{
+        "direction": "outgoing",
+        "interface": "*",
+        "object-path": "*",
+        "method": "*"
+    }],
+    "dbus-gateway-config-system": []
+}]
```
tobsan: It does not make much sense to keep this config here, it should work with empty manifests as well I guess.

tobsan (changes requested):
Except for the manifest comment, my only issue is that I don't think it is pythonic to put parenthesis around tuples that are returned (or when matching). Do like this instead.

return a,b
a,b = blah()

servicetest/general/test_general.py
thenor: Ok. I just felt it was more "service test" like for me to have a mimimal manifest, but I can also remove it completely since I don't need the config in it. I will at least try not having it.

thenor:
To comment here also, and not just in person, I attempted to not use parenthesis, but I got an error. When I added the parenthesis it worked so I kept it that way. I will see if I can find a nice way to remove the parenthesis.

servicetest/testframework/lib.py
```diff
@@ -262,7 +268,7 @@ def terminate(self):
 
     def __create_container(self, config):
         self.__container_id, success = self.__agent.Create(config)
-        return success
+        return (self.__container_id, success)
```
thenor: What doesn't work is this:
        return self.__agent.Create(config)
But I can remove the parenthesis and return more than one value.

servicetest/general/test_general.py
thenor: I realized I didn't do anything with this in my latest update. I can try clearing it.

tobsan (approved):

thenor:
The last update is just a squash for nicer commits, to avoid the "updated after comments" like commit.

PR #262, opened 2017-01-05
namespaces everything to "softwarecontainer"

tobsan:
Instead of namespacing parts and saying "using namespace softwarecontainer" in softwarecontainer-common.h (which everything includes), we now namespace properly.


kursatkobya (approved):
nice

DunderRoffe (approved):
Awesome sauce

PR #265, opened 2017-01-09
Refactor checkContainer to getContainer

DunderRoffe:
Refactor checkContainer(id, &container) to getContainer(id) which makes for cleaner code


tobsan (approved):
It would be nicer with a shared_ptr instead of a raw pointer. Is that doable without changing too much of the code that uses getContainer?

thenor (approved):
Is this the entire 221 user story or only part of it?

tobsan (approved):

PR #268, opened 2017-01-10
Updated configs implementation

breakreturn:


CMakeLists.txt
```diff
@@ -151,57 +152,84 @@ if(ENABLE_ALL_DOC)
...
+# Setup configs that goes into the main config file. Macros for configs are defined in Config.cmake
+#
+
+# Used as prefix to config keys that belong to the SoftwareContainer group
+set(SC_CONFIG_GROUP "SC")
+
+add_boolean_config(${SC_CONFIG_GROUP}
+                   KEEP_CONTAINERS_ALIVE
+                   FALSE
+                   "false")
```
thenor: How come the "default value" is (cmake) bool and "config file value" a string?

cmake_modules/Configs.cmake
```diff
@@ -0,0 +1,139 @@
...
+
+    set("${UNIQUE_KEY}_ACTIVATE" "# ")
+    set("${UNIQUE_KEY}_CONFIG_FILE_VAR" ${CONFIG_FILE_VALUE})
+endmacro()
+
+
+# The boolean and integer configs are added in the same way, so this macro just
+# passes along the args to add_boolean_config().
+macro(add_boolean_config GROUP KEY DEFAULT_VALUE CONFIG_FILE_VALUE)
+    add_integer_config(${GROUP} ${KEY} ${DEFAULT_VALUE} ${CONFIG_FILE_VALUE})
```
thenor, line 122: [Copied from CMakeList.txt] How come the "default value" is (cmake) bool and "config file value" a string?

CMakeLists.txt
breakreturn: The default value is used for the definition that is used in the code later. The CMake "types" FALSE and TRUE will work for that, but they can't be used for the config file value which needs to be a string set to "true" or "false". I'll add some comment about this so it's more clear why it's done like this.

CMakeLists.txt
```diff
@@ -152,57 +155,84 @@ if(ENABLE_ALL_DOC)
...
+
+add_boolean_config(${SC_CONFIG_GROUP}
+                   USE_SESSION_BUS
+                   FALSE
+                   "false")
```
tobsan: Perhaps it is just me, but I would prefer one-liners where columns are aligned above this.

agent/src/config/configdefinition.h
```diff
@@ -0,0 +1,89 @@
...
+#include "softwarecontainer-common.h"
+#include "configtypes.h"
+
+
+namespace softwarecontainer {
+
+/**
+ * 
+ */
+typedef std::vector<std::tuple<std::string, std::string>> MandatoryConfigs;
```
tobsan: Why is this not a pair as well?

tobsan: Also, it would be nice to typedef std::pair<std::string, std::string> to something nice.

agent/src/config/configitem.h
```diff
@@ -0,0 +1,87 @@
...
+namespace softwarecontainer {
+
+class ConfigItem
+{
+public:
+    ConfigItem(const std::string &group, const std::string &name);
+    ~ConfigItem() {}
+
+    std::string key();
+    std::string group();
```
tobsan: You are doing equality checking on key and group together elsewhere in the code. Consider defining operator == for ConfigItem to reduce verbosity.

DunderRoffe (approved):
Looks good

agent/src/config/mainconfigsource.cpp
```diff
@@ -0,0 +1,247 @@
...
+
+    BoolConfig config(group, name, value);
+    config.setSource(ConfigSourceType::Main);
+
+    log_debug() << "Constructed BoolConfig with key and value: " << name << " " << value;
+
+    return config;
+}
+
+std::string MainConfigSource::_getStringValue(const std::string &group, const std::string &key) const
```
tobsan: Why this naming with leading _ about?

tobsan (changes requested):
Tests where a bit hard to read, but rest of code is nice and readable.

agent/src/config/config.h
```diff
@@ -212,73 +49,96 @@ LOG_DECLARE_CLASS_CONTEXT("CFG", "SoftwareContainer general config");
...
-    void loadConfig();
 
-    std::unique_ptr<ConfigLoaderAbstractInterface> m_loader;
-    std::unique_ptr<Glib::KeyFile> m_config;
-    std::unique_ptr<ConfigDefaults> m_defaults;
-    std::map<std::string, std::string> m_stringOptions;
-    std::map<std::string, int> m_intOptions;
-    std::map<std::string, bool> m_boolOptions;
-};
+
```
thenor: Is this something that should be kept? I heard you mention test cases, but is this also something that might be good to keep?

CMakeLists.txt
thenor: I agree. But this is a special case since it will be more readable, despite the longer lines.

agent/unit-test/config_unittest.cpp
```diff
@@ -60,505 +66,1038 @@ LOG_DECLARE_CLASS_CONTEXT("CFGL", "SoftwareContainer general config loader");
...
+
+
+
+
+
+
+
+
+
+// /*
```
thenor: Is it possible to save this somewhere? I'm not sure where, but unless they are test cases which fail or even are expected to fail or need some refactoring to work... then it feels a bit strange for me to keep them. You mentioned something about them. Is it that you are using them for debugging while developing? Perhaps they can be moved to another test suite which can be removed when not needed anymore? Unless they will be removed very soon.

thenor (approved):

CMakeLists.txt
breakreturn: Yeah it's nicer with one-liners, but some of these will be very long and I figured it was nicer to break them in the same way to make it clear they are all the same kind of call. I can make them one-liners again and just break the few that becomes overly long?

agent/src/config/config.h
breakreturn: I'm going to work this into the proper comments. It's not fully correct but it can be partly reused so I just kept it here for now, will be gone before merge.

agent/src/config/configdefinition.h
breakreturn: No good reason I think, it's probably a left over from previous implementations. Will change it.

agent/src/config/configitem.h
breakreturn: Good idea. Will take a look at that as well.

agent/src/config/mainconfigsource.cpp
breakreturn: Remains from an old implementation with public getters and internal ones. Will rename it.

CMakeLists.txt
tobsan: Sounds like a plan.

agent/src/config/config.cpp
```diff
@@ -17,205 +17,203 @@
...
-               std::vector<std::pair<std::string, std::vector<std::string>>> &dependencies,
-               const std::map<std::string, std::string> &stringOptions,
-               const std::map<std::string, int> &intOptions,
-               const std::map<std::string, bool> &boolOptions) :
-    m_loader(std::move(loader)),
-    m_defaults(std::move(defaults)),
-    m_stringOptions(stringOptions),
-    m_intOptions(intOptions),
-    m_boolOptions(boolOptions)
+void Config::checkDependencies()
```
DunderRoffe: assertDependencies is more explanatory

agent/src/config/config.cpp
```diff
@@ -17,205 +17,203 @@
...
-}
 
-std::string Config::getStringValue(const std::string &group, const std::string &key) const
-{
-    return getValue<std::string>(group, key, m_stringOptions);
+    return true;
 }
 
-int Config::getIntegerValue(const std::string &group, const std::string &key) const
+void Config::checkForMandatory()
```
DunderRoffe: I think assertMandatoryIsPresent or something along that line is more explanatory than checkForMandatory

agent/src/config/config.cpp
```diff
@@ -17,205 +17,203 @@
...
+            std::string message = "Mandatory config " + mandatory.first + "::" + mandatory.second + " not found";
+            log_error() << message;
+            throw ConfigMandatoryError(message);
+        }
+    }
 }
 
-template<typename T>
-T Config::getValue(const std::string &group, const std::string &key, const std::map<std::string, T> &options) const
+void Config::getConfigsFromSources()
```
DunderRoffe: As this is a void method consider `read` over `get`

DunderRoffe (changes requested):
All in all I think this is a great improvement. However I have a few suggestions.

agent/src/config/config.cpp
```diff
@@ -17,205 +17,203 @@
...
 }
 
-int Config::getIntegerValue(const std::string &group, const std::string &key) const
+void Config::assertAllMandatoryPresent()
 {
-    return getValue<int>(group, key, m_intOptions);
-}
+    // There might not be any configs which are mandatory
+    if (m_mandatory.empty()) {
+        return;
```
DunderRoffe: This check is not needed as you iterate over the list. I.e. if `m_mandatory` is empty then the for loop below will never be entered, hence no exception can be thrown.

agent/src/config/config.cpp
```diff
@@ -17,205 +17,203 @@
...
+    }
 
-bool Config::getBooleanValue(const std::string &group, const std::string &key) const
-{
-    return getValue<bool>(group, key, m_boolOptions);
+    bool found = false;
```
DunderRoffe: Move the declaration into the for loop

DunderRoffe (changes requested):

agent/src/config/config.cpp
```diff
@@ -17,205 +17,203 @@
...
-     */
-    for (std::string group : mandatory.groups()) {
-        log_debug() << "Checking for mandatory config group: \"" << group << "\"";
-        if (!m_config->has_group(group)) {
-            log_error() << "Missing mandatory config group: \"" + group + "\"";
-            throw ConfigMandatoryError("Missing mandatory config group: \"" + group + "\"");
-        }
+    // There might not be any dependencies
+    if (m_dependencies.empty()) {
+        return;
```
tobsan: This check is not needed since you iterate over the dependencies later. If they are empty, there will be nothing to iterate over.

agent/src/config/config.cpp
```diff
@@ -17,205 +17,203 @@
...
 
-void Config::loadConfig()
+bool Config::allDepsSatisfied(const std::vector<UniqueKey> &dependencies) const
 {
-    try {
-        m_config = m_loader->loadConfig();
-    } catch (Glib::Error &error) {
-        throw ConfigError();
+    for (const UniqueKey &dependency : dependencies) {
+        bool found = (std::find(m_allConfigs.begin(), m_allConfigs.end(), dependency) != m_allConfigs.end());
```
tobsan: Why the outer parantheses?

agent/src/config/config.cpp
```diff
@@ -17,205 +17,189 @@
...
-        if (!m_config->has_group(group)) {
-            log_error() << "Missing mandatory config group: \"" + group + "\"";
-            throw ConfigMandatoryError("Missing mandatory config group: \"" + group + "\"");
+        // Check if the dependee exists in the configs we got from the sources
+        bool found = (std::find(m_allConfigs.begin(), m_allConfigs.end(), config) != m_allConfigs.end());
+
+        // If a dependee was present among the configs, all its dependencies needs to be present as well
+        if (found) {
+            if (allDepsSatisfied(dep.second)) {
+                log_debug() << "All dependencies found for config " << group << "::" << key;
```
tobsan: Logging that everything is fine is a bit superfluous imho. You can remove the log and make:

```
if (found && !allDepsSatisfied(dep.second)) {
    log_error();
    throw ConfigDependencyError();
}
```

agent/src/config/config.cpp
```diff
@@ -17,205 +17,189 @@
...
-                default:
-                    break;
-            }
-        } catch (ConfigError &error) {
-            log_error() << "Missing mandatory config: \"" << key << "\"";
-            throw ConfigMandatoryError("Missing mandatory config: \"" + key + "\"");
+bool Config::allDepsSatisfied(const std::vector<UniqueKey> &dependencies) const
+{
+    for (const UniqueKey &dependency : dependencies) {
+        bool found = (std::find(m_allConfigs.begin(), m_allConfigs.end(), dependency) != m_allConfigs.end());
```
tobsan: Why the outer parantheses?

agent/src/config/config.cpp
```diff
@@ -17,205 +17,189 @@
...
-void Config::loadConfig()
+void Config::assertAllMandatoryPresent()
 {
-    try {
-        m_config = m_loader->loadConfig();
-    } catch (Glib::Error &error) {
-        throw ConfigError();
+    for (const UniqueKey &mandatory : m_mandatory) {
+        // A mandatory config must be found among the configs we got from the sources
+        bool found = (std::find(m_allConfigs.begin(), m_allConfigs.end(), mandatory) != m_allConfigs.end());
```
tobsan: Why the outer parentheses?

agent/src/config/config.cpp
```diff
@@ -17,205 +17,189 @@
...
+    }
+
+    for (auto config : configs) {
+        if (config.source() == ConfigSourceType::Default) {
+            return config;
+        }
+    }
+
+    log_error() << "Config has an incorrect source";
+    throw ConfigError("Config has an incorrect source");
```
tobsan: Having three for loops like this does not really look nice. Consider using std::find_if instead, although it would be the same amount of code probably...Up to you. You could also sort the vector according to source and take the first element?

agent/src/config/configloader.h
```diff
@@ -17,8 +18,12 @@
  * For further information see LICENSE
  */
 
+
 #pragma once
 
+#include "softwarecontainer-common.h"
```
tobsan: Nothing in this header needs anything from `softwarecontainer-common.h`. If it is needed in the cpp file, include it there instead.

agent/src/config/configsource.h
```diff
@@ -20,25 +20,25 @@
 #pragma once
 
 #include "softwarecontainer-common.h"
```
tobsan: Nothing in this class needs anything from `softwarecontainer-common.h` (since you no longer declare a log context).

agent/src/config/configtypes.h
```diff
@@ -0,0 +1,70 @@
...
+ * WHETHER IN AN ACTION OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION,
+ * ARISING OUT OF OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS
+ * SOFTWARE.
+ *
+ * For further information see LICENSE
+ */
+
+#pragma once
+
+#include "softwarecontainer-common.h"
```
tobsan: Nothing in this header needs `softwarecontainer-common.h`

agent/src/config/fileconfigloader.cpp
```diff
@@ -17,6 +18,8 @@
  * For further information see LICENSE
  */
 
+
+#include "softwarecontainer-common.h"
```
tobsan: This include should be in `fileconfigloader.h` instead of the cpp file, since you declare a log context there.

agent/src/config/mainconfigsource.cpp
```diff
@@ -0,0 +1,179 @@
...
+ * BE LIABLE FOR ANY SPECIAL, DIRECT, INDIRECT, OR CONSEQUENTIAL DAMAGES
+ * OR ANY DAMAGES WHATSOEVER RESULTING FROM LOSS OF USE, DATA OR PROFITS,
+ * WHETHER IN AN ACTION OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION,
+ * ARISING OUT OF OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS
+ * SOFTWARE.
+ *
+ * For further information see LICENSE
+ */
+
+#include "softwarecontainer-common.h"
```
tobsan: This is already included in the header file.

agent/src/config/mainconfigsource.cpp
```diff
@@ -0,0 +1,179 @@
...
+        std::vector<Glib::ustring> groups = m_config->get_groups();
+        for (auto &group : groups) {
+            std::vector<Glib::ustring> keys = m_config->get_keys(group);
+
+            for (auto &key : keys) {
+                m_configsByGroup[group].push_back(key);
+            }
+        }
+    } catch (Glib::Error &error) {
+        throw ConfigFileError("Error rading main config file");
```
tobsan: reading*

agent/src/config/mainconfigsource.h
```diff
@@ -0,0 +1,66 @@
...
+ * WARRANTIES WITH REGARD TO THIS SOFTWARE INCLUDING ALL IMPLIED
+ * WARRANTIES OF MERCHANTABILITY AND FITNESS. IN NO EVENT SHALL THE AUTHOR
+ * BE LIABLE FOR ANY SPECIAL, DIRECT, INDIRECT, OR CONSEQUENTIAL DAMAGES
+ * OR ANY DAMAGES WHATSOEVER RESULTING FROM LOSS OF USE, DATA OR PROFITS,
+ * WHETHER IN AN ACTION OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION,
+ * ARISING OUT OF OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS
+ * SOFTWARE.
+ *
+ * For further information see LICENSE
+ */
```
tobsan, line 18: This class has a complete lack of comments.

tobsan (changes requested):
See inline comments (mostly cosmetics), plus two general things:

1) Comments are relatively sparse
2) Shouldn't softwarecontainer.cpp use SC_BRIDGE_IP and SC_BRIDGE_NETMASK_BITS when creating the network gateway?

agent/src/config/config.cpp
DunderRoffe: This is not lisp ;)

agent/src/config/config.cpp
DunderRoffe: Sorting is more work than iterating so I say we keep the loops

agent/src/config/mainconfigsource.h
DunderRoffe: Could maybe use some comments. But most of it is barely worth commenting. I.e. self explanatory

DunderRoffe (approved):

agent/src/config/config.cpp
breakreturn: Agree

agent/src/config/config.cpp
breakreturn: I think it's more readable actually. But I don't mind changing it, it's not strictly needed i suppose.

tobsan (approved):
Looks good to me.

PR #269, opened 2017-01-10
Refactored flow for excuting in container and what is exposed in the lib

DunderRoffe:
This pull request refactors the flow of executing code inside of a container. The idea is that only FunctionJob and CommandJob is used for this purpose.

The second part of this pull request changes what is exposed in the library. I.e. moves header files from include to appropriate location under src.


libsoftwarecontainer/src/gateway/gateway.cpp
```diff
@@ -119,6 +119,11 @@ bool Gateway::hasContainer()
 
 std::shared_ptr<ContainerAbstractInterface> Gateway::getContainer()
 {
+    if (!hasContainer()) {
```
tobsan, line 122: Imho this should be an exception rather than just an error.

libsoftwarecontainer/src/gateway/network/iptableentry.h
```diff
@@ -97,6 +97,12 @@ class IPTableEntry
      */
     std::string interpretPolicy(void);
 
+
+    /**
+     * @brief Creates a string with infromation about the entry
```
tobsan: infromation?

tobsan (approved):

PR #271, opened 2017-01-11
Use a single entry-point for mounting

tobsan:
See the commit about this. Also a commit removing some unused variables.

### DIFF
--- a/libsoftwarecontainer/component-test/softwarecontainerlib_unittest.cpp
+++ b/libsoftwarecontainer/component-test/softwarecontainerlib_unittest.cpp
@@ -42,6 +42,11 @@ public:
         getSc().startGateways(config);
     }
 
+    ReturnCode bindMountInContainer(const std::string src, const std::string dst, bool readOnly)
+    {
+        return getSc().getContainer()->bindMountInContainer(src, dst, readOnly);
+    }
+
     Glib::RefPtr<Glib::MainContext> getMainContext()
     {
         return m_context;
@@ -311,8 +316,7 @@ TEST_F(SoftwareContainerApp, TestFileMounting) {
     job1.start();
     ASSERT_EQ(job1.wait(), NON_EXISTENT);
 
-    ReturnCode result = getSc().getContainer()->bindMountInContainer(tempFilename, tempFilename, true);
-    ASSERT_TRUE(isSuccess(result));
+    ASSERT_TRUE(isSuccess(bindMountInContainer(tempFilename, tempFilename, true)));
 
     FunctionJob job2(getSc(), [&] () {
         return isFile(tempFilename) ? EXISTENT : NON_EXISTENT;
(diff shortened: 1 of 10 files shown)


tobsan:
Just realized I didn't update docs, hang on with merging.

libsoftwarecontainer/include/container.h
```diff
@@ -219,6 +208,34 @@ class Container :
...
+     *
+     * Will create missing parent directories.
+     *
+     * @param pathOnHost The path to the directory that shall be bind mounted on the host system.
+     * @param pathInContainer Where to mount the file in the container.
+     * @param readonly Sets if the mount should be read only or read write
+     *
+     * @return SUCCESS if everything worked as expected, FAILURE otherwise.
+     */
+    ReturnCode bindMountFolderInContainer(const std::string &pathOnHost, const std::string &pathInContainer, bool readonly = true);
```
breakreturn: Could we rename this to not use "folder"?

libsoftwarecontainer/src/container.cpp
```diff
@@ -516,6 +516,24 @@ ReturnCode Container::destroy(unsigned int timeout)
     return ReturnCode::SUCCESS;
 }
 
+ReturnCode Container::bindMountInContainer(const std::string &pathOnHost,
```
breakreturn: Throughout it's path-ON-host and path-IN-container. Any reason for them to be different w.r.t. "in" and "on"?

libsoftwarecontainer/include/container.h
tobsan: Gotcha, will change to directory.

libsoftwarecontainer/src/container.cpp
tobsan: no, I will streamline.

libsoftwarecontainer/component-test/softwarecontainerlib_unittest.cpp
```diff
@@ -365,7 +365,7 @@ TEST_F(SoftwareContainerApp, TestFileMounting) {
     job1.start();
     ASSERT_EQ(job1.wait(), NON_EXISTENT);
 
-    ReturnCode result = getSc().getContainer()->bindMountFileInContainer(tempFilename, tempFilename, true);
+    ReturnCode result = getSc().getContainer()->bindMountInContainer(tempFilename, tempFilename, true);
```
thenor: Is it not possible to write the result and assert as a one-liner?

thenor: Comment is for all the below cases which are written in the same way.

DunderRoffe (approved):
Looks nice. Although I like "bindMount" better than "bindMountInContainer".

libsoftwarecontainer/component-test/softwarecontainerlib_unittest.cpp
tobsan: Hmmm yes, I'll see what I can do.

thenor (approved):

PR #274, opened 2017-01-12
SoftwareContainer Exceptions

thenor:
Add an internal SoftwareContainer exception class and
remove internal throwing of ReturnCode.

### DIFF
--- a/agent/src/softwarecontaineragent.cpp
+++ b/agent/src/softwarecontaineragent.cpp
@@ -42,7 +42,7 @@ SoftwareContainerAgent::SoftwareContainerAgent(
         m_shutdownContainers = !config.getBoolValue(ConfigDefinition::SC_GROUP,
                                                     ConfigDefinition::SC_KEEP_CONTAINERS_ALIVE_KEY);
     } catch (ConfigError &error) {
-        throw ReturnCode::FAILURE;
+        throw;
     }
 
     // Get all configs for the workspace
@@ -57,7 +57,7 @@ SoftwareContainerAgent::SoftwareContainerAgent(
         lxcConfigPath = config.getStringValue(ConfigDefinition::SC_GROUP,
                                               ConfigDefinition::SC_LXC_CONFIG_PATH_KEY);
     } catch (ConfigError &error) {
-        throw ReturnCode::FAILURE;
+        throw;
     }
 
     // Create and set all values on workspace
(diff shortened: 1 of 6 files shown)


agent/src/softwarecontaineragent.cpp
```diff
@@ -79,15 +79,15 @@ SoftwareContainerAgent::SoftwareContainerAgent(
...
+        throw;
     }
 
     try {
         m_filteredConfigStore = std::make_shared<FilteredConfigStore>(serviceManifestDir);
         m_defaultConfigStore = std::make_shared<DefaultConfigStore>(defaultServiceManifestDir);
     } catch (ReturnCode err) {
         log_error() << "Failed to initialize ConfigStore";
-        throw ReturnCode::FAILURE;
+        throw SoftwareContainerAgentError();
```
tobsan, line 97: It would make sense to use the message that we log as message to the exception also, in all places in this file.

agent/src/softwarecontaineragent.h
```diff
@@ -47,6 +48,104 @@ namespace softwarecontainer {
...
+    }
+
+    PreloadError(const std::string &message):
+        SoftwareContainerAgentError(message)
+    {
+    }
+
+    virtual const char *what() const throw()
+    {
+        return SoftwareContainerAgentError::m_message.c_str();
```
tobsan: Since you're inheriting `SoftwareContainerAgentError`, you should return `m_message.c_str()` instead.

agent/src/softwarecontaineragent.h
```diff
@@ -47,6 +48,104 @@ namespace softwarecontainer {
...
+    }
+
+    WorkspaceError(const std::string &message):
+        SoftwareContainerAgentError(message)
+    {
+    }
+
+    virtual const char *what() const throw()
+    {
+        return SoftwareContainerAgentError::m_message.c_str();
```
tobsan: Use `m_message.c_str()` here also, and in all the places in this file where you inherit something where m_message is protected.

tobsan (changes requested):

agent/src/softwarecontaineragent.cpp
thenor: I agree. I have already started doing this. Also the exceptions might be called something more specific.

agent/src/softwarecontaineragent.h
breakreturn: Since the parent is initialized with the message, shouldn't it be enough to just let the method be in the parent and not implement it again here?

thenor:
I will abandon this pull request and instead create a new one, since team members had dependencies to the SoftwareContainerError base class.

PR #275, opened 2017-01-12
bindMount creates any leading missing dirs

tobsan:
This goes for both files and directories.


DunderRoffe (approved):
Overall I think this looks  good. However, you could merge the multiple depth folder test with the generic folder mounting test. Having them separate does not add that much value.

tobsan:
@DunderRoffe since they are handled slightly differently in the code it does make sense to test both of them imho.

PR #276, opened 2017-01-12
devicenodegateway creates parent paths when needed

tobsan:
Otherwise mknod will not work if the parent path does not exist.

### DIFF
--- a/common/filetoolkitwithundo.h
+++ b/common/filetoolkitwithundo.h
@@ -68,7 +68,7 @@ protected:
      *
      * This is useful if one creates a file in some other way and want it deleted later.
      */
-    void watchFile(const std::string &path);
+    void markFileForDeletion(const std::string &path);
 
     /**
      * @brief overlayMount Mount a directory with an overlay on top of it. An overlay protects
--- a/common/filetoolkitwithundo.h
+++ b/common/filetoolkitwithundo.h
@@ -45,7 +45,6 @@ public:
             bool readOnly,
             bool enableWriteBuffer=false);
 
-    ReturnCode writeToFile(const std::string &path, const std::string &content);
 
     /**
      * @brief tempDir Creates a temporary directory at templatePath.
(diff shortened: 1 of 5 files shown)


libsoftwarecontainer/src/gateway/devicenode/devicenodegateway.cpp
```diff
@@ -65,12 +66,21 @@ ReturnCode DeviceNodeGateway::applySettings()
     for (auto &dev : devlist) {
         log_info() << "Mapping device " << dev.name;
 
+        std::string devicePathInContainerOnHost = getContainer()->rootFS() + dev.name;
+        std::string deviceParent = dirname(strdup(devicePathInContainerOnHost.c_str()));
+
+        if (!isDirectory(deviceParent) && isError(createDirectory(deviceParent))) {
```
DunderRoffe, line 79: In order to get better error output, would be nice to catch if `deviceParent` already exists but is not a directory.

libsoftwarecontainer/src/gateway/devicenode/devicenodegateway.cpp
```diff
@@ -83,6 +93,8 @@ ReturnCode DeviceNodeGateway::applySettings()
             if (job.wait() != 0) {
```
DunderRoffe, line 108: Would be nice to have a better, or maybe more verbose, way of determine if a job is a failure or not. `job.wait() != 0` is not very self explanatory.

Since all of the code that is run inside a container now use either FunctionJob or CommanJob, I think we should invest in some convenience in these classes. For instance, isSuccess() or isError() would be nice.

Might be out of scope for the pull request though...

common/filetoolkitwithundo.h
```diff
@@ -59,6 +58,18 @@ class FileToolkitWithUndo
...
+     * @brief Writes to a file (and optionally create it)
+     */
+    ReturnCode writeToFile(const std::string &path, const std::string &content);
+
+    /*
+     * @brief Creates a file cleanup handler for a specific file.
+     *
+     * This is useful if one creates a file in some other way and want it deleted later.
+     */
+    void watchFile(const std::string &path);
```
DunderRoffe: If this method is purely used for deleting files later then I think we can find a better name :)

DunderRoffe (changes requested):

libsoftwarecontainer/src/gateway/devicenode/devicenodegateway.cpp
tobsan: This is a good idea, but yes, out of scope here I think.

common/filetoolkitwithundo.h
tobsan: I agree.

DunderRoffe (approved):

PR #277, opened 2017-01-13
Add cgroup gateway service tests

kursatkobya:
Service tests are written 
Documents are updated

### DIFF
--- a/doc/chapters/gateways/10-cgroups-gateway.rst
+++ b/doc/chapters/gateways/10-cgroups-gateway.rst
@@ -18,10 +18,22 @@ with the value to apply.
 No syntax or other checks for correctness is performed on the key/value pairs,
 see the `lxc.container.conf` man page for more details about valid settings.
 
-Before appying the configuration it is advised to check if the cgroup option is enabled in
-the system's kernel. Following command will list available cgroup options::
+Kernel Options
+--------------
+To be able to limit memory usage with cgroups, ``CONFIG_CGROUPS``, ``CONFIG_MEMCG`` and
+``CONFIG_MEMCG_SWAP`` should be enabled. Before appying the configuration it is
+advised to check if the cgroup option is enabled with swapaccount=1 feauture in the system's
+kernel. The Following command will list available cgroup options::
 
 	cat /proc/1/cgroup
+
+Limiting Memory
+---------------
+To be able to limit memory, ``memory.limit_in_bytes`` should be set to desired value with gateway
+configuration options. The operating system also uses swap memory which means if limiting rss and swap
+memories is the desired action then ``memory.memsw.limit_in_bytes`` setting should also be specified
+in gateway configuration. Note that value of ``memory.memsw.limit_in_bytes`` indicates rss + swap
+memory, thus it cannot be less than ``memory.limit_in_bytes``.
 
 Whitelisting
 ------------
@@ -38,15 +50,19 @@ Example gateway config::
     [
         {
             "setting": "memory.limit_in_bytes",
-            "value": "12000"
+            "value": "12000000"
         },
         {
             "setting": "memory.limit_in_bytes",
-            "value": "500"
+            "value": "5000"
+        },
+        {
+            "setting": "memory.memsw.limit_in_bytes",
+            "value": "12000000"
         }
     ]
 
 The root object is an array of setting key/value pair objects. Each key/value pair
 must have the ``setting`` and ``value`` defined. In the example above value of memory.limit_in_bytes
-will be set to 12000 due to whitelisting policy mentioned above.
+will be set to 12000000 due to whitelisting policy mentioned above.
 
--- a/servicetest/cgroups/test_cgroups.py
+++ b/servicetest/cgroups/test_cgroups.py
@@ -1,0 +1,178 @@
+
+# Copyright (C) 2016-2017 Pelagicore AB
+#
+# Permission to use, copy, modify, and/or distribute this software for
+# any purpose with or without fee is hereby granted, provided that the
+# above copyright notice and this permission notice appear in all copies.
+#
+# THE SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL
+# WARRANTIES WITH REGARD TO THIS SOFTWARE INCLUDING ALL IMPLIED
+# WARRANTIES OF MERCHANTABILITY AND FITNESS. IN NO EVENT SHALL THE AUTHOR
+# BE LIABLE FOR ANY SPECIAL, DIRECT, INDIRECT, OR CONSEQUENTIAL DAMAGES
+# OR ANY DAMAGES WHATSOEVER RESULTING FROM LOSS OF USE, DATA OR PROFITS,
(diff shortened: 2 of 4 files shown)


doc/chapters/gateways/10-cgroups-gateway.rst
```diff
@@ -18,11 +18,23 @@ with the value to apply.
 No syntax or other checks for correctness is performed on the key/value pairs,
 see the `lxc.container.conf` man page for more details about valid settings.
 
-Before appying the configuration it is advised to check if the cgroup option is enabled in
-the system's kernel. Following command will list available cgroup options::
+Kernel Options
+--------------
+To be able to limit memory usage with cgroups, ``CONFIG_CGROUPS``, ``CONFIG_MEMCG`` and
+``CONFIG_MEMCG_SWAP`` should be enabled. And also before appying the configuration it is
```
thenor: I would prefer to not start the sentence with "And", so I would remove "And also" (so the sentence starts with "Before applying" (also typo in applying).

doc/chapters/gateways/10-cgroups-gateway.rst
```diff
@@ -18,11 +18,23 @@ with the value to apply.
...
+advised to check if the cgroup option is enabled with swapaccount=1 feauture in the system's
+kernel. Following command will list available cgroup options::
```
thenor: Add "the": "The following command..."

doc/chapters/gateways/10-cgroups-gateway.rst
```diff
@@ -18,11 +18,23 @@ with the value to apply.
...
 
 	cat /proc/1/cgroup
 
+Limiting Memory
+---------------
+To be able to limit memory, ``memory.limit_in_bytes`` should be set to desired value with gateway
+configuration options. But the operating system uses swap memory as well. if limiting all memories
```
thenor: I would prefer if you don't start sentences with "But".

servicetest/cgroups/test_cgroups.py
```diff
@@ -0,0 +1,178 @@
+
+# Copyright (C) 2016 - 2017 Pelagicore AB
```
thenor: For consistency, can we decide how to write this? ping @tobsan (component owner). "2016-2017" or "2016 - 2017"?

servicetest/cgroups/test_cgroups.py
```diff
@@ -0,0 +1,178 @@
...
+    """
+    def test_small_threshold(self):
+        """ Test if the memory limiting with cgroup gateway is working as expected
+            which behavior is when allocated more memory than limit it should fail
+        """
+        try:
+            sc = Container()
+            sc.start(DATA)
+
+            success = sc.set_capabilities(["test.cap.small.threshold"])
```
thenor: Feels weird to call the variable "success", I would prefer e.g. "result" since it is not expected to be successful.

thenor: Is this how other tests have been written? In other suites, I mean.

thenor:
Just a few questions left before I want to set "Approved".

servicetest/cgroups/test_cgroups.py
kursatkobya: the return value of the function is not used in every case. But it is written this way when they are used. And also in many places in the testframework lib those kind of variables named as success. But for this instance i agree with you and will change it.

servicetest/testframework/testhelper.py
```diff
@@ -180,10 +171,44 @@ def __file_name(self):
...
+
+    # Below methods are executed inside container
+    def allocate(self, size):
+        buffer = ""
+        step = size / 10
+        while (size > len(buffer)):
+            try:
+                buffer = buffer + ('a' * step)
+            except:
+                return False
```
DunderRoffe, line 191: Is it possible to release the memory in this case or is it too late? If it is too late, it would be nice with a comment about that.

servicetest/cgroups/test_cgroups.py
thenor: I think the majority of files that already have this have used "2016-2017" so perhaps best to remove the spaces. I think it looks nicer if it's the same in all files.

tobsan: This really doesn't matter, but I also thing 2016-2017 looks nicer.

kursatkobya: changed that way

DunderRoffe (approved):

doc/chapters/gateways/10-cgroups-gateway.rst
```diff
@@ -18,11 +18,23 @@ with the value to apply.
...
+``CONFIG_MEMCG_SWAP`` should be enabled. Before appying the configuration it is
+advised to check if the cgroup option is enabled with swapaccount=1 feauture in the system's
+kernel. The Following command will list available cgroup options::
 
 	cat /proc/1/cgroup
 
+Limiting Memory
+---------------
+To be able to limit memory, ``memory.limit_in_bytes`` should be set to desired value with gateway
+configuration options. The operating system also uses swap memory which means if limiting rss and swap
```
tobsan, line 33: what is rss here? Do you mean RAM?

servicetest/cgroups/test_cgroups.py
```diff
@@ -0,0 +1,178 @@
...
+                                           " --test-dir " +
+                                           sc.get_bind_dir() +
+                                           " --do-allocate " +
+                                           str(memory_limitation))
+            assert result is True
+            # wait 5 seconds for previous operation to end
+            time.sleep(5)
+            helper = CGroupHelper(CURRENT_DIR)
+            allocation_return = helper.result()
+            assert allocation_return < memory_limitation
```
tobsan, line 157: Dunno if this is possible, but shouldn't one also check that it is not too small? Like the allocated memory is "close" to the limit?

doc/chapters/gateways/10-cgroups-gateway.rst
kursatkobya: resident set size (RSS), basically part of a RAM.

servicetest/cgroups/test_cgroups.py
kursatkobya: We can write a static value to examine but since it is hard to automatize it i would avoid it. Yet to automatize we need to calculate the kernel and user space uses before we allocate. 

Maybe memory.usage_in_bytes can be compared before and after operation.

doc/chapters/gateways/10-cgroups-gateway.rst
tobsan: Ok, that should be described then I think.

servicetest/cgroups/test_cgroups.py
tobsan: In the test helper it looks like it tries to allocate 1/10th at a time, perhaps one could make a check that it is something like 0.9*maximum? But if this is not reasonable, then nevermind.

kursatkobya: It can be variable due to kernel space and user space programs allocation it could fail just after 0.9 *maximum  once, and could fail after allocation 0.6 of maximum in another time.

PR #282, opened 2017-01-16
Fixed bitshifting error in netw-GW

breakreturn:


kursatkobya (approved):

tobsan:
Will this be part of an upcoming PR instead?

breakreturn:
Yes, Kursat is working on a test for the problem and I think it makes more sense to have the fix committed together with that.

PR #284, opened 2017-01-16
Add exception classes for Agent and ConfigStore

thenor:
Add ConfigStoreError, which inherits from SoftwareContainerError
Use ConfigStoreError in BaseConfigStore and SoftwareContainerAgent.

### DIFF
--- a/agent/src/capability/baseconfigstore.cpp
+++ b/agent/src/capability/baseconfigstore.cpp
@@ -23,25 +23,19 @@ namespace softwarecontainer {
 
 BaseConfigStore::BaseConfigStore(const std::string &inputPath)
 {
-    ReturnCode retval = ReturnCode::SUCCESS;
-
     if (inputPath.empty()) {
         //File path is empty, this is ok
         return;
     }
 
     if (isDirectory(inputPath)) {
-        retval = readCapsFromDir(inputPath);
+        readCapsFromDir(inputPath);
     } else if (isFile(inputPath)) {
-        retval = readCapsFromFile(inputPath);
+        readCapsFromFile(inputPath);
     } else {
-        //Could not find a matching clause
-        log_error() << "\"" << inputPath << "\"" << " does not exist";
-        retval = ReturnCode::FAILURE;
-    }
-
-    if (isError(retval)) {
-        throw ReturnCode::FAILURE;
+        std::string errorMessage = "Path to service manifest does not exist: \"" + inputPath + "\"";
+        log_error() << errorMessage;
+        throw ServiceManifestPathError(errorMessage);
     }
 }
 
@@ -49,147 +43,157 @@ BaseConfigStore::~BaseConfigStore()
 {
 }
 
-ReturnCode BaseConfigStore::readCapsFromDir(const std::string &dirPath)
+void BaseConfigStore::readCapsFromDir(const std::string &dirPath)
 {
+    std::string errorMessage;
+
     if (dirPath.compare("/") == 0) {
-        log_warning() << "Searching for configuration files from root dir not allowed";
-        return ReturnCode::FAILURE;
+        errorMessage = "Searching for configuration files from root dir not allowed";
+        log_error() << errorMessage;
+        throw ServiceManifestPathError(errorMessage);
     }
     if (isDirectoryEmpty(dirPath)) {
         log_warning() << "Path to configuration files is empty: " << dirPath;
-        return ReturnCode::SUCCESS;
+        return;
     }
 
     std::vector<std::string> files = fileList(dirPath);
 
     if (files.empty()) {
         log_info() << "No configuration files found: " << dirPath;
-        return ReturnCode::SUCCESS;
+        return;
     }
 
     for (std::string file : files) {
         std::string filePath = dirPath + file;
 
-        if (isError(readCapsFromFile(filePath))) {
-            log_warning() << "Could not parse a file in directory: " << filePath;
-            return ReturnCode::FAILURE;
-        }
+        readCapsFromFile(filePath);
     }
-
-    return ReturnCode::SUCCESS;
 }
 
-ReturnCode BaseConfigStore::readCapsFromFile(const std::string &filePath)
+void BaseConfigStore::readCapsFromFile(const std::string &filePath)
 {
+    std::string errorMessage;
+
     if (!isJsonFile(filePath)) {
-        log_debug() << "File is not a json file: " << filePath;
-        return ReturnCode::FAILURE;
+        errorMessage = "File is not a json file: " + filePath;
+        log_debug() << errorMessage;
+        throw ServiceManifestParseError(errorMessage);
     }
 
     json_error_t error;
     json_t *fileroot = json_load_file(filePath.c_str(), 0, &error);
 
     if (nullptr == fileroot) {
-        log_error() << "Could not parse Service Manifest: "
-                    << filePath << ":" << error.line <<" : " << error.text;
-        return ReturnCode::FAILURE;
+        errorMessage = "Could not parse Service Manifest: "
+            + filePath + ":" + std::to_string(error.line) + " : " + std::string(error.text);
+        log_error() << errorMessage;
+        throw ServiceManifestParseError(errorMessage);
     } else if (!json_is_object(fileroot)) {
-        log_error() << "Configuration is not a json object: \n"
-                    << filePath << ":" << error.line <<" : " << error.text;
-        return ReturnCode::FAILURE;
+        errorMessage = "Configuration is not a json object: "
+            + filePath + ":" + std::to_string(error.line) + " : " + std::string(error.text);
+        log_error() << errorMessage;
+        throw ServiceManifestParseError(errorMessage);
     }
 
     json_t *capabilities = json_object_get(fileroot, "capabilities");
     if (nullptr == capabilities) {
-        log_error() << "Could not parse Capabilities";
-        return ReturnCode::FAILURE;
+        errorMessage = "Could not parse Capabilities in file: " + filePath;
+        log_error() << errorMessage;
+        throw CapabilityParseError(errorMessage);
     }
     if (!json_is_array(capabilities)) {
-        log_error() << "Capabilities is not an array";
-        return ReturnCode::FAILURE;
+        errorMessage = "Capabilities is not an array, in file: " + filePath;
+        log_error() << errorMessage;
+        throw CapabilityParseError(errorMessage);
     }
 
     // Can't use json_decref on fileroot, it removes objects too early
-
-    return parseCapabilities(capabilities);
+    parseCapabilities(capabilities);
 }
 
-ReturnCode BaseConfigStore::parseCapabilities(json_t *capabilities)
+void BaseConfigStore::parseCapabilities(json_t *capabilities)
 {
     size_t i;
     json_t *capability;
+    std::string errorMessage;
 
     log_debug() << "Size of capabilities is " << std::to_string(json_array_size(capabilities));
 
     json_array_foreach(capabilities, i, capability) {
         if (!json_is_object(capability)) {
-            log_error() << "Capability is not a json object";
-            return ReturnCode::FAILURE;
+            errorMessage = "Capability is not a json object";
+            log_error() << errorMessage;
+            throw CapabilityParseError(errorMessage);
         }
 
         std::string capName;
         if (!JSONParser::read(capability, "name", capName)) {
-            log_error() << "Could not read Capability name";
-            return ReturnCode::FAILURE;
+            errorMessage = "Could not read Capability name";
+            log_error() << errorMessage;
+            throw CapabilityParseError(errorMessage);
         }
 
         json_t *gateways = json_object_get(capability, "gateways");
         if (nullptr == gateways) {
-            log_error() << "Could not read Gateways";
-            return ReturnCode::FAILURE;
+            errorMessage = "Could not read Gateways";
+            log_error() << errorMessage;
+            throw CapabilityParseError(errorMessage);
         }
         if (!json_is_array(gateways)) {
-            log_error() << "Gateways is not an array";
-            return ReturnCode::FAILURE;
+            errorMessage = "Gateways is not an array";
+            log_error() << errorMessage;
+            throw CapabilityParseError(errorMessage);
         }
 
         log_debug() << "Found capability \"" << capName << "\", parsing gateways...";
 
-        ReturnCode parseGatewayResult = parseGatewayConfigs(capName, gateways);
-        if (isError(parseGatewayResult)) {
-            return ReturnCode::FAILURE;
-        }
+        parseGatewayConfigs(capName, gateways);
     }
-    return ReturnCode::SUCCESS;
 }
 
-ReturnCode BaseConfigStore::parseGatewayConfigs(std::string capName, json_t *gateways)
+void BaseConfigStore::parseGatewayConfigs(std::string capName, json_t *gateways)
 {
     if (m_capMap.count(capName) > 0) {
         log_debug() << "Capability " << capName << " already loaded.";
-        return ReturnCode::SUCCESS;
+        return;
     }
 
     size_t i;
     json_t *gateway;
+    std::string errorMessage;
 
     GatewayConfiguration gwConf;
     json_array_foreach(gateways, i, gateway) {
         if (!json_is_object(gateway)) {
-            log_error() << "Gateway is not a json object";
-            return ReturnCode::FAILURE;
+            errorMessage = "The \"gateway\" key in the Service Manifest is not a json object";
+            log_error() << errorMessage;
+            throw CapabilityParseError(errorMessage);
         }
 
         std::string gwID;
         if (!JSONParser::read(gateway, "id", gwID)) {
-            log_error() << "Could not read Gateway ID";
-            return ReturnCode::FAILURE;
+            errorMessage = "Could not read the ID of the \"gateway\""
+                " object in the Service Manifest";
+            log_error() << errorMessage;
+            throw CapabilityParseError(errorMessage);
         }
 
         json_t *confs = json_object_get(gateway, "config");
         if (nullptr == confs) {
-            log_error() << "Could not read Gateway Config";
-            return ReturnCode::FAILURE;
+            errorMessage = "Could not read the \"gateway\" object's configuration element "
+                "(" + gwID+ ")";
+            log_error() << errorMessage;
+            throw CapabilityParseError(errorMessage);
         }
         if (!json_is_array(confs)) {
-            log_error() << "Gateway Config is not an array";
-            return ReturnCode::FAILURE;
+            errorMessage = "The \"gateway\" object's configuration is not an array (" + gwID+ ")";
+            log_error() << errorMessage;
+            throw CapabilityParseError(errorMessage);
         }
         gwConf.append(gwID, confs);
     }
     m_capMap[capName] = gwConf;
-
-    return ReturnCode::SUCCESS;
 }
 
 std::vector<std::string> BaseConfigStore::fileList(const std::string &dirPath)
--- a/agent/src/capability/baseconfigstore.h
+++ b/agent/src/capability/baseconfigstore.h
@@ -25,6 +25,7 @@
 #pragma once
 #include "softwarecontainer-common.h"
 #include "gatewayconfig.h"
+#include "configstoreerror.h"
 
 #include <jsonparser.h>
 
@@ -40,7 +41,11 @@ public:
      * (of file type json) in the input path, and stores the Capabilities'
      * Gateway configurations.
      *
-     * @throws ReturnCode::FAILURE if parsing of the json file(s) fails
+     * @throws ServiceManifestPathError if the path to the json file(s) is incorrectly formatted
+     * or if the path is not allowed
+     * @throws ServiceManifestParseError if parsing of the json file(s) fails
+     * @throws CapabilityParseError if parsing of one or more Capabilities or Gateway
+     * Configurations in a file is unsuccessful
      */
     BaseConfigStore(const std::string &filePath);
 
@@ -57,11 +62,10 @@ private:
      * If a capability with the same name already exists in the storage
      * the gateways will be appended to the previously stored gatway list.
      *
-     * @return ReturnCode::SUCCESS if successful
-     * @return ReturnCode::FAILURE if parsing of directory or files is unsuccessful
+     * @throws ServiceManifestPathError if the path to the Service Manifest(s) is not allowed
      *
      */
-    virtual ReturnCode readCapsFromDir(const std::string &dirPath);
+    virtual void readCapsFromDir(const std::string &dirPath);
 
     /**
      * @brief Reads a Service Manifest file, of type json, and adds the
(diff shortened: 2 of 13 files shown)


agent/src/capability/baseconfigstore.cpp
```diff
@@ -23,173 +23,201 @@ namespace softwarecontainer {
...
         return;
     }
 
     if (isDirectory(inputPath)) {
-        retval = readCapsFromDir(inputPath);
+        try {
+            readCapsFromDir(inputPath);
+        } catch(ConfigStoreError &error) {
+            throw;
+        }
```
tobsan: If you're not going to do anything with the error except re-throwing it, you don't have to catch it at all. Then it will just propagate upwards.

agent/src/capability/baseconfigstore.cpp
```diff
@@ -23,173 +23,201 @@ namespace softwarecontainer {
...
 
-        if (isError(readCapsFromFile(filePath))) {
-            log_warning() << "Could not parse a file in directory: " << filePath;
-            return ReturnCode::FAILURE;
+        try {
+            readCapsFromFile(filePath);
+        } catch(ConfigStoreError &error) {
+            errorMessage = "Could not parse a file in directory";
+            log_warning() << errorMessage << ": " << filePath;
+            throw ServiceManifestPathError(errorMessage);
```
tobsan: Why not include the file path in the exception message also?

agent/src/capability/baseconfigstore.cpp
```diff
@@ -23,173 +23,201 @@ namespace softwarecontainer {
...
         std::string filePath = dirPath + file;
 
-        if (isError(readCapsFromFile(filePath))) {
-            log_warning() << "Could not parse a file in directory: " << filePath;
-            return ReturnCode::FAILURE;
+        try {
+            readCapsFromFile(filePath);
+        } catch(ConfigStoreError &error) {
+            errorMessage = "Could not parse a file in directory";
+            log_warning() << errorMessage << ": " << filePath;
```
tobsan: Shouldn't this really be log_error() instead?

agent/src/capability/baseconfigstore.cpp
```diff
@@ -23,173 +23,201 @@ namespace softwarecontainer {
...
+void BaseConfigStore::readCapsFromFile(const std::string &filePath)
 {
+    std::string errorMessage;
+
     if (!isJsonFile(filePath)) {
-        log_debug() << "File is not a json file: " << filePath;
-        return ReturnCode::FAILURE;
+        errorMessage = "File is not a json file";
+        log_debug() << errorMessage << ": " << filePath;
+        throw ServiceManifestPathError(errorMessage);
```
tobsan: Here also, why not include the file path in the exception message?

agent/src/capability/baseconfigstore.cpp
```diff
@@ -23,173 +23,201 @@ namespace softwarecontainer {
...
+        errorMessage = "Could not parse Service Manifest";
+        log_error() << errorMessage << ": " << filePath << ":" << error.line <<" : " << error.text;
+        throw ServiceManifestPathError(errorMessage);
     } else if (!json_is_object(fileroot)) {
-        log_error() << "Configuration is not a json object: \n"
-                    << filePath << ":" << error.line <<" : " << error.text;
-        return ReturnCode::FAILURE;
+        errorMessage = "Configuration is not a json object";
+        log_error() << errorMessage << ": " << filePath << ":" << error.line <<" : " << error.text;
+        throw ServiceManifestPathError("Service Manifest parse error - " + errorMessage);
```
tobsan: The type here implies service manifest error, no need to repeat that on this level.

agent/src/capability/baseconfigstore.cpp
```diff
@@ -23,173 +23,201 @@ namespace softwarecontainer {
...
+        throw CapabilityParseError(errorMessage);
     }
 
     // Can't use json_decref on fileroot, it removes objects too early
-
-    return parseCapabilities(capabilities);
+    try {
+        parseCapabilities(capabilities);
+    } catch(CapabilityParseError &error) {
+        throw;
```
tobsan: No need to catch to just rethrow.

agent/src/capability/baseconfigstore.cpp
```diff
@@ -23,173 +23,201 @@ namespace softwarecontainer {
...
 
     log_debug() << "Size of capabilities is " << std::to_string(json_array_size(capabilities));
 
     json_array_foreach(capabilities, i, capability) {
         if (!json_is_object(capability)) {
-            log_error() << "Capability is not a json object";
-            return ReturnCode::FAILURE;
+            errorMessage = "Capability is not a json object";
+            log_error() << errorMessage;
+            throw CapabilityParseError("Capability parse error - " + errorMessage);
```
tobsan: We already know this is a capability parse error, the type says that. This goes for the rest of the occurrences in this file as well.

agent/src/capability/baseconfigstore.cpp
```diff
@@ -23,173 +23,201 @@ namespace softwarecontainer {
...
         log_debug() << "Found capability \"" << capName << "\", parsing gateways...";
 
-        ReturnCode parseGatewayResult = parseGatewayConfigs(capName, gateways);
-        if (isError(parseGatewayResult)) {
-            return ReturnCode::FAILURE;
+        // TODO (ethenor): Is this correct?
+        try {
+            parseGatewayConfigs(capName, gateways);
+        } catch(CapabilityParseError &error) {
+            throw;
```
tobsan: No need to catch just to re-throw.

agent/src/capability/baseconfigstore.cpp
```diff
@@ -23,173 +23,201 @@ namespace softwarecontainer {
...
         }
     }
-    return ReturnCode::SUCCESS;
+    //return;
```
tobsan: Why the comment? If nothing else happens in this method, the return can be left out.

agent/src/softwarecontaineragent.cpp
```diff
@@ -42,7 +42,7 @@ SoftwareContainerAgent::SoftwareContainerAgent(
         m_shutdownContainers = !config.getBoolValue(ConfigDefinition::SC_GROUP,
                                                     ConfigDefinition::SC_KEEP_CONTAINERS_ALIVE_KEY);
     } catch (ConfigError &error) {
-        throw ReturnCode::FAILURE;
+        throw;
```
tobsan: You don't have to catch if you're just going to throw the same without doing anything else.

agent/src/softwarecontaineragent.cpp
```diff
@@ -68,13 +68,15 @@ SoftwareContainerAgent::SoftwareContainerAgent(
         m_softwarecontainerWorkspace->m_containerConfigPath = lxcConfigPath;
         m_softwarecontainerWorkspace->m_containerShutdownTimeout = shutdownTimeout;
     } catch (ReturnCode err) {
-        log_error() << "Failed to set up workspace";
-        throw ReturnCode::FAILURE;
+        std::string errorMessage = "Failed to set up workspace";
+        log_error() << errorMessage;
+        throw SoftwareContainerAgentError("SoftwareContainer Agent - " + errorMessage);
```
tobsan: We know this is an agent error by looking at the type.

agent/src/softwarecontaineragent.cpp
```diff
@@ -85,16 +87,17 @@ SoftwareContainerAgent::SoftwareContainerAgent(
                                                    ConfigDefinition::SC_SERVICE_MANIFEST_DIR_KEY);
         defaultServiceManifestDir = config.getStringValue(ConfigDefinition::SC_GROUP,
                                                           ConfigDefinition::SC_DEFAULT_SERVICE_MANIFEST_DIR_KEY);
-    } catch (ConfigError &error) {
-        throw ReturnCode::FAILURE;
+    } catch (SoftwareContainerAgentError &error) {
+        throw;
```
tobsan: Don't catch just to throw.

agent/src/softwarecontaineragent.cpp
```diff
@@ -85,16 +87,17 @@ SoftwareContainerAgent::SoftwareContainerAgent(
...
     try {
         m_filteredConfigStore = std::make_shared<FilteredConfigStore>(serviceManifestDir);
         m_defaultConfigStore = std::make_shared<DefaultConfigStore>(defaultServiceManifestDir);
-    } catch (ReturnCode err) {
-        log_error() << "Failed to initialize ConfigStore";
-        throw ReturnCode::FAILURE;
+    } catch (SoftwareContainerAgentError &error) {
+        std::string errorMessage = "Failed to initialize ConfigStore";
+        log_error() << errorMessage;
+        throw SoftwareContainerAgentError("SoftwareContainer Agent - " + errorMessage);
```
tobsan: Don't need to mention that this is the agent.

agent/src/softwarecontaineragent.h
```diff
@@ -47,6 +48,66 @@ namespace softwarecontainer {
...
+    {
+    }
+
+    virtual const char *what() const throw()
+    {
+        return m_message.c_str();
+    }
+};
+
+class WorkspaceError : public SoftwareContainerAgentError
```
tobsan: This error is not used anywhere yet. Also, to me it makes little sense to have this inherit the agent errors, since the workspace is part of the library.

agent/src/softwarecontaineragent.h
```diff
@@ -47,6 +48,66 @@ namespace softwarecontainer {
...
+    }
+
+protected:
+    std::string m_message;
+};
+
+class PreloadError : public SoftwareContainerAgentError
+{
+public:
+    PreloadError():
```
tobsan: This error is not currently used.

agent/unit-test/configstore_unittest.cpp
```diff
@@ -73,14 +73,14 @@ TEST_F(ConfigStoreTest, constructorFileOk2) {
  * should throw an exception of type ReturnCode.
  */
 TEST_F(ConfigStoreTest, constructorEvilFile) {
-    ASSERT_THROW(BaseConfigStore(testDataDir + evilManifest), ReturnCode);
+    ASSERT_THROW(BaseConfigStore(testDataDir + evilManifest), ConfigStoreError);
```
tobsan: This is much nicer than the old code!

tobsan (changes requested):
No real logical changes, mostly messages and some potential to remove code!

agent/src/capability/baseconfigstore.cpp
breakreturn: I guess this comment is about the message repeating the type of error? In this case it's repeated since it's in the errorMessage as well as the prefixed string, but I think it has its uses to repeat the info that the type would imply. If we catch the error on one of its base types it would be good to be able to generate useful log messages from it without knowing the type.

agent/src/capability/baseconfigstore.cpp
```diff
@@ -23,173 +23,201 @@ namespace softwarecontainer {
...
-        return ReturnCode::FAILURE;
+        errorMessage = "Could not parse Service Manifest";
+        log_error() << errorMessage << ": " << filePath << ":" << error.line <<" : " << error.text;
+        throw ServiceManifestPathError(errorMessage);
     } else if (!json_is_object(fileroot)) {
-        log_error() << "Configuration is not a json object: \n"
-                    << filePath << ":" << error.line <<" : " << error.text;
-        return ReturnCode::FAILURE;
+        errorMessage = "Configuration is not a json object";
+        log_error() << errorMessage << ": " << filePath << ":" << error.line <<" : " << error.text;
```
breakreturn: This info could be part of the errorMessage so it's passed along with the exception as well.

agent/src/capability/baseconfigstore.cpp
```diff
@@ -23,173 +23,201 @@ namespace softwarecontainer {
...
+        throw ServiceManifestPathError("Service Manifest parse error - " + errorMessage);
     }
 
     json_t *capabilities = json_object_get(fileroot, "capabilities");
     if (nullptr == capabilities) {
-        log_error() << "Could not parse Capabilities";
-        return ReturnCode::FAILURE;
+        errorMessage = "Could not parse Capabilities";
```
breakreturn: I think this should say "could not parse/find expected 'capabilities' key" or something like that, this message makes it unnecessarily vague.

agent/src/capability/baseconfigstore.cpp
```diff
@@ -23,173 +23,201 @@ namespace softwarecontainer {
...
     size_t i;
     json_t *gateway;
+    std::string errorMessage;
 
     GatewayConfiguration gwConf;
     json_array_foreach(gateways, i, gateway) {
         if (!json_is_object(gateway)) {
-            log_error() << "Gateway is not a json object";
-            return ReturnCode::FAILURE;
+            errorMessage = "Gateway is not a json object";
```
breakreturn: I think this message could be more to the point, e.g what has gone wrong. We should not use the word "Gateway" like this, since it's a very domain specific thing. In this case, we happen to have a key with the same word, but it's the "gateway" json key, not a "Gateway". I'd like to reduce the possibility of confusion here.

agent/src/capability/baseconfigstore.h
```diff
@@ -91,7 +92,7 @@ class BaseConfigStore
      * @return ReturnCode::SUCCESS if successful
      * @return ReturnCode::FAILURE if parsing of gateway configurations is unsuccessful
```
breakreturn: All comments need to be updated, remove return values and add @throws

agent/src/capability/baseconfigstore.cpp
tobsan: Good point, but I only think that applies if that is how we are going to use it, on a case-by-case basis.

agent/src/capability/baseconfigstore.cpp
breakreturn: I don't really understand the thinking behind the error types. It should be commented or documented what kinds of errors happen when and what they mean. For example readCapsFromFile will throw any of the two exceptions it can throw and in some code paths it is just propagated, and in some it is caught and made into a specific exception.

agent/src/softwarecontaineragent.cpp
```diff
@@ -85,16 +87,17 @@ SoftwareContainerAgent::SoftwareContainerAgent(
...
+        throw;
     }
 
     try {
         m_filteredConfigStore = std::make_shared<FilteredConfigStore>(serviceManifestDir);
         m_defaultConfigStore = std::make_shared<DefaultConfigStore>(defaultServiceManifestDir);
-    } catch (ReturnCode err) {
-        log_error() << "Failed to initialize ConfigStore";
-        throw ReturnCode::FAILURE;
+    } catch (SoftwareContainerAgentError &error) {
```
breakreturn: What's the reason to not catch the config store "base" exception? This looks like the config stores might throw an agent error. Not sure they even can.

agent/unit-test/configstore_unittest.cpp
```diff
@@ -73,14 +73,14 @@ TEST_F(ConfigStoreTest, constructorFileOk2) {
...
 }
 
 /* Constructing a BaseConfigStore with a directory path,
  * that contains a file that can't be parsed results in an exception.
  */
 TEST_F(ConfigStoreTest, constructorDir) {
-    ASSERT_THROW(BaseConfigStore(testDataDir + ""), ReturnCode);
+    ASSERT_THROW(BaseConfigStore(testDataDir + ""), ConfigStoreError);
 }
```
breakreturn, line 85: Comments need to be updated throughout this file.

agent/src/capability/baseconfigstore.cpp
thenor: tobsan: Previous reasoning was that if you have an entire directory of files, then not being able to parse one of them (or all of them) is only a warning. Maybe we want to reconsider?
breakreturn: When I read this I realize that "path" is not logical when the actual error is "parse".
Do you mean it should be commented here, or in the h file?

agent/src/capability/baseconfigstore.cpp
thenor: I also see the danger of using "path" and "parse" very similar words. Here I have used the "path" exception, but actually refer to it as a "parse" error.

breakreturn (changes requested):
I'm missing an overall understanding of what kind of exceptions are thrown when, why they are sometimes caught and turned into another type and when not. After some comments are added that describes what errors the interface throws in each method, it will be easier to review the logic and tests.

agent/src/capability/baseconfigstore.cpp
breakreturn: So I don't think we should repeat the same string, just to be clear. But I don't think we should rely on the type of the error to generate good logging. The type is good for logic and flow control etc., but it's not useful for logging IMHO. At this level, it's not possible to know how the exception will be used by users of the interface, so assuming that the user might or might not catch on specific types will be an easier assumption to maintain I think.

agent/src/capability/baseconfigstore.cpp
breakreturn: Maybe commented in both places. The interface should state what errors might happen, and here it might be good with a more internal note about why the error is turned into something else.

thenor: I will add the file path to those places where there is more information available.

thenor: I will remove this specific throw of exception. I judge that this is not a fatal error, hence the "warning" for log, so with that reasoning it shouldn't throw an exception. But we might want to discuss in detail which situations are severe enough to be error and throw an error.

thenor: I will do the opposite of my last comment. It seems our earlier reasoning made this an error since we would return FAILURE before, so I will change it to log_error and keep the throw.

agent/src/capability/baseconfigstore.cpp
thenor: I didn't see this comment. I will fix this.

agent/src/softwarecontaineragent.h
thenor: I will remove this. Preloading will be removed, and any situation where this could have been used will soon be obsolete. Also I feel that in similar situations I could instead use the SCAgentError.

agent/src/softwarecontaineragent.h
thenor: I will instead use the SCAgentError.

agent/src/capability/baseconfigstore.cpp
```diff
@@ -23,173 +23,190 @@ namespace softwarecontainer {
...
+            errorMessage = "Could not read Gateways";
+            log_error() << errorMessage;
+            throw CapabilityParseError(errorMessage);
         }
         if (!json_is_array(gateways)) {
-            log_error() << "Gateways is not an array";
-            return ReturnCode::FAILURE;
+            errorMessage = "Gateways is not an array";
+            log_error() << errorMessage;
+            throw CapabilityParseError(errorMessage);
```
tobsan, line 146: You could consider making log_error() + throw SomeError into a function to avoid duplication everywhere. `logAndThrow()` perhaps?

agent/src/main.cpp
```diff
@@ -250,8 +250,10 @@ int main(int argc, char **argv)
     } catch (ConfigError &error) {
         log_error() << "Could not load configuration";
         return 1;
-    } catch (ReturnCode failure) {
-        log_error() << "Agent initialization failed";
+    } catch (SoftwareContainerError &error) {
+        // Currently only case where any other exception can occur is during Agent init
+        std::string errorMessage = "Agent initialization failed: " + std::string(error.what());
```
tobsan: In this case no need to store the errorMessage since we can't propagate anywhere.

tobsan (approved):
Looks good. Only some minor issues for me.

agent/src/capability/baseconfigstore.cpp
```diff
@@ -23,173 +23,190 @@ namespace softwarecontainer {
...
         std::string filePath = dirPath + file;
 
-        if (isError(readCapsFromFile(filePath))) {
-            log_warning() << "Could not parse a file in directory: " << filePath;
-            return ReturnCode::FAILURE;
+        try {
+            readCapsFromFile(filePath);
+        } catch(ConfigStoreError &error) {
+            log_error() << "Could not parse a file in directory";
+            throw;
```
breakreturn: Why do we catch and re-throw here? Wouldn't it be enough to log what happened where the error occurred?

agent/src/capability/baseconfigstore.h
```diff
@@ -57,41 +62,40 @@ class BaseConfigStore
...
     /**
      * @brief Reads a Service Manifest file, of type json, and adds the
      * capabilities to the internal storage.
      * If a capability with the same name already exists in the storage
      * the gateways will be appended to the previously stored gatway list.
      *
-     * @return ReturnCode::SUCCESS if successful
-     * @return ReturnCode::FAILURE if parsing of file is unsuccessful
+     * @throws ServiceManifestParseError if parsing of the file is unsuccessful
+     * @throws CapabilityParseError if parsing of one or more Capabilities in file
```
breakreturn, line 77: If we don't re-throw from underlying methods, we don't need to state here that we throw both these exceptions.

agent/src/main.cpp
```diff
@@ -250,8 +250,10 @@ int main(int argc, char **argv)
     } catch (ConfigError &error) {
         log_error() << "Could not load configuration";
         return 1;
-    } catch (ReturnCode failure) {
-        log_error() << "Agent initialization failed";
+    } catch (SoftwareContainerError &error) {
+        // Currently only case where any other exception can occur is during Agent init
```
breakreturn: I don't understand this comment. Is it a comment about the fact that we are catching an exception that is not config related?

breakreturn (changes requested):
ConfigError should inherit SoftwareContainerError as well, but it seems to be missing here.

Overall I'd like us to stop logging and commenting about gateway config and capabilities keys etc. as if it was something else. Saying "Capabilities" and "Gateways" sounds like it might be something other that what we are actually meaning. We seem to often mean the name of the json key for example.

agent/src/capability/baseconfigstore.h
thenor: Ok.

agent/src/main.cpp
thenor: Maybe it is an unnecessary comment, but since I catch on the top level on a very large block of code (which does more that just initialize the Agent) it felt strange to me add the text "agent init failed". But maybe I don't even need that text? Not sure.

agent/src/main.cpp
thenor: Oh yes.

agent/src/capability/baseconfigstore.cpp
thenor: This is actually quite difficult. I understand why you want to change it. But what do you call the object(s) in the Service Manifest? Tag/key/...? "The \"gateway\" object is not a json object"?

agent/src/capability/baseconfigstore.cpp
thenor: Sounds like a good idea, but to get this PR in I think it would be a good idea to make a new PR for that. Do you consent?

agent/src/capability/baseconfigstore.cpp
thenor: I think it will. I will remove the re-throw.

agent/src/capability/baseconfigstore.h
thenor: In the readCapsFromFile it should be correct, but it was not correct for readCapsFromDir.

agent/src/capability/baseconfigstore.cpp
breakreturn: Yeah something like that looks more correct I think.

breakreturn (approved):

PR #286, opened 2017-01-17
Use exceptions to communicate failure over dbus

DunderRoffe:
Remove bool success returns in DBus API

Add support for converting SoftwareContainerError in agent to
`org.freedesktop.DBus.Error.Failed` response on DBus API.

Refactor servicetests to handle DBusException instead of
checking success value.

Refactor agent componenttests to handle SoftwareContainerError
instead of checking success value.


agent/component-test/softwarecontaineragent_unittest.cpp
```diff
@@ -111,19 +112,15 @@ class SoftwareContainerAgentTest: public ::testing::Test
 };
 
 TEST_F(SoftwareContainerAgentTest, CreatAndCheckContainer) {
-    ContainerID id;
-    bool success = sca->createContainer(valid_config, id);
-    ASSERT_TRUE(success);
-    ASSERT_TRUE(sca->getContainer(id) != nullptr);
+    ContainerID id = sca->createContainer(valid_config);
```
kursatkobya: Maybe using like `ASSERT_NO_THROW(id = sca->createContainer(valid_config));` would be better

agent/component-test/softwarecontaineragent_unittest.cpp
```diff
@@ -111,19 +112,15 @@ class SoftwareContainerAgentTest: public ::testing::Test
...
+    ASSERT_NO_THROW(sca->getContainer(id));
 }
 
 TEST_F(SoftwareContainerAgentTest, DeleteContainer) {
-    ContainerID id;
-    bool success = sca->createContainer(valid_config, id);
-    ASSERT_TRUE(success);
+    ContainerID id = sca->createContainer(valid_config);
```
kursatkobya: `ASSERT_NO_THROW(id = sca->createContainer(valid_config));` could be better

agent/component-test/softwarecontaineragent_unittest.cpp
```diff
@@ -140,119 +137,106 @@ TEST_F(SoftwareContainerAgentTest, CreateContainerWithConf) {
...
+    ASSERT_NO_THROW(sca->resumeContainer(id));
 }
 
 // Freeze an already frozen container
 TEST_F(SoftwareContainerAgentTest, FreezeFrozenContainer) {
-    ContainerID id;
-    bool success = sca->createContainer(valid_config, id);
-    ASSERT_TRUE(success);
-    ASSERT_TRUE(sca->getContainer(id) != nullptr);
+    ContainerID id = sca->createContainer(valid_config);
```
kursatkobya: using `ASSERT_NO_THROW` would be beter

kursatkobya: here as well

agent/component-test/softwarecontaineragent_unittest.cpp
```diff
@@ -140,119 +137,106 @@ TEST_F(SoftwareContainerAgentTest, CreateContainerWithConf) {
...
+    ASSERT_NO_THROW(sca->suspendContainer(id));
 }
 
 // Freeze an already frozen container, and then resume it
 TEST_F(SoftwareContainerAgentTest, DoubleFreezeContainerAndThaw) {
-    ContainerID id;
-    bool success = sca->createContainer(valid_config, id);
-    ASSERT_TRUE(success);
-    ASSERT_TRUE(sca->getContainer(id));
+    ContainerID id = sca->createContainer(valid_config);
```
kursatkobya: and here

agent/component-test/softwarecontaineragent_unittest.cpp
```diff
@@ -140,119 +137,106 @@ TEST_F(SoftwareContainerAgentTest, CreateContainerWithConf) {
...
+    ASSERT_NO_THROW(sca->resumeContainer(id));
 }
 
 // Double suspend and then double resume
 TEST_F(SoftwareContainerAgentTest, DoubleFreezeAndDoubleThawContainer) {
-    ContainerID id;
-    bool success = sca->createContainer(valid_config, id);
-    ASSERT_TRUE(success);
-    ASSERT_TRUE(sca->getContainer(id) != nullptr);
+    ContainerID id = sca->createContainer(valid_config);
```
kursatkobya: here

agent/component-test/softwarecontaineragent_unittest.cpp
```diff
@@ -140,119 +137,106 @@ TEST_F(SoftwareContainerAgentTest, CreateContainerWithConf) {
...
+    ASSERT_NO_THROW(sca->resumeContainer(id));
 }
 
 // Make sure you can still shutdown a frozen container
 TEST_F(SoftwareContainerAgentTest, ShutdownFrozenContainer) {
-    ContainerID id;
-    bool success = sca->createContainer(valid_config, id);
-    ASSERT_TRUE(success);
-    ASSERT_TRUE(sca->getContainer(id) != nullptr);
+    ContainerID id = sca->createContainer(valid_config);
```
kursatkobya: h

agent/src/softwarecontaineragent.cpp
```diff
@@ -119,94 +120,113 @@ bool SoftwareContainerAgent::triggerPreload()
...
+    }
+    
+    if (containerID < 0) {
+        std::string errorMessage("Invalid Container ID: "
+                                + std::to_string(containerID)
+                                + ". ID can not be negative");
+        log_error() << errorMessage;
+        throw SoftwareContainerError(errorMessage);
+    }
+    if (1 > m_containers.count(containerID)) {
```
kursatkobya, line 137: maybe it could be better to use i`f (!m_containers.count(containerID)) ` 
since it can be only 1 or 0, < operator doesn't seem relevant to me

agent/src/softwarecontaineragent.cpp
```diff
@@ -119,94 +120,113 @@ bool SoftwareContainerAgent::triggerPreload()
...
 {
-    return (containerID < INT32_MAX)
-        && (containerID >= 0)
-        && (1 == m_containers.count(containerID));
+    if (containerID >= INT32_MAX) {
+        std::string errorMessage("Invalid Container ID: "
+                                + std::to_string(containerID)
+                                + ". ID can not be greater than INT32");
+        log_error() << errorMessage;
+        throw SoftwareContainerError(errorMessage);
```
kursatkobya, line 127: I think the error message is going to be written to output at the place SoftwareContainerError is caught. So that can create duplicate error messages.  If this is true maybe it would be sensible to remover logs before throwing errors.

agent/src/softwarecontaineragent.cpp
DunderRoffe: I do not like using `0` as `false` since they are fundamentally different types. Could go for `0 == m_containers.count(containerID)` but then negative numbers would be allowed. Although, negative numbers are not possible anyway but it still feels nice :)

kursatkobya: `(0 == m_containers.count(containerID))` make sense as well :)

agent/src/softwarecontaineragent.cpp
DunderRoffe: I think we should try to log as quickly as possible when something happens to lessen the time spent debugging. You could argue that this can be done as well by tracing the exception, but that is not how we handle exceptions AFAIK.

agent/src/softwarecontaineragent.cpp
```diff
@@ -316,106 +338,88 @@ bool SoftwareContainerAgent::execute(ContainerID containerID,
...
-bool SoftwareContainerAgent::shutdownContainer(ContainerID containerID)
+void SoftwareContainerAgent::shutdownContainer(ContainerID containerID)
 {
     profilefunction("shutdownContainerFunction");
     if (!m_shutdownContainers) {
-        log_info() << "Not shutting down containers, so not shutting down " << containerID;
-        return false;
+        std::string errorMessage("Not shutting down containers, so not shutting down ID: " + std::to_string(containerID));
+        log_error() << errorMessage;
+        throw SoftwareContainerError(errorMessage);
```
tobsan: This should not be an error imo, this is just a fact. Just returning would suffice here.

tobsan (changes requested):
Looks good, only had one comment.

tobsan (approved):

PR #287, opened 2017-01-17
Vagrantfile: use icecc + bump cookbook + various

tobsan:
To use icecc we need to be available on the same network, hence
the network setup change.

Since we never use the VAGRANT_CPUS setting, I removed that too.

Also, instead of taking time to select the "best" mirror, I've
changed so that we only use the httpredir mirror, which more or
less does the same thing. So now we just do an apt-get update

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -24,14 +24,15 @@ require './cookbook/system-config/vagrant-reload.rb'
 
 Vagrant.configure(2) do |config|
     config.vm.box = "debian/contrib-jessie64"
+
+    # Common names for network adapters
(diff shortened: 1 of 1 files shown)


DunderRoffe (changes requested):
Looks nice, but before merging it I think it would be nice to verify that this works outside the office.

tobsan:
This sets no IP addresses or hostnames, and will always fall back to use whatever CC is set to if there are no servers available. But yes, it would be nice to just test it once outside the pcore offices.

DunderRoffe (approved):

tobsan:
@DunderRoffe Just tried running without the machine being on the same network. It then can't use the icecream network (obviously), but works fine otherwise.

PR #288, opened 2017-01-18
Renamed test file to reflect test level

breakreturn:
The test file for the agent component test should not have "unit test" in its name.

Also removed unused code in tests and updated values to be the same as elsewhere in the code.


tobsan (approved):

DunderRoffe (approved):
Merging if CI succeeds

kursatkobya (approved):

PR #290, opened 2017-01-18
Update CGroups Gateway to consider K,M and G symbols

kursatkobya:
Add support for memory.memsw.limit_in_bytes
Apply exceptions
Expand unit tests regarding these
Expand service tests regarding these
Update documentation

### DIFF
--- a/libsoftwarecontainer/src/gateway/cgroups/cgroupsgateway.cpp
+++ b/libsoftwarecontainer/src/gateway/cgroups/cgroupsgateway.cpp
@@ -32,7 +32,9 @@ CgroupsGateway::CgroupsGateway()
 
 ReturnCode CgroupsGateway::readConfigElement(const json_t *element)
 {
-    if (isError(m_parser.parseCGroupsGatewayConfiguration(element))) {
+    try {
+        m_parser.parseCGroupsGatewayConfiguration(element);
+    } catch (CgroupsGatewayError &e) {
         log_error() << "Could not parse CGroups configuration element";
         return ReturnCode::FAILURE;
     }
--- a/libsoftwarecontainer/src/gateway/cgroups/cgroupsparser.cpp
+++ b/libsoftwarecontainer/src/gateway/cgroups/cgroupsparser.cpp
@@ -18,6 +18,7 @@
  */
 
 #include "cgroupsparser.h"
+#include "cgroupsgateway.h"
 #include "jsonparser.h"
 
 namespace softwarecontainer {
@@ -27,25 +28,79 @@ CGroupsParser::CGroupsParser()
 {
 }
 
-ReturnCode CGroupsParser::parseCGroupsGatewayConfiguration(const json_t *element)
+std::string CGroupsParser::convertToBytes(const std::string settingValue, const int multiply)
+{
+    char *end;
+    long long limit_in_bytes = std::strtoll(settingValue.c_str(), &end, 10);
+    if (limit_in_bytes ==  std::numeric_limits<long long>::max() ||
+        limit_in_bytes ==  std::numeric_limits<long long>::min()) {
+        if (errno == ERANGE) {
+            std::string errMessage = settingValue + " is out of range";
+            log_error() << errMessage;
+            throw LimitRangeError(errMessage);
+        }
+    }
+
+    if (limit_in_bytes > llrint(std::numeric_limits<long long>::max() / multiply)) {
+        std::string errMessage = settingValue + " * " + std::to_string(multiply) + " is out of range";
+        log_error() << errMessage;
+        throw LimitRangeError(errMessage);
+    }
+
+    limit_in_bytes *= multiply;
+    return std::to_string(limit_in_bytes);
+}
+
+std::string CGroupsParser::suffixCorrection(const std::string settingValue)
+{
+    auto suffix = settingValue.back();
+    std::string removeUnit = settingValue;
+    removeUnit.pop_back();
+
+    if (std::isdigit(suffix)) {
+        //the value is digit already no need to do anything
+        return settingValue;
+    } else if ('K' == suffix || 'k' == suffix) {
+        //the value is Kilobyte unit convert it to byte
+        int KB = 1024;
+        return convertToBytes(removeUnit, KB);
+    } else if ('M' == suffix || 'm' == suffix) {
+        //the value is Megabyte unit convert it to byte
+        int MB = 1024 * 1024;
+        return convertToBytes(removeUnit, MB);
+    } else if ('G' == suffix || 'g' == suffix) {
+        //the value is Gigabyte unit convert it to byte
+        int GB = 1024 * 1024 * 1024;
+        return convertToBytes(removeUnit, GB);
+    }
+    std::string errMessage = "Bad suffix on setting value " + settingValue;
+    log_error() << errMessage;
+    throw BadSuffixError(errMessage);
+}
+
+void CGroupsParser::parseCGroupsGatewayConfiguration(const json_t *element)
 {
     std::string settingKey;
     std::string settingValue;
 
     if (!JSONParser::read(element, "setting", settingKey)) {
-        log_error() << "Key \"setting\" either not a string or not in json configuration";
-        return ReturnCode::FAILURE;
+        std::string errMessage = "Key \"setting\" either not a string or not in json configuration";
+        log_error() << errMessage;
+        throw JSonError(errMessage);
     }
 
     if (!JSONParser::read(element, "value", settingValue)) {
-        log_error() << "Key \"value\" either not a string or not in json configuration";
-        return ReturnCode::FAILURE;
+        std::string errMessage = "Key \"value\" either not a string or not in json configuration";
+        log_error() << errMessage;
+        throw JSonError(errMessage);
     }
 
-    if ("memory.limit_in_bytes" == settingKey) {
+    if (("memory.limit_in_bytes" == settingKey) || ("memory.memsw.limit_in_bytes" == settingKey)) {
+        settingValue = suffixCorrection(settingValue);
+
         if (m_settings.count(settingKey)) {
             // if the new value is greater than old one set is as the value
-            if (std::stoi(m_settings[settingKey]) < std::stoi(settingValue)) {
+            if (std::stoll(m_settings[settingKey]) < std::stoll(settingValue)) {
                 m_settings[settingKey] = settingValue;
             }
         } else {
@@ -55,8 +110,6 @@ ReturnCode CGroupsParser::parseCGroupsGatewayConfiguration(const json_t *element
         log_warning() << settingKey << " is not supported by CGroups Gateway" ;
         m_settings[settingKey] = settingValue;
     }
-
-    return ReturnCode::SUCCESS;
 }
 
 const std::map<std::string, std::string> &CGroupsParser::getSettings()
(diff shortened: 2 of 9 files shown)


libsoftwarecontainer/src/gateway/cgroups/cgroupsparser.cpp
```diff
@@ -27,36 +28,111 @@ CGroupsParser::CGroupsParser()
...
-            if (std::stoi(m_settings[settingKey]) < std::stoi(settingValue)) {
+            if (std::stoll(m_settings[settingKey]) < std::stoll(settingValue)) {
                 m_settings[settingKey] = settingValue;
             }
         } else {
+            try {
+                settingValue = suffixCorrection(settingValue);
+            } catch (LimitRangeError &error) {
+                log_error() << settingValue << " is not appropriate value for " << settingKey;
+                throw LimitRangeError();
```
DunderRoffe: Seems like this code is identical to the code in the other case of the if statement or at least very similar. This can be cleaned up

DunderRoffe (changes requested):

libsoftwarecontainer/src/gateway/cgroups/cgroupsparser.cpp
kursatkobya: Done

libsoftwarecontainer/src/gateway/cgroups/cgroupsgateway.cpp
```diff
@@ -32,7 +32,9 @@ CgroupsGateway::CgroupsGateway()
 
 ReturnCode CgroupsGateway::readConfigElement(const json_t *element)
 {
-    if (isError(m_parser.parseCGroupsGatewayConfiguration(element))) {
+    try{
```
tobsan: Missing space before {

libsoftwarecontainer/src/gateway/cgroups/cgroupsparser.cpp
```diff
@@ -27,25 +28,94 @@ CGroupsParser::CGroupsParser()
 {
 }
 
-ReturnCode CGroupsParser::parseCGroupsGatewayConfiguration(const json_t *element)
+std::string CGroupsParser::byteRegulation(const std::string settingValue,
+                                                        const int multiply)
```
tobsan: Naming is confusing and indentation is bad.

libsoftwarecontainer/src/gateway/cgroups/cgroupsparser.cpp
```diff
@@ -27,25 +28,94 @@ CGroupsParser::CGroupsParser()
...
+{
+    char *end;
+    long long limit_in_bytes = std::strtoll(settingValue.c_str(), &end, 10);
+    if (errno == ERANGE){
```
tobsan: Missing space before {. Also, shouldn't you first check for the return value before checking errno?

libsoftwarecontainer/src/gateway/cgroups/cgroupsparser.cpp
```diff
@@ -27,25 +28,94 @@ CGroupsParser::CGroupsParser()
...
+        log_error() << settingValue << " is out of range";
+        throw LimitRangeError();
+    }
+
+    if (limit_in_bytes * multiply > std::numeric_limits<long long>::max()) {
```
tobsan: Wouldn't this cause overflow and then the comparison would be false anyway? A better way would probably be to check if max() of long long divided by multiply is less than limit_in_bytes.

libsoftwarecontainer/src/gateway/cgroups/cgroupsparser.cpp
```diff
@@ -27,25 +28,94 @@ CGroupsParser::CGroupsParser()
...
+        //the value is digit already no need to do anything
+        return settingValue;
+    } else if ('K' == suffix || 'k' == suffix) {
+        //the value is Kilobyte unit convert it to byte
+        int KB = 1024;
+
+        try {
+            return byteRegulation(removeUnit, KB);
+        } catch (LimitRangeError &error) {
+            throw LimitRangeError();
```
tobsan: No need to catch if you just want to throw the same thing.

libsoftwarecontainer/src/gateway/cgroups/cgroupsparser.cpp
```diff
@@ -27,25 +28,94 @@ CGroupsParser::CGroupsParser()
...
+        }
+    } else if ('M' == suffix || 'm' == suffix) {
+        //the value is Megabyte unit convert it to byte
+        int MB = 1024 * 1024;
+
+        try {
+            return byteRegulation(removeUnit, MB);
+        } catch (LimitRangeError &error) {
+            throw LimitRangeError();
```
tobsan: Also here, why catch just to throw?

libsoftwarecontainer/src/gateway/cgroups/cgroupsparser.cpp
```diff
@@ -27,25 +28,94 @@ CGroupsParser::CGroupsParser()
...
+        }
+    } else if ('G' == suffix || 'g' == suffix) {
+        //the value is Gigabyte unit convert it to byte
+        int GB = 1024 * 1024 * 1024;
+
+        try {
+            return byteRegulation(removeUnit, GB);
+        } catch (LimitRangeError &error) {
+            throw LimitRangeError();
```
tobsan: No need to catch just to throw.

libsoftwarecontainer/src/gateway/cgroups/cgroupsparser.cpp
```diff
@@ -27,25 +28,94 @@ CGroupsParser::CGroupsParser()
...
+
+void CGroupsParser::parseCGroupsGatewayConfiguration(const json_t *element)
 {
     std::string settingKey;
     std::string settingValue;
 
     if (!JSONParser::read(element, "setting", settingKey)) {
         log_error() << "Key \"setting\" either not a string or not in json configuration";
-        return ReturnCode::FAILURE;
+        throw JSonError();
```
tobsan: An error message in the exception would be good.

libsoftwarecontainer/src/gateway/cgroups/cgroupsparser.cpp
```diff
@@ -27,25 +28,94 @@ CGroupsParser::CGroupsParser()
...
     }
 
     if (!JSONParser::read(element, "value", settingValue)) {
         log_error() << "Key \"value\" either not a string or not in json configuration";
-        return ReturnCode::FAILURE;
+        throw JSonError();
```
tobsan: Use an error message here. Same as used in the logging probably.

libsoftwarecontainer/src/gateway/cgroups/cgroupsparser.cpp
```diff
@@ -27,25 +28,94 @@ CGroupsParser::CGroupsParser()
...
     }
 
-    if ("memory.limit_in_bytes" == settingKey) {
+    if (("memory.limit_in_bytes" == settingKey) || ("memory.memsw.limit_in_bytes" == settingKey)) {
+        try {
+            settingValue = suffixCorrection(settingValue);
+        } catch (LimitRangeError &error) {
+            log_error() << settingValue << " is not appropriate value for " << settingKey;
+            throw LimitRangeError();
```
tobsan: Use a message here.

libsoftwarecontainer/src/gateway/cgroups/cgroupsparser.cpp
```diff
@@ -27,25 +28,94 @@ CGroupsParser::CGroupsParser()
...
+        } catch (BadSuffixError  &error) {
+            throw BadSuffixError();
```
tobsan: Use a message here (or don't catch at all, since you're just re-throwing the same type of exception)

libsoftwarecontainer/src/gateway/cgroups/cgroupsparser.h
```diff
@@ -40,6 +46,25 @@ class CGroupsParser
...
+
+    /*
+     * @brief Regulate the setting value due to its multiplier
+     *
+     * @param settingValue : decimal value without unit information
+     * @param multiply     : multiplication value either KB, MB or GB
+     *
+     * @return A string to regulated final value
+     */
+    std::string byteRegulation(const std::string settingValue, const int multiply);
```
tobsan: something like checkByteLimits or something along those lines would make more sense.

tobsan (changes requested):

libsoftwarecontainer/src/gateway/cgroups/cgroupsparser.cpp
kursatkobya: Yeah space will be there. 
It is not mandatory to check return value, if the converted value falls out of range of a range error occurs and  LLONG_MAX or LLONG_MIN is returned. checking whether it is set to LLONG_MAX is nto so different than checking errno in my opinion.

libsoftwarecontainer/src/gateway/cgroups/cgroupsparser.cpp
kursatkobya: I think, since i am not assigning it it won't create overflow. The other way you propose can create floating comparison which i didn't like that much.

DunderRoffe (approved):

libsoftwarecontainer/src/gateway/cgroups/cgroupsparser.h
kursatkobya: will use convertToBytes

libsoftwarecontainer/src/gateway/cgroups/cgroupsparser.cpp
```diff
@@ -27,25 +28,74 @@ CGroupsParser::CGroupsParser()
...
+        limit_in_bytes ==  std::numeric_limits<long long>::min()) {
+        if (errno == ERANGE) {
+            log_error() << settingValue << " is out of range";
+            throw LimitRangeError(settingValue + " is out of range");
+        }
+    }
+
+    if (limit_in_bytes > llrint(std::numeric_limits<long long>::max() / multiply)) {
+        log_error() << settingValue << " * " << std::to_string(multiply) <<" is out of range";
+        throw LimitRangeError();
```
tobsan: It would be nice to have an error message here also.

libsoftwarecontainer/src/gateway/cgroups/cgroupsparser.cpp
```diff
@@ -27,25 +28,74 @@ CGroupsParser::CGroupsParser()
...
+        //the value is Megabyte unit convert it to byte
+        int MB = 1024 * 1024;
+        return convertToBytes(removeUnit, MB);
+    } else if ('G' == suffix || 'g' == suffix) {
+        //the value is Gigabyte unit convert it to byte
+        int GB = 1024 * 1024 * 1024;
+        return convertToBytes(removeUnit, GB);
+    }
+    log_error() << "Bad suffix on setting value " << settingValue;
+    throw BadSuffixError("Bad suffix on setting value " + settingValue);
```
tobsan: Consider creating a variable holding the message, since you're both logging and throwing the same message. That goes for all places in the code.

tobsan (changes requested):

tobsan (approved):

PR #295, opened 2017-01-19
Use Vagrant reload plugin to only reload if needed

tobsan:
We only want to reload if the grub config changed, so we can now
pass a condition to the reload plugin.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -20,7 +20,7 @@
 # For further information see LICENSE
 #
 
-require './cookbook/system-config/vagrant-reload.rb'
+require './cookbook/host-system-config/vagrant-reload/plugin.rb'
 
 Vagrant.configure(2) do |config|
     config.vm.box = "debian/contrib-jessie64"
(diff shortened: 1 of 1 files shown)


Vagrantfile
```diff
@@ -65,7 +65,13 @@ Vagrant.configure(2) do |config|
 
     # Change grub to support cgroups memory if necessary
     config.vm.provision "shell", path: "cookbook/system-config/grub-enable-cgroup-memory.sh"
-    config.vm.provision :reload
+    fileCheck = lambda {
+        if File.exists?("VAGRANT_NEEDS_RELOAD")
+            File.delete("VAGRANT_NEEDS_RELOAD")
+            return true
+        end
```
rpannek, line 74: Wouldn't it be nicer if you returned false instead of nil so it would return a boolean?

breakreturn (approved):

Vagrantfile
tobsan: Yeah, that would be reasonable. The plugin checks explicitly for true, so anything else is considered false, but it is still good practice to return false I think.

rpannek (approved):

kursatkobya (approved):

thenor (approved):

PR #296, opened 2017-01-19
Fix agent parser manifest directories bug

kursatkobya:


agent/src/main.cpp
```diff
@@ -172,6 +172,10 @@ int main(int argc, char **argv)
      */
     std::vector<StringConfig> stringConfigs = std::vector<StringConfig>();
     if (serviceManifestDir != ConfigDefinition::SERVICE_MANIFEST_DIR_INITIAL_VALUE) {
+        if ('/' != *serviceManifestDir.rbegin()) {
```
tobsan: Use `serviceManifestDir.back()` instead

tobsan (changes requested):

agent/src/main.cpp
tobsan: EDIT: My bad, this is a Glib string, not a std::string. Never mind!

tobsan (approved):

tobsan (approved):

agent/src/main.cpp
```diff
@@ -121,8 +121,8 @@ int main(int argc, char **argv)
     int userID = 0;
     bool keepContainersAlive = ConfigDefinition::KEEP_CONTAINERS_ALIVE_INITIAL_VALUE;
     int timeout = ConfigDefinition::SHUTDOWN_TIMEOUT_INITIAL_VALUE;
-    Glib::ustring serviceManifestDir = ConfigDefinition::SERVICE_MANIFEST_DIR_INITIAL_VALUE;
-    Glib::ustring defaultServiceManifestDir = ConfigDefinition::DEFAULT_SERVICE_MANIFEST_DIR_INITIAL_VALUE;
```
breakreturn, line 117: Could you change "configPath" as well so it's all std::string?

breakreturn (approved):

DunderRoffe (approved):

PR #298, opened 2017-01-20
Propagate configs from agent to SC

breakreturn:
SoftwareContainer now takes configs from the agent on creation.
The agent handles the static and dynamic configs and pass them
all to SC together.

This is a step towards removing the Workspace as well.

### DIFF
--- a/agent/src/softwarecontaineragent.cpp
+++ b/agent/src/softwarecontaineragent.cpp
@@ -22,16 +22,17 @@
 #include "softwarecontaineragent.h"
 #include "config/configerror.h"
 #include "config/configdefinition.h"
+#include "config/softwarecontainerconfig.h"
 #include "softwarecontainererror.h"
 
 namespace softwarecontainer {
 
 
-SoftwareContainerAgent::SoftwareContainerAgent(
-        Glib::RefPtr<Glib::MainContext> mainLoopContext,
-        std::shared_ptr<Config> config):
+SoftwareContainerAgent::SoftwareContainerAgent(Glib::RefPtr<Glib::MainContext> mainLoopContext,
+                                               std::shared_ptr<Config> config):
     m_mainLoopContext(mainLoopContext),
-    m_config(config)
+    m_config(config),
+    m_containerConfig(SoftwareContainerConfig())
 {
     m_containerIdPool.push_back(0);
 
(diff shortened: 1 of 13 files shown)


agent/src/softwarecontaineragent.cpp
```diff
@@ -192,17 +205,11 @@ ContainerID SoftwareContainerAgent::findSuitableId()
...
-    int netmaskBits = m_config->getIntValue(ConfigDefinition::SC_GROUP,
-                                            ConfigDefinition::SC_BRIDGE_NETMASK_BITS_KEY);
-
     log_debug() << "Created container with ID :" << containerID;
 
     auto container = SoftwareContainerPtr(new SoftwareContainer(m_softwarecontainerWorkspace,
                                                                 containerID,
-                                                                bridgeIp,
-                                                                netmaskBits));
+                                                                m_containerConfig));
```
tobsan: Shouldn't this be a deep copy of the config, rather than a shared pointer at this point? If I'm reading this correctly, this will mean that the following would be true:

1) Call create(enableWriteBuffer = true) = SC-0
2) Call create(enableWriteBuffer = false) = SC-1

Wouldn't the second call change the enableWriteBuffer setting in SC-0 to false as well? That is not desired behaviour I think, even if we only use those values in the create step.

tobsan (changes requested):

agent/src/softwarecontaineragent.cpp
breakreturn: Yes, this is not nice at all. I think it's better for the whole design to perhaps have two separate objects in the agent that each are responsible for static and dynamic configs respectively, and "re-pack" it into an object sent to SC on creation, and do all necessary copying at that point. Will rework this.

agent/src/softwarecontaineragent.cpp
```diff
@@ -22,17 +22,19 @@
...
-        Glib::RefPtr<Glib::MainContext> mainLoopContext,
-        std::shared_ptr<Config> config):
+SoftwareContainerAgent::SoftwareContainerAgent(Glib::RefPtr<Glib::MainContext> mainLoopContext,
+                                               std::shared_ptr<Config> config):
     m_mainLoopContext(mainLoopContext),
-    m_config(config)
+    m_config(config),
+    m_containerConfig(SoftwareContainerConfig())
 {
+    m_config = config;
```
tobsan: Do you really need this as well as the initializer list statement?

agent/src/softwarecontaineragent.cpp
```diff
@@ -135,6 +147,7 @@ void SoftwareContainerAgent::readConfigElement(const json_t *element)
     }
 
     m_softwarecontainerWorkspace->m_enableWriteBuffer = wo;
+    m_containerConfig.setEnableWriteBuffer(wo);
```
tobsan, line 149: This is theoretically open for race conditions. It would be nicer if this parse method returned a config object that the caller could pass on. But I'm not sure it is worth fixing.

tobsan (approved):

PR #299, opened 2017-01-20
Jenkinsfile: try to force shutdown of Vagrant

tobsan:

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -24,6 +24,10 @@ def runInVagrant = { String workspace, String command ->
     sh "cd ${workspace} && vagrant ssh -c '${command}'"
 }
 
+def shutdownVagrant = {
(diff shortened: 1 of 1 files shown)


tobsan:
I've had this run through Jenkins a couple of times now and it really seems to kill off the VMs properly. I haven't yet seen an occurence where the shutdown call fails, so the loop has never run as of yet.

kursatkobya (approved):

Jenkinsfile
```diff
@@ -139,8 +143,10 @@ node {
         currentBuild.result = "FAILURE"
     }
 
-    // Always try to shut down the machine
-    // Shutdown the machine
-    sh "cd ${workspace} && vagrant destroy -f || true"
+    finally {
+        // Always try to shut down the machine
+        // Shutdown the machine
```
breakreturn: The two comments are a bit redundant no?

thenor (approved):

PR #301, opened 2017-01-23
Move network settings to config file, update scripts

tobsan:
Instead of statically configuring via CMake, the scripts take their assumed settings as args instead.

The agent also creates a config object with all dynamically configurable values when creating a new SC instance.


agent/component-test/softwarecontaineragent_componenttest.cpp
```diff
@@ -115,6 +116,20 @@ class SoftwareContainerAgentTest: public ::testing::Test
     }
 };
 
+class ContainerConfigParserTest : public ::testing::Test
```
breakreturn: I think this should be moved to the unit-tests instead if possible.

agent/src/config/configdefinition.cpp
```diff
@@ -80,6 +84,43 @@ const std::string ConfigDefinition::SC_BRIDGE_NETMASK_BITS_KEY = "bridge-netmask
...
+                    ConfigDefinition::SC_CREATE_BRIDGE_KEY,
+                    ConfigType::Boolean,
+                    ConfigDefinition::convertDefineToFlag(
+                        SC_CREATE_BRIDGE_MANDATORY_FLAG /* set by cmake */)),
+    std::make_tuple(ConfigDefinition::SC_GROUP,
+                    ConfigDefinition::SC_BRIDGE_DEVICE_KEY,
+                    ConfigType::String,
+                    ConfigDefinition::convertDefineToFlag(
+                        SC_BRIDGE_DEVICE_MANDATORY_FLAG /* set by cmake */)),
+    std::make_tuple(ConfigDefinition::SC_GROUP,
```
breakreturn, line 98: Duplicate entry?

agent/src/containerconfigparser.h
```diff
@@ -0,0 +1,81 @@
...
+ */
+
+#pragma once
+
+#include "softwarecontainer-common.h"
+#include "softwarecontainererror.h"
+
+namespace softwarecontainer {
+
+class ContainerConfigParseError
```
breakreturn: Any reason this is not inheriting SoftwareContainerError?

agent/src/containerconfigparser.h
```diff
@@ -0,0 +1,81 @@
...
+
+    /**
+     * @brief Parse config needed for starting up the container in a correct manner.
+     *
+     * This reads a config string into json format, makes sure it is formatted
+     * correctly, and then calls readConfigElement for its elements.
+     *
+     * @param config a configuration string
+     */
+    ContainerOptions parseConfig(const std::string &config);
```
breakreturn: I think we should document "throws" on all public methods that can throw. I know we only want to document that on the method that actually throws, but I think that should be for private code. The user of the interface should be able to read this info on the methods part of the interface IMO.

agent/src/softwarecontaineragent.h
```diff
@@ -296,6 +266,11 @@ class SoftwareContainerAgent
      * be passed to SoftwareContainer as a unique_ptr.
      */
     SoftwareContainerConfig m_containerConfig;
+
+    /*
+     * Responsible for parsing any dynamic values given when creating a container
+     */
+    ContainerConfigParser m_containerParser;
```
breakreturn: I don't like this name very much. It can read a bit like "a parser for a container to use", or "parses a container". Maybe configParser or something like that?

agent/src/containerconfigparser.h
breakreturn: I think "parse" would be enough for this name, the context etc. should always make the "Config" part redundant. A matter of taste ofc, but for what it's worth.

agent/src/softwarecontaineragent.cpp
```diff
@@ -223,8 +173,9 @@ ContainerID SoftwareContainerAgent::createContainer(const std::string &config)
     profilefunction("createContainerFunction");
 
     try {
-        parseConfig(config);
-    } catch (SoftwareContainerError &err) {
+        ContainerConfigParser::ContainerOptions options = m_containerParser.parseConfig(config);
+        m_containerConfig.setEnableWriteBuffer(options.enableWriteBuffer);
+    } catch (ContainerConfigParseError &err) {
```
breakreturn: Since ContainerConfigParseError is internal to SC and should inherit SoftwareContainerError, and since we catch the outermost base class, I think we could skip catching and throwing here. If we inherit the base class and initialize it with the message already when creating the original exception, we would maintain the same log abilities.

libsoftwarecontainer/include/softwarecontainer.h
```diff
@@ -119,21 +117,22 @@ class SoftwareContainer :
     ReturnCode activateGateways();
     ReturnCode shutdownGateways();
 
-    std::shared_ptr<Workspace> m_workspace;
-    ContainerID m_containerID;
-    std::string m_bridgeIp;
-    int m_netmaskBits;
-
-    ObservableWritableProperty<ContainerState> m_containerState;
+    // The new and improved Workspace
```
breakreturn: This doesn't provide much for the understanding of the code as it stands now. Developers familiar with the now removed Workspace class will understand the reference, but others will be left with a mystery they will have to quest around in the git history for.

agent/CMakeLists.txt
```diff
@@ -57,8 +58,14 @@ install(TARGETS softwarecontainer-agent DESTINATION bin)
 # At this point, there should be no more definitions being added/changed so it's OK to set them
 set_all_deferred_config_definitions()
 
+# Network settings should only be in the config if the network gateway is enabled
+if(ENABLE_NETWORKGATEWAY)
+    configure_file(softwarecontainer-config.network.in softwarecontainer-config.network)
```
breakreturn, line 63: The config configuration (whoa!) has become so non-obvious that it might deserve some comment in the cmake file before configuring things? Like, "the network in-file is configured separately and later made part into the config in-file", or something like that.

breakreturn (changes requested):

agent/src/containerconfigparser.h
tobsan: No, my bad.

breakreturn (approved):

PR #305, opened 2017-01-24
Agent restructuring + introducing type for dynamic options

tobsan:
This also unifies the flow for create() so that we use one small method instead of three tiny ones. See the separate commits for details.


agent/component-test/CMakeLists.txt
```diff
@@ -40,6 +39,8 @@ set(SOFTWARECONTAINERAGENT_TEST_FILES
     ${SOFTWARECONTAINERAGENT_DIR}/src/config/mainconfigsource.cpp
     ${SOFTWARECONTAINERAGENT_DIR}/src/config/configitem.cpp
     ${SOFTWARECONTAINERAGENT_DIR}/src/config/configdefinition.cpp
+    ${SOFTWARECONTAINERAGENT_DIR}/src/containeroptions/dynamiccontaineroptions.cpp
+    ${SOFTWARECONTAINERAGENT_DIR}/src/containeroptions/containerconfigparser.cpp
```
breakreturn: Should this file (and the class in it) now actually be "containeroptionparser"?

agent/src/containeroptions/containerconfigparser.cpp
```diff
@@ -67,7 +70,7 @@ ContainerConfigParser::ContainerOptions ContainerConfigParser::parse(const std::
     size_t index;
     json_t *element;
 
-    ContainerOptions options;
+    DynamicContainerOptions options;
```
breakreturn: It's a bit unusual in this component to pass a reference to the data to modify like this. Maybe this could be a member on the class instead?

agent/src/containeroptions/dynamiccontaineroptions.h
```diff
@@ -0,0 +1,53 @@
...
+#include "config/softwarecontainerconfig.h"
+
+#include <memory>
+
+namespace softwarecontainer {
+
+class DynamicContainerOptions
+{
+public:
+    DynamicContainerOptions();
```
breakreturn: I think the most consistent style here would be to either define the default constructor and destructor and have them completely in the header, or skip both. Since we don't define any other constructor it should be generated for us.

agent/src/softwarecontaineragent.cpp
```diff
@@ -152,50 +154,31 @@ ContainerID SoftwareContainerAgent::findSuitableId()
...
 {
     profilepoint("createContainerStart");
     profilefunction("createContainerFunction");
 
-    ContainerConfigParser::ContainerOptions options = m_configParser.parse(config);
-    m_containerConfig.setEnableWriteBuffer(options.enableWriteBuffer);
-
-    std::pair<ContainerID, SoftwareContainerPtr> pair = std::move(getContainerPair());
+    // Set options for this container
+    DynamicContainerOptions options = m_configParser.parse(config);
```
breakreturn: This section of the code makes me feel the naming of the members (especially), and the parameter etc. could be more separated between "config" and "options". Maybe rename m_configParser?

breakreturn (changes requested):

agent/src/containeroptions/dynamiccontaineroptions.h
```diff
@@ -0,0 +1,51 @@
...
+
+#include <memory>
+
+namespace softwarecontainer {
+
+class DynamicContainerOptions
+{
+public:
+    /*
+     * @brief creates a softwarecontainerconfig with the dynamic options set
```
breakreturn: Should be a camelcased class name right?

agent/src/containeroptions/dynamiccontaineroptions.h
```diff
@@ -0,0 +1,51 @@
...
+     *
+     * @param conf a SoftwareContainerConfig with static settings
+     * @return a SoftwareContainerConfig pointer where the static options that this object
```
breakreturn: Either this comment is wrong or I am confused. Should it be "where the dynamic options that this object holds are set" ?

tobsan:
Updated in accordance with comments given.

breakreturn (approved):

PR #308, opened 2017-01-24
Use Glib::build_filename() for combining paths

tobsan:
This is less error-prone and ensures we avoid both missing
slashes and double slashes in path names.

Also some general minor cleanups of things that I came across while doing this.

### DIFF
--- a/agent/src/capability/baseconfigstore.h
+++ b/agent/src/capability/baseconfigstore.h
@@ -23,11 +23,13 @@
  */
 
 #pragma once
+
 #include "softwarecontainer-common.h"
 #include "gatewayconfig.h"
 #include "configstoreerror.h"
 
 #include <jsonparser.h>
+
 
 namespace softwarecontainer {
 
@@ -45,11 +47,9 @@ public:
      * or if the path is not allowed
      * @throws ServiceManifestParseError if parsing of the json file(s) fails
      * @throws CapabilityParseError if parsing of one or more Capabilities or Gateway
-     * Configurations in a file is unsuccessful
+     *         Configurations in a file is unsuccessful
      */
     BaseConfigStore(const std::string &filePath);
-
-    virtual ~BaseConfigStore();
 
 protected:
     std::map<std::string, GatewayConfiguration> m_capMap;
--- a/agent/unit-test/configstore_unittest.cpp
+++ b/agent/unit-test/configstore_unittest.cpp
@@ -21,6 +21,8 @@
 #include "capability/filteredconfigstore.h"
 #include "capability/defaultconfigstore.h"
 
+#include <glibmm.h>
+
 #include <gmock/gmock.h>
 #include <gtest/gtest.h>
 #include <unistd.h>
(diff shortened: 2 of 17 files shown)


agent/src/capability/baseconfigstore.h
```diff
@@ -45,12 +47,10 @@ class BaseConfigStore
      * or if the path is not allowed
      * @throws ServiceManifestParseError if parsing of the json file(s) fails
      * @throws CapabilityParseError if parsing of one or more Capabilities or Gateway
-     * Configurations in a file is unsuccessful
+     *         Configurations in a file is unsuccessful
```
kursatkobya, line 50: Why do we need indentation here ?

common/softwarecontainer-common.cpp
```diff
@@ -110,6 +111,16 @@ std::string parentPath(const std::string &path_)
...
 }
 
+std::string buildPath(const std::string &arg1, const std::string &arg2)
+{
+    return Glib::build_filename(arg1, arg2);
+}
+
+std::string buildPath(const std::string &arg1, const std::string &arg2, const std::string &arg3)
+{
+    return Glib::build_filename(arg1, arg2, arg3);
```
DunderRoffe: In order to lessen the work needed to swap Glib for some other lib in the future, it would be really neat if this was defined as a nested case of the one with two arguments.

agent/src/capability/baseconfigstore.h
tobsan: It looks better if you want to add a new line below @throws

common/softwarecontainer-common.h
```diff
@@ -146,4 +146,10 @@ ReturnCode writeToFile(const std::string &path, const std::string &content);
 ReturnCode readFromFile(const std::string &path, std::string &content);
 bool parseInt(const char *args, int *result);
 
+/*
+ * Create a path from parts
```
DunderRoffe: A bit vague would be nice with an example.

DunderRoffe (changes requested):

agent/src/capability/baseconfigstore.h
kursatkobya: ah it looks like tab more than empty line

tobsan: No, it just indents the line to be aligned with the above line.

agent/unit-test/configstore_unittest.cpp
```diff
@@ -58,29 +64,31 @@ TEST_F(ConfigStoreTest, constructorEmptyStrOk) {
...
-    ASSERT_NO_THROW(BaseConfigStore(testDataDir + manifestPath));
+    ASSERT_NO_THROW(BaseConfigStore(buildPath(testDataDir, manifestPath)));
 }
 
 /* Constructing a FilteredConfigStore with a directory path,
  * even if all files can not be parsed, should not throw an exception.
  */
 TEST_F(ConfigStoreTest, constructorDirNoJsonFiles) {
     // No Service Manifests, but ok dir
+    // TODO: Get rid of this + ""
```
breakreturn, line 82: So why do we have the '+' here?

common/recursivecopy.cpp
```diff
@@ -88,8 +90,7 @@ ReturnCode RecursiveCopy::copy(std::string src, std::string dst)
     return retval;
 }
 
-RecursiveCopy::RecursiveCopy() { }
-
-RecursiveCopy::~RecursiveCopy() { }
+RecursiveCopy::RecursiveCopy() {}
+RecursiveCopy::~RecursiveCopy() {}
```
breakreturn, line 94: I'm not sure what the design intention here is really, but are these needed?

breakreturn (changes requested):
Some minor comments and questions.

kursatkobya (approved):
looks good

agent/unit-test/configstore_unittest.cpp
tobsan: It was already there, it does not compile without it, and I did not want to spend time on that (because that isn't really part of this story).

common/recursivecopy.cpp
tobsan: Yes, they are needed because they are deleted in the header file.

breakreturn (approved):

DunderRoffe (approved):

PR #313, opened 2017-01-25
Sanatize the creation of temp files in tests

DunderRoffe:

### DIFF
--- a/libsoftwarecontainer/component-test/softwarecontainer_test.cpp
+++ b/libsoftwarecontainer/component-test/softwarecontainer_test.cpp
@@ -37,12 +37,12 @@ void SoftwareContainerTest::exit()
     m_ml->quit();
 }
 
-/* 
+/*
  * Set up the workspace with all the config values it needs.
- */ 
+ */
 std::unique_ptr<SoftwareContainerConfig> SoftwareContainerTest::createConfig()
 {
-    std::unique_ptr<SoftwareContainerConfig> config = 
+    std::unique_ptr<SoftwareContainerConfig> config =
         std::unique_ptr<SoftwareContainerConfig>(new SoftwareContainerConfig(
 #ifdef ENABLE_NETWORKGATEWAY
         // Setup network stuff
@@ -57,7 +57,7 @@ std::unique_ptr<SoftwareContainerConfig> SoftwareContainerTest::createConfig()
         SHARED_MOUNTS_DIR_TESTING, // Should be set be CMake
         1 // timeout
     ));
-    
+
     config->setEnableWriteBuffer(false);
     return config;
 }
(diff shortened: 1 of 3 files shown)


libsoftwarecontainer/component-test/softwarecontainer_test.cpp
```diff
@@ -78,29 +78,70 @@ void SoftwareContainerTest::TearDown()
...
+    }
+    dirsToRemove.clear();
+}
+
+std::string SoftwareContainerTest::createTempFile(const std::string &prefix)
+{
+    std::string strWithPrefix(prefix + "/tmp/SC-tmpFileXXXXXX");
+    size_t n = strWithPrefix.size() + 1;
+    char *fileTemplate = new char[n];
+    strncpy(fileTemplate, strWithPrefix.c_str(), n);
```
kursatkobya, line 104: I think this 4 line is repeated down as well maybe it worth to create a function for it.

libsoftwarecontainer/component-test/softwarecontainer_test.cpp
```diff
@@ -78,29 +78,70 @@ void SoftwareContainerTest::TearDown()
...
+    filesToRemove.push_back(filename);
+    return filename;
+}
+
+std::string SoftwareContainerTest::createTempDir(const std::string &prefix)
+{
+    std::string strWithPrefix(prefix + "/tmp/SC-tmpDirXXXXXX");
+    size_t n = strWithPrefix.size() + 1;
+    char *dirTemplate = new char[n];
+    strncpy(dirTemplate, strWithPrefix.c_str(), n);
```
kursatkobya, line 127: also copy function can be used here 
`strWithPrefix.copy(dirTemplate, strWithPrefix.size())`

libsoftwarecontainer/component-test/softwarecontainer_test.cpp
```diff
@@ -78,29 +78,70 @@ void SoftwareContainerTest::TearDown()
...
 /*
  * Create a temporary directory or file, and optionally remove it.
  * Removal is useful if one only wants a unique tmp name
  */
-std::string SoftwareContainerTest::getTempPath(bool directory, bool shouldUnlink)
+std::string SoftwareContainerTest::getTempPath(bool directory)
 {
-    char fileTemplate[] = "/tmp/SC-tmpXXXXXX";
-    int fd = 0;
+    std::string name;
```
kursatkobya, line 144: it could be nice if you can initialize it here, maybe using a lambda like ...
```
std::string name = [&]{
    if (directory) {
        return createTempDir();
   } else {
        return createTempFile();
   }
} ();
```

tobsan: But you would still need the checks later to do unlink/rmdir, or you would do them in the lambda

PR #314, opened 2017-01-26
Clean ghost containers on startup

kursatkobya:
Add function to clean ghost containers on startup.
Add service tests for this feature .

### DIFF
--- a/agent/src/main.cpp
+++ b/agent/src/main.cpp
@@ -19,7 +19,6 @@
 
 
 #include <glibmm/optioncontext.h>
-#include <lxc/lxccontainer.h>
 
 #include "config/config.h"
 #include "config/configerror.h"
@@ -53,51 +52,9 @@ int signalHandler(void *data) {
     return 0;
 }
 
-void cleanGhostContainers(void)
-{
-    char **containers = nullptr;
-    struct lxc_container **cret = nullptr;
-    const char * basePath = lxc_get_global_config_item("lxc.lxcpath");
-    auto num = list_all_containers(basePath, &containers, &cret);
-    if (num > 0 && num < std::numeric_limits<int>::max()) {
-        log_warning() << num << " ghost containers found";
-        for (; num; num--) {
-            log_debug() << "Ghost container named " << containers[num-1] << " will be deleted";
-
-            const char * state = cret[num-1]->state(cret[num-1]);
-            if ("RUNNING" == std::string(state)) {
-                bool success = cret[num-1]->stop(cret[num-1]);
-                if (!success) {
-                    std::string errorMsg = "Unable to stop ghost container " +
-                                           std::string(containers[num-1]);
-                    throw SoftwareContainerAgentError(errorMsg);
-                }
-            }
-
-            bool success = cret[num-1]->destroy(cret[num-1]);
-            if (!success) {
-                std::string errorMsg = "Unable to destroy ghost container " +
-                                       std::string(containers[num-1]);
-                throw SoftwareContainerAgentError(errorMsg);
-            }
-
-            log_debug() << "Ghost container " << containers[num-1] << " is successfully destroyed";
-            delete cret[num-1];
-            delete containers[num-1];
-        }
-        delete cret;
-        delete containers;
-    }
-}
 
 int main(int argc, char **argv)
 {
-    try {
-        cleanGhostContainers();
-    } catch (SoftwareContainerAgentError &err) {
-        log_warning() << err.what();
-    }
-
     // Config file path default 'SC_CONFIG_FILE' shuld be set by the build system
     Glib::OptionEntry configOpt;
     configOpt.set_long_name("config");
--- a/agent/src/main.cpp
+++ b/agent/src/main.cpp
@@ -19,6 +19,7 @@
 
 
 #include <glibmm/optioncontext.h>
+#include <lxc/lxccontainer.h>
 
 #include "config/config.h"
 #include "config/configerror.h"
@@ -52,9 +53,51 @@ int signalHandler(void *data) {
     return 0;
 }
 
+void cleanGhostContainers(void)
+{
+    char **containers = nullptr;
+    struct lxc_container **cret = nullptr;
+    const char * basePath = lxc_get_global_config_item("lxc.lxcpath");
+    auto num = list_all_containers(basePath, &containers, &cret);
+    if (num > 0 && num < std::numeric_limits<int>::max()) {
+        log_warning() << num << " ghost containers found";
+        for (; num; num--) {
+            log_debug() << "Ghost container named " << containers[num-1] << " will be deleted";
+
+            const char * state = cret[num-1]->state(cret[num-1]);
+            if ("RUNNING" == std::string(state)) {
+                bool success = cret[num-1]->stop(cret[num-1]);
+                if (!success) {
+                    std::string errorMsg = "Unable to stop ghost container " +
+                                           std::string(containers[num-1]);
+                    throw SoftwareContainerAgentError(errorMsg);
+                }
+            }
+
+            bool success = cret[num-1]->destroy(cret[num-1]);
+            if (!success) {
+                std::string errorMsg = "Unable to destroy ghost container " +
+                                       std::string(containers[num-1]);
+                throw SoftwareContainerAgentError(errorMsg);
+            }
+
+            log_debug() << "Ghost container " << containers[num-1] << " is successfully destroyed";
+            delete cret[num-1];
+            delete containers[num-1];
+        }
+        delete cret;
+        delete containers;
+    }
+}
 
 int main(int argc, char **argv)
 {
+    try {
+        cleanGhostContainers();
+    } catch (SoftwareContainerAgentError &err) {
+        log_warning() << err.what();
+    }
+
     // Config file path default 'SC_CONFIG_FILE' shuld be set by the build system
     Glib::OptionEntry configOpt;
     configOpt.set_long_name("config");
--- a/agent/src/softwarecontaineragent.cpp
+++ b/agent/src/softwarecontaineragent.cpp
@@ -18,6 +18,7 @@
  */
 
 #include <cstdint>
+#include <lxc/lxccontainer.h>
 
 #include "softwarecontaineragent.h"
 
@@ -73,6 +74,8 @@ SoftwareContainerAgent::SoftwareContainerAgent(Glib::RefPtr<Glib::MainContext> m
     m_filteredConfigStore = std::make_shared<FilteredConfigStore>(serviceManifestDir);
     m_defaultConfigStore  = std::make_shared<DefaultConfigStore>(defaultServiceManifestDir);
 
+    removeOldContainers();
+
     m_containerConfig = SoftwareContainerConfig(
 #ifdef ENABLE_NETWORKGATEWAY
                                                 createBridge,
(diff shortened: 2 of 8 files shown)


agent/src/main.cpp
```diff
@@ -52,9 +53,51 @@ int signalHandler(void *data) {
     return 0;
 }
 
+void cleanGhostContainers(void)
```
breakreturn: I don't like this name that much. Could it be more descriptive? E.g. removeOldContainers or something like that.

agent/src/main.cpp
```diff
@@ -52,9 +53,51 @@ int signalHandler(void *data) {
...
+{
+    char **containers = nullptr;
+    struct lxc_container **cret = nullptr;
```
breakreturn: "cret"? A more descriptive name would be nice.

agent/src/main.cpp
```diff
@@ -52,9 +53,51 @@ int signalHandler(void *data) {
...
+    const char * basePath = lxc_get_global_config_item("lxc.lxcpath");
```
breakreturn: The * goes next to the variable name. Goes for a few other places in the file as well.

agent/src/main.cpp
```diff
@@ -52,9 +53,51 @@ int signalHandler(void *data) {
...
+    auto num = list_all_containers(basePath, &containers, &cret);
+    if (num > 0 && num < std::numeric_limits<int>::max()) {
+        log_warning() << num << " ghost containers found";
```
breakreturn: I think we should describe what happens better, "ghost" is perhaps better explained as "previously started", or some other wording that explains what is going on.

agent/src/main.cpp
breakreturn: Comments for this method would be nice.

agent/src/main.cpp
```diff
@@ -52,9 +53,51 @@ int signalHandler(void *data) {
...
+            delete containers[num-1];
+        }
+        delete cret;
+        delete containers;
+    }
+}
 
 int main(int argc, char **argv)
 {
+    try {
```
breakreturn: What's the reason we do this in main and not in the agent?

servicetest/agent/test_agent.py
```diff
@@ -0,0 +1,84 @@
...
+# Test suites #
+
+
+@pytest.mark.usefixtures("agent")
+class TestClearGhostContainers(object):
+    """ This test suite is stands for testing agent to examine cleaning
+        ghost containers feature
+    """
+
+    def testme(self):
```
breakreturn: This is not a good test name.

breakreturn (changes requested):

agent/src/main.cpp
kursatkobya: sure

agent/src/main.cpp
```diff
@@ -52,9 +53,51 @@ int signalHandler(void *data) {
...
+void cleanGhostContainers(void)
+{
+    char **containers = nullptr;
+    struct lxc_container **cret = nullptr;
+    const char * basePath = lxc_get_global_config_item("lxc.lxcpath");
+    auto num = list_all_containers(basePath, &containers, &cret);
+    if (num > 0 && num < std::numeric_limits<int>::max()) {
+        log_warning() << num << " ghost containers found";
+        for (; num; num--) {
+            log_debug() << "Ghost container named " << containers[num-1] << " will be deleted";
```
tobsan: Instead of indexing in the array on every access, couldn't you save that as a variable inside the loop instead?

`struct lxc_container *container = containers[num-1];`

agent/src/main.cpp
```diff
@@ -52,9 +53,51 @@ int signalHandler(void *data) {
...
+
+            const char * state = cret[num-1]->state(cret[num-1]);
+            if ("RUNNING" == std::string(state)) {
```
tobsan: Use `is_running()` from `lxccontainer.h` instead if you only want to check for running and don't care about other states.

agent/src/main.cpp
```diff
@@ -52,9 +53,51 @@ int signalHandler(void *data) {
...
+                bool success = cret[num-1]->stop(cret[num-1]);
```
tobsan: Dunno if it would make more sense to use shutdown() before calling stop(). What do you think?

agent/src/main.cpp
```diff
@@ -52,9 +53,51 @@ int signalHandler(void *data) {
...
+
+            bool success = cret[num-1]->destroy(cret[num-1]);
+            if (!success) {
+                std::string errorMsg = "Unable to destroy ghost container " +
+                                       std::string(containers[num-1]);
+                throw SoftwareContainerAgentError(errorMsg);
+            }
+
+            log_debug() << "Ghost container " << containers[num-1] << " is successfully destroyed";
+            delete cret[num-1];
```
tobsan: I think you should use `lxc_container_put()` instead

agent/src/main.cpp
tobsan: Yes, this should be done in the agent. Partly because one might have configured to put containers in a non-standard location, which one would need to know when getting all containers from lxc.

tobsan (changes requested):

agent/src/main.cpp
kursatkobya: i could do but, this way seem better to me. Do you object for readability concerns ?

tobsan: Yes, this is hard to read.

agent/src/main.cpp
kursatkobya: Running containers cannot be destroyed before stopping. But after stopping they can be destroyed. I also tested this scenario and it works this way.

agent/src/main.cpp
kursatkobya: instead of delete ? is that function frees the resource ?

agent/src/main.cpp
kursatkobya: I thought this operation is earlier initialization process before creating agent object. That's why it is implemented there. Surely it can be moved into agent.

agent/src/main.cpp
tobsan: Yes, but there are two ways to stop a container before destroying it: shutdown() and stop(), both are on the LXC API. shutdown() is the nicer version, and stop() is the SIGKILL version. Please use shutdown() as the first measure, and then use stop() if shutdown() returns false.

agent/src/main.cpp
tobsan: Just noticed that the API explicitly says that you should free the args, so never mind my first comment :)

agent/src/softwarecontaineragent.cpp
```diff
@@ -91,6 +94,47 @@ SoftwareContainerAgent::~SoftwareContainerAgent()
...
+void SoftwareContainerAgent::removeOldContainers()
+{
+    char **containerNames = nullptr;
+    struct lxc_container **containerList = nullptr;
+
+    const char *basePath = lxc_get_global_config_item("lxc.lxcpath");
+    auto num = list_all_containers(basePath, &containerNames, &containerList);
+    if (num > 0 && num < std::numeric_limits<int>::max()) {
+        log_warning() << num << " unused deprecated containers found";
+        for (; num; num--) {
```
tobsan: This type of truth checking is not nice. I would prefer a regular for loop with a new variable which you can compare to num. That way, you also can get rid of the `if (num > 0)` check above I guess.

TL;DR: Numbers are not bools :)

servicetest/agent/test_agent.py
```diff
@@ -0,0 +1,85 @@
...
+
+            The second phase of the test is running agent again to see
+            ghost container is cleaned.
+        """
+        sc = Container()
+        container_id = sc.start(DATA)
+        os.system("pkill --signal SIGKILL -f softwarecontainer-agent")
+        time.sleep(0.5)
+
+        os.system("lxc-ls >> output")
```
tobsan, line 83: It would be nice to be able to use the LXC Python binding here. Dunno if that would mean lots of more work, but it does feel more robust than parsing the lxc-ls output.

tobsan (changes requested):

servicetest/agent/test_agent.py
```diff
@@ -0,0 +1,85 @@
...
+from testframework import Container, SoftwareContainerAgentHandler
+
+
+CURRENT_DIR = os.path.dirname(os.path.realpath(__file__))
+
+
+# Provide what the testframework requires #####
+
+# This function is used by the 'agent' fixture to know where the log should be stored
+def logfile_path():
```
breakreturn, line 34: We don't seem to ignore this logfile for git anywhere. I propose we follow the same "testoutput dir" pattern as most of the other tests do, and put the log file there, and ignore the directory in .gitignore

breakreturn (changes requested):

agent/src/softwarecontaineragent.cpp
kursatkobya: checking upper and lower limits wouldn't hurt, contrarily it helps a lot in aspect of secure code imo. Yet i can add another index variable if it is really needed.

kursatkobya: ah you were talking about   `for (; num; num--)` 
aye aye sir

servicetest/agent/test_agent.py
kursatkobya: let me try this

breakreturn (approved):

agent/src/softwarecontaineragent.cpp
```diff
@@ -18,6 +18,7 @@
  */
 
 #include <cstdint>
+#include <lxc/lxccontainer.h>
```
tobsan, line 21: Could you add so that we link against LXC in the CMakeLists.txt for the agent as well?

agent/src/softwarecontaineragent.cpp
```diff
@@ -91,6 +94,47 @@ SoftwareContainerAgent::~SoftwareContainerAgent()
...
+/* we are fetching lxcpath from config now. But in near future we should use lxcConfigPath instead
+ * */
+void SoftwareContainerAgent::removeOldContainers()
+{
+    char **containerNames = nullptr;
+    struct lxc_container **containerList = nullptr;
+
+    const char *basePath = lxc_get_global_config_item("lxc.lxcpath");
+    auto num = list_all_containers(basePath, &containerNames, &containerList);
+    if (num > 0 && num < std::numeric_limits<int>::max()) {
```
tobsan: Wouldn't you rather want to change this to check for -1 (the error case) and otherwise just use the for loop to iterate over the found containers?

tobsan (changes requested):

agent/src/softwarecontaineragent.cpp
kursatkobya: if we are trusting the library that we are sure that `list_all_containers` only return -1  on error and count of containers on success, then you are right. This check is questioning borders instead of error check, which is safer. I didn't look in `list_all_containers` code so i don't know whether it has checks or not.

agent/src/softwarecontaineragent.cpp
```diff
@@ -91,6 +94,50 @@ SoftwareContainerAgent::~SoftwareContainerAgent()
...
+void SoftwareContainerAgent::removeOldContainers()
+{
+    char **containerNames = nullptr;
+    struct lxc_container **containerList = nullptr;
+
+    const char *basePath = lxc_get_global_config_item("lxc.lxcpath");
+    auto num = list_all_containers(basePath, &containerNames, &containerList);
+
+    if (-1 == num) {
+        throw SoftwareContainerAgentError("An error is occurred while trying to get container list");
```
breakreturn, line 114: Probably good to log here as well.

agent/src/softwarecontaineragent.cpp
```diff
@@ -91,6 +94,51 @@ SoftwareContainerAgent::~SoftwareContainerAgent()
...
+void SoftwareContainerAgent::removeOldContainers()
+{
+    char **containerNames = nullptr;
+    struct lxc_container **containerList = nullptr;
+
+    const char *basePath = lxc_get_global_config_item("lxc.lxcpath");
+    auto num = list_all_containers(basePath, &containerNames, &containerList);
+
+    if (-1 == num) {
+        log_warning() << "An error is occurred while trying to get deprecated container list";
```
tobsan: This should be log_error().

tobsan (approved):
Looks good, just had one minor comment

PR #315, opened 2017-01-26
Remove success flag in api docs

frznlogic:


breakreturn:
The "Methods" paragraph still mentions the success value.

breakreturn:
I think the docs should say that errors are raised as D-Bus errors

breakreturn (changes requested):
See comments

breakreturn (approved):

tobsan (approved):

PR #316, opened 2017-01-27
Added new states for SoftwareContainer class

breakreturn:
SC now throws exceptions if an operation is made when SC
is in an inappropriate state. Failures in the Container
class now results in exceptions as well.

All methods on the SC interface are not updated fully to
use the new states and exceptions yet.

### DIFF
--- a/libsoftwarecontainer/include/softwarecontainer.h
+++ b/libsoftwarecontainer/include/softwarecontainer.h
@@ -24,6 +24,7 @@
 #include "signalconnectionshandler.h"
 #include "config/softwarecontainerconfig.h"
 #include "filetoolkitwithundo.h"
+#include "softwarecontainererror.h"
 
 #include <glibmm.h>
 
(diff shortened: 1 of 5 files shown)


libsoftwarecontainer/include/softwarecontainer.h
```diff
@@ -43,6 +44,105 @@ class Gateway;
...
+ *
+ *  * shutdown() - successful call triggers a transition to TERMINATED
+ *  * resume() - successful call triggers a transition to READY
+ *  * getContainer()
+ *
+ * When SC is in state 'INVALID', something has gone wrong in the underlying container
+ * implementation and the container should be considered to be in a broken state from
+ * which it cannot recover.
+ *
+ * The follwoing methods are allowed to be called in every state:
```
tobsan: following*

tobsan (changes requested):
Looks good, only had a typo to note.

libsoftwarecontainer/include/softwarecontainer.h
breakreturn: fixed

tobsan (approved):

PR #317, opened 2017-01-27
Fixes some of the issues reported by clang.

tobsan:


common/softwarecontainer-common.cpp
```diff
@@ -93,22 +95,15 @@ bool existsInFileSystem(const std::string &path)
...
-
-    auto path = path_;
-
-    // Remove trailing backslashes
-    while ((path.size() > 0) && (path[path.size() - 1] == separator_char)) {
-        path.resize(path.size() - 1);
-    }
+    std::string path2 = path_; // copy constructor
+    char *path = (char *)path2.c_str();
+    std::string parent = dirname(path);
```
kursatkobya, line 105: i think its not nice to hack constant char to alter it. Althought it is nice to use dirname, it should be better if you make proper conversion.

DunderRoffe: I agree, we should use strncpy instead

tobsan: I actually had it as strdup() before, but wanted to avoid having to use that, since it might fail.

breakreturn (approved):
I won't have time to take a proper look at this, if other reviewers are satisfied, I am too.

kursatkobya (changes requested):
just a little comment

common/softwarecontainer-common.cpp
kursatkobya: i think it would be wise to allocate new area from stack to a char pointer and copy that string into it with strcpy. otherwise this function can be error prone since you hack constant. And also it would be dangerous if the caller calls with static string like parentPath("/gateways/app")

kursatkobya (approved):

PR #318, opened 2017-01-27
Updated SC class with regards to state usage

breakreturn:
All relevant methods respects the state of SC now.
Also added an error type for invalid containers.

### DIFF
--- a/libsoftwarecontainer/src/softwarecontainer.cpp
+++ b/libsoftwarecontainer/src/softwarecontainer.cpp
@@ -168,6 +168,15 @@ void SoftwareContainer::addGateway(Gateway *gateway)
 
 ReturnCode SoftwareContainer::startGateways(const GatewayConfiguration &gwConfig)
 {
+    assertValidState();
+
+    if (m_containerState != ContainerState::READY) {
+        std::string message = "Invalid to start gateways on non ready container " +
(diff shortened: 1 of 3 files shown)


libsoftwarecontainer/src/softwarecontainer.cpp
```diff
@@ -330,6 +374,23 @@ ReturnCode SoftwareContainer::shutdownGateways()
...
+    std::string id = std::string(m_container->id());
+
+    if (m_containerState == ContainerState::INVALID) {
+        std::string message = "Container is invalid " + id;
+        log_error() << message;
+        throw InvalidContainerError(message);
+    }
+
+    if (m_containerState != ContainerState::READY
+        && m_containerState != ContainerState::SUSPENDED)
```
tobsan: Does it make sense to allow access to container implementation when it is suspended? What can you possibly want to do except resuming/destroying it?

tobsan (commented):
Just had one comment, looks good otherwise.

libsoftwarecontainer/src/softwarecontainer.cpp
breakreturn: I don't know. I'll make it so it's only valid in state READY, ok?

PR #324, opened 2017-01-31
common: nicer stat usage

tobsan:

### DIFF
--- a/common/softwarecontainer-common.cpp
+++ b/common/softwarecontainer-common.cpp
@@ -35,20 +35,19 @@
 namespace softwarecontainer {
     LOG_DECLARE_DEFAULT_CONTEXT(defaultLogContext, "MAIN", "Main context");
 
-struct stat getStat(const std::string &path)
+bool getStat(const std::string &path, struct stat &st)
 {
-    struct stat st;
-    memset(&st, 0, sizeof(st));
-
-    if (stat(path.c_str(), &st) == 0) {
-        return st;
-    }
-    return st;
+    return stat(path.c_str(), &st) == 0;
 }
 
 bool isDirectory(const std::string &path)
 {
-    return S_ISDIR(getStat(path).st_mode);
+    struct stat st;
+    if (getStat(path, st)) {
+        return S_ISDIR(st.st_mode);
+    }
+
+    return false;
 }
 
 bool isDirectoryEmpty(const std::string &path) {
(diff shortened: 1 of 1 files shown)


common/softwarecontainer-common.cpp
```diff
@@ -35,20 +35,22 @@
 namespace softwarecontainer {
     LOG_DECLARE_DEFAULT_CONTEXT(defaultLogContext, "MAIN", "Main context");
 
-struct stat getStat(const std::string &path)
+bool getStat(const std::string &path, struct stat &st)
 {
-    struct stat st;
-    memset(&st, 0, sizeof(st));
-
     if (stat(path.c_str(), &st) == 0) {
```
jonte: Maybe shorten the function to:

`return stat(path.c_str(), &st) == 0;`

breakreturn:
No tests changed? I assume we don't have any then. Might be a good idea to add unit tests for this?

common/softwarecontainer-common.cpp
```diff
@@ -75,22 +77,38 @@ bool isDirectoryEmpty(const std::string &path) {
 
 bool isFile(const std::string &path)
 {
-    return S_ISREG(getStat(path).st_mode);
+    struct stat st;
+    if (getStat(path, st)) {
+        return S_ISREG(st.st_mode);
```
jonte, line 79: I might be nitpicking, but maybe `return getStat(path, st) && S_ISREG(st.st_mode);`

I'm OK w/ the semantics here though.

common/softwarecontainer-common.cpp
```diff
@@ -75,22 +77,38 @@ bool isDirectoryEmpty(const std::string &path) {
...
-    return S_ISSOCK(getStat(path).st_mode);
+    struct stat st;
+    if (getStat(path, st)) {
+        return S_ISSOCK(st.st_mode);
+    }
+
+    return false;
 }
 
 bool existsInFileSystem(const std::string &path)
```
jonte, line 105: This seems like a weird call overall.. Why would you ever check this? What you want is to stat if the file exists and is in some way accessible, right?

tobsan:
@breakreturn we already have unit tests for this. Since this only changes the internals of the functions, those tests are still valid and running.

jonte (approved):
Looks good, I have some nitpicking if you're interested in polishing.

common/softwarecontainer-common.cpp
tobsan: Yes.

PR #326, opened 2017-01-31
Sets core dump limit when running inside container

tobsan:

### DIFF
--- a/doc/chapters/debugging/index.rst
+++ b/doc/chapters/debugging/index.rst
@@ -38,12 +38,12 @@ Cleanup
 -------
 
 The internal filesystem will always be deleted and recreated when a container
-is restarted (destroyed and recreated). This means that anything you want to 
-save during debugging needs to be saved to a bind mounted directory before you 
+is restarted (destroyed and recreated). This means that anything you want to
+save during debugging needs to be saved to a bind mounted directory before you
 shut down the container.
 
 ``Note:`` This does not mean that the bind mounted filesystems will be deleted,
-and large parts of the filesystems are usually bind mounted. 
+and large parts of the filesystems are usually bind mounted.
 
 `SoftwareContainerAgent` will also delete all container instances upon startup,
 meaning that container instances will be lost when restarting.
@@ -108,7 +108,7 @@ able to connect to the process inside the container.
 
 1. Start a container
 2. Start `QtCreator` as root (sudo)
-3. Open a project in `QtCreator` 
+3. Open a project in `QtCreator`
 4. Build the code into a runnable binary
 5. Run the binary program inside the container
 6. If you are using `DBUS`, a `PID` is returned, otherwise find it in the outside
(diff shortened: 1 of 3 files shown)


doc/chapters/debugging/index.rst
```diff
@@ -120,11 +120,23 @@ able to connect to the process inside the container.
...
+create a core dump in case of, for example, a segmentation fault.
+
+The naming of a core dump file is the same inside a container as on the host, meaning this can
+be set by looking at `/proc/sys/kernel/core_pattern` or by running `sysctl kernel.core_pattern`.
+Because the container and host share a kernel, the setting on the host will be visible in the
+container.
+
+Note that the core_pattern can contain a path to where to store core dumps, so a wise decision from
+a systems perspective, would be to set it to some directory which is then mounted into each
+container, either by issuing `BindMount` calls on D-Bus, or by adding it as a FileGateway
```
frznlogic, line 133: Imho, the pattern should put the core dump in the CWD and execute from a directory with RW access. Setting a specific path will wreak havoc on apps that might crash outside containers.

frznlogic (changes requested):
Looks good except for the change in question.

libsoftwarecontainer/src/container.h
```diff
@@ -206,6 +206,16 @@ class Container :
     ReturnCode setEnvironmentVariable(const std::string &var, const std::string &val);
 
 private:
+    /*
```
frznlogic: Needs to be /** then ;)

frznlogic (changes requested):

doc/chapters/debugging/index.rst
tobsan: I don't quite follow. Let's say that the pattern is `/var/coredumps/%e-%t`. That would work outside a container (given that `/var/coredumps/` exists), and if one were to bind-mount it to `/var/coredumps/` in the container, it would work there as well.

If one wants to store all core dumps somewhere, setting this pattern to some path + %whatever seems reasonable, no?

frznlogic: Good point. Adding that an alternative is to save it in CWD, given that CWD is RW, would be nice however.

doc/chapters/debugging/index.rst
```diff
@@ -120,11 +120,27 @@ able to connect to the process inside the container.
...
+configuration in a default capability (which would then be automatically enabled for every
+container).
+
+An alternative is to not use such a specific pattern, and make sure you always execute your
```
frznlogic: "not use a pattern containing a path" would be preferable imho :)

doc/chapters/debugging/index.rst
```diff
@@ -120,11 +120,27 @@ able to connect to the process inside the container.
...
+
+Note that the core_pattern can contain a path to where to store core dumps, so a wise decision from
+a systems perspective, would be to set it to some directory which is then mounted into each
+container, either by issuing `BindMount` calls on D-Bus, or by adding it as a FileGateway
+configuration in a default capability (which would then be automatically enabled for every
+container).
+
+An alternative is to not use a pattern containing a path, and make sure you always execute your
+application from a cwd which is also bind-mounted (which the application directory always is), and
+then your core dumps will be stored together with the application binary.
```
frznlogic, line 139: Actually... one more request, it would be nice with a pointer to how the pattern should look and what it can contain... some other documentation I guess.

frznlogic (approved):

PR #330, opened 2017-02-02
Create helper for job return values

kursatkobya:
* Add a function to jobs for checking exit code status
* Refactor code to use this function when needed

### DIFF
--- a/libsoftwarecontainer/include/jobabstract.h
+++ b/libsoftwarecontainer/include/jobabstract.h
@@ -52,6 +52,22 @@ public:
     pid_t pid();
 
     /**
+     * Helper about return value of jobs.
+     * Describes whether the exit status of a command or a function indicates success or not
+     *
+     * Note that this method will interpret exit status only for the last job
+     */
+    bool isSuccess();
+
+    /**
+     * Helper about return value of jobs.
+     * Describes whether the exit status of a command or a function indicates error or not
+     *
+     * Note that this method will interpret exit status only for the last job
+     */
+    bool isError();
+
+    /**
      * That method always returns true as soon as the start() method has been called, even if the command fails to start,
      * since we don't know if the exec() occurring after the fork into the container actually succeeds...
      */
@@ -63,6 +79,7 @@ protected:
     EnvironmentVariables m_env;
 
     ExecutablePtr m_executable;
+    int m_exitStatus;
 
     pid_t m_pid = 0;
     int m_stdin[2] = {UNASSIGNED_STREAM, UNASSIGNED_STREAM};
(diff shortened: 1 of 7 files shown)


libsoftwarecontainer/src/jobs/jobabstract.cpp
```diff
@@ -57,7 +58,8 @@ void JobAbstract::captureStderr()
 
 int JobAbstract::wait()
 {
-    return waitForProcessTermination(m_pid);
+    m_exitStatus = waitForProcessTermination(m_pid);
+    return m_exitStatus;
```
tobsan, line 62: `m_exitStatus` is a bool, not an int. Either make it an int and use it as such (I would prefer that since there are more than two possible exit codes) or use it as a bool.

libsoftwarecontainer/src/jobs/jobabstract.cpp
```diff
@@ -80,6 +82,24 @@ pid_t JobAbstract::pid()
     return m_pid;
 }
 
+bool JobAbstract::isSuccess()
+{
+    if (0 == m_exitStatus) {
```
tobsan, line 87: Again, this is a bool, not an int.

libsoftwarecontainer/src/jobs/jobabstract.cpp
```diff
@@ -80,6 +82,24 @@ pid_t JobAbstract::pid()
...
+        return true;
+    }
+
+    return false;
+}
+
+bool JobAbstract::isError()
+{
+    if (0 == m_exitStatus) {
```
tobsan, line 96: And here as well...

tobsan (changes requested):
I like these changes except the mixup of bool/int.

frznlogic (approved):
Looks good to me, would be nice with another set of eyes though.

libsoftwarecontainer/src/jobs/jobabstract.cpp
frznlogic: I agree, use it as an int.

kursatkobya: it should have been int, and fixed

libsoftwarecontainer/include/jobabstract.h
```diff
@@ -52,6 +52,18 @@ class JobAbstract
...
 
     /**
+     * Helper about return value of jobs.
+     * Describes whether the exit status of a command or a function indicates success or not
+     */
+    bool isSuccess();
+
+    /**
+     * Helper about return value of jobs.
+     * Describes whether the exit status of a command or a function indicates error or not
```
DunderRoffe, line 64: Add note that says this will be reset when the job are run again.

tobsan (approved):

PR #331, opened 2017-02-02
adds service tests for core dump functionality

tobsan:

### DIFF
--- a/servicetest/coredump/test_coredump.py
+++ b/servicetest/coredump/test_coredump.py
@@ -1,0 +1,174 @@
+
+# Copyright (C) 2016-2017 Pelagicore AB
+#
+# Permission to use, copy, modify, and/or distribute this software for
+# any purpose with or without fee is hereby granted, provided that the
+# above copyright notice and this permission notice appear in all copies.
+#
+# THE SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL
+# WARRANTIES WITH REGARD TO THIS SOFTWARE INCLUDING ALL IMPLIED
+# WARRANTIES OF MERCHANTABILITY AND FITNESS. IN NO EVENT SHALL THE AUTHOR
+# BE LIABLE FOR ANY SPECIAL, DIRECT, INDIRECT, OR CONSEQUENTIAL DAMAGES
+# OR ANY DAMAGES WHATSOEVER RESULTING FROM LOSS OF USE, DATA OR PROFITS,
+# WHETHER IN AN ACTION OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION,
+# ARISING OUT OF OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS
+# SOFTWARE.
+#
+# For further information see LICENSE
+
+import pytest
+
+import os
+import time
+import signal
+import shutil
+
+from testframework.testhelper import CoreDumpHelper
+from testframework import Container
+from testframework import Capability
+from testframework import DefaultManifest
+
+CURRENT_DIR = os.path.dirname(os.path.realpath(__file__))
+TESTOUTPUT_DIR = CURRENT_DIR + "/testoutput"
+
+# This function is used by the 'agent' fixture to know where the log should be stored
+def logfile_path():
+    return TESTOUTPUT_DIR + "/test.log"
+
+def coredump_path():
+    return "/tmp/coredumps"
+
+# These default values are used to pass various test specific values and
+# configurations to the Container helper. Tests that need to add, remove or
+# update entries can simply base their dict on this one for convenience.
+DATA = {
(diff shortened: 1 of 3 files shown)


servicetest/coredump/test_coredump.py
```diff
@@ -0,0 +1,175 @@
...
+    test_pattern = coredump_path() + "/coredump"
+    old_pattern = ""
+
+    with open(pattern_path, "r") as core_file:
+        old_pattern = core_file.read().strip()
+
+    with open(pattern_path, "w") as core_file:
+        core_file.write(test_pattern)
+
+    with open(pattern_path, "r") as core_file:
```
breakreturn: Is this here to make sure that write() works as expected?

servicetest/coredump/test_coredump.py
```diff
@@ -0,0 +1,175 @@
...
+        assert(core_file.read().strip() == test_pattern)
+
+    yield test_pattern
+
+    # Teardown
+    with open(pattern_path, "w") as core_file:
+        core_file.write(old_pattern)
+
+    with open(pattern_path, "r") as core_file:
+        assert(core_file.read().strip() == old_pattern)
```
breakreturn: I think we could remove these asserts that are not part of the test. Unless there's a good reason for us to not trust that the file operations went as expected.

servicetest/coredump/test_coredump.py
```diff
@@ -0,0 +1,175 @@
...
+@pytest.fixture
+def create_coredump_dir(scope="module"):
+    """ Create the directories needed for getting the core dumps
+    """
+
+    # Create the core dump directory if it doesn't already exist
+    if not os.path.exists(coredump_path()):
+        os.makedirs(coredump_path())
+
+    yield
```
frznlogic, line 97: Not sure what yield will do in this case? We are not yielding a value of any kind?

servicetest/coredump/test_coredump.py
```diff
@@ -0,0 +1,175 @@
...
+                              sc.get_bind_dir() +
+                              "/testhelper.py --test-dir " +
+                              sc.get_bind_dir() + "/testoutput" +
+                              " --wait-for-core-dump " + coredump_path()
+                              )
+            # Let the helper run
+            time.sleep(0.25)
+
+            # Generate a segfault, causing core to be dumped
+            os.kill(pid, signal.SIGSEGV)
```
frznlogic: Isn't there a smarter way of generating a core dump? Couldn't the testhelper create one directly with a call like raise(SIGABRT) in c/c++?

frznlogic: os.abort()?

servicetest/testframework/testhelper.py
```diff
@@ -206,12 +207,43 @@ def result(self):
...
+        with open(self._base_path + "/core_limits", "w") as fh:
+            fh.write(str(soft))
+
+    def limitResult(self):
+        with open(self._base_path + "/core_limits", "r") as fh:
+            return fh.read()
+
+    def waitForSignal(self):
+        while True:
+            time.sleep(0.25)
```
frznlogic: Would be nice with more comments I guess... you can read yourself to this stuff from the code I guess though.

frznlogic (changes requested):
Looks good, but some small comments in the code.

servicetest/coredump/test_coredump.py
breakreturn: yield returns control to the consuming test, and when the test is done, the fixture does whatever comes after the yield, i.e. cleanup/teardown things.

servicetest/coredump/test_coredump.py
tobsan: Good point, I will change to that instead.

servicetest/testframework/testhelper.py
tobsan: If the helper aborts itself this function would not be needed.

servicetest/coredump/test_coredump.py
breakreturn: But isn't it nice to have the test code explicit in what the sequence look like? If the core dump is generated implicitly by using a certain helper it can perhaps be confusing for the reader.

frznlogic: You would generate it explicitly from the Helper function, you get rid of several sleeps etc? Should be good imho ?

breakreturn: Yeah true I guess, but the test would not be as explicit IMO. But that can be mitigated with some more detailed comments on what's going on and how the sequence looks.

servicetest/testframework/testhelper.py
frznlogic: It was a general comment on the whole class =)

tobsan:
Updated according to comments.

breakreturn (approved):

frznlogic (approved):

DunderRoffe (approved):

PR #333, opened 2017-02-02
networkgw: Uses the bridge device from config

tobsan:
We used to have a cmake define here instead.

### DIFF
--- a/libsoftwarecontainer/component-test/networkgateway_componenttest.cpp
+++ b/libsoftwarecontainer/component-test/networkgateway_componenttest.cpp
@@ -29,7 +29,7 @@ public:
         NetworkGateway(11, // container id
                        std::string(BRIDGE_DEVICE_TESTING),
                        std::string(BRIDGE_IP_TESTING), // bridge ip
-                       std::stoi(BRIDGE_NETMASK_BITS_TESTING)) // bridge netmask 
+                       std::stoi(BRIDGE_NETMASK_BITLENGTH_TESTING)) // bridge netmask 
     {
     }
 
--- a/libsoftwarecontainer/component-test/networkgateway_componenttest.cpp
+++ b/libsoftwarecontainer/component-test/networkgateway_componenttest.cpp
@@ -25,8 +25,8 @@ class MockNetworkGateway :
     public NetworkGateway
 {
 public:
-    MockNetworkGateway() :
-        NetworkGateway(11, // container id
+    MockNetworkGateway(uint32_t containerID) :
+        NetworkGateway(containerID,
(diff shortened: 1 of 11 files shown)


libsoftwarecontainer/component-test/networkgateway_componenttest.cpp
```diff
@@ -26,10 +26,10 @@ class MockNetworkGateway :
 {
 public:
     MockNetworkGateway() :
-        NetworkGateway(1, // container id
+        NetworkGateway(11, // container id
```
kursatkobya: why this id is 11 ? randomizing this would be nice

kursatkobya (approved):
looks good, one small idea is there if you want to apply

libsoftwarecontainer/component-test/networkgateway_componenttest.cpp
```diff
@@ -26,10 +26,10 @@ class MockNetworkGateway :
...
+                       std::string(BRIDGE_DEVICE_TESTING),
                        std::string(BRIDGE_IP_TESTING), // bridge ip
-                       std::stoi(BRIDGE_NETMASK_BITS_TESTING) // netmask bits
-        )
+                       std::stoi(BRIDGE_NETMASK_BITS_TESTING)) // bridge netmask
```
breakreturn: This comment makes sense given the name of the config I guess, but this reminds me that the naming of this config is off. Shouldn't it be something like "netmask bit length" instead? I mean, it's not the actual bitmask really.

breakreturn (approved):

libsoftwarecontainer/component-test/networkgateway_componenttest.cpp
kursatkobya: it actually indicates that netmask will be how many bits. So netmask bit count  maybe

tobsan: Yeah, I agree. In the config, that is already changed. I can update it here also.

tobsan: Yeah, I agree. In the config, that is already changed. I can update it here also.

libsoftwarecontainer/component-test/networkgateway_componenttest.cpp
tobsan: I guess that could be done, will look into it!

tobsan: I guess that could be done, will look into it!

tobsan:
@kursatkobya I applied both your suggestions, thanks!

PR #336, opened 2017-02-02
CMake: Don't include doc in ALL.

tobsan:
One can simply type "make doc" to build them.


tobsan:
Will rework this actually, never mind this.

PR #337, opened 2017-02-06
Fail on unknown gw

DunderRoffe:


tobsan (approved):
This looks good. One could change the m_gateways member though to be a map instead of a vector, with IDs as keys.

DunderRoffe:
Yes that is a good idea. However, I fell that it is out of scope for this pull-request.

libsoftwarecontainer/src/softwarecontainer.cpp
```diff
@@ -197,6 +197,19 @@ ReturnCode SoftwareContainer::startGateways(const GatewayConfiguration &gwConfig
 
 ReturnCode SoftwareContainer::configureGateways(const GatewayConfiguration &gwConfig)
 {
+    assertValidState();
+
+    // Make sure that all the gateway ids in the given GatewayConfig can map
+    // to a gateway in SoftwareContainer
+    for (auto &id : gwConfig.ids()) {
+        for (auto &gateway : m_gateways) {
+            if (id != m_gateways) {
```
kursatkobya: how come you compare string id with vector of  Gateway pointer
is this really compiles ? Or maybe i misread the code

tobsan (approved):

tobsan (approved):

common/gatewayconfig.cpp
```diff
@@ -39,6 +39,22 @@ GatewayConfiguration::~GatewayConfiguration()
     }
 }
 
+ReturnCode GatewayConfiguration::append(const std::string &id, const std::string &jsonConf)
+{
+    json_error_t jsonError;
+    size_t flags = 0; // default flags
+
+    json_t *json = json_loads(jsonConf.c_str(), flags, &jsonError);
+    if (json == nullptr) {
```
kursatkobya: it would be nice if you write other way around : `nullptr == json`

libsoftwarecontainer/src/softwarecontainer.cpp
```diff
@@ -197,6 +197,23 @@ ReturnCode SoftwareContainer::startGateways(const GatewayConfiguration &gwConfig
...
+    // to a gateway in SoftwareContainer
+    for (auto &id : gwConfig.ids()) {
+        bool match = false;
+        for (auto &gateway : m_gateways) {
+            if (id == gateway->id()) {
+                match = true;
+            }
+        }
+        if (!match) {
+            throw GatewayError("Could not find any gateway matching id: " + id);
```
kursatkobya, line 212: Maybe this is a really small possibility but if there is no enabled gateway this could create false throw. Maybe it would be reasonable to check if there is any member in `m_gateways`.

kursatkobya (approved):
Looks nice. I gave small minor comments with my approval.

libsoftwarecontainer/src/softwarecontainer.cpp
DunderRoffe: This is how it should be. Specifying config for non-existent/loaded gateways is not supported.

PR #339, opened 2017-02-08
Replace ReturnCode with bool (softwarecontainer + gateways)

thenor:
This commit only removes occurrences of ReturnCode in softwarecontainer.cpp/.h and gateway (including classes that inherit from gateway).

### DIFF
--- a/libsoftwarecontainer/src/gateway/cgroups/cgroupsgateway.cpp
+++ b/libsoftwarecontainer/src/gateway/cgroups/cgroupsgateway.cpp
@@ -30,32 +30,32 @@ CgroupsGateway::CgroupsGateway()
 {
 }
 
-ReturnCode CgroupsGateway::readConfigElement(const json_t *element)
+bool CgroupsGateway::readConfigElement(const json_t *element)
 {
     try {
         m_parser.parseCGroupsGatewayConfiguration(element);
     } catch (CgroupsGatewayError &e) {
         log_error() << "Could not parse CGroups configuration element";
-        return ReturnCode::FAILURE;
+        return false;
     }
 
-    return ReturnCode::SUCCESS;
+    return true;
 }
 
 bool CgroupsGateway::activateGateway()
 {
-    ReturnCode success = ReturnCode::FAILURE;
+    bool success = false;
     auto cgroupSettings = m_parser.getSettings();
     for (auto& setting: cgroupSettings) {
-        success = getContainer()->setCgroupItem(setting.first, setting.second);
-        if (success != ReturnCode::SUCCESS) {
+        success = isSuccess(getContainer()->setCgroupItem(setting.first, setting.second));
+        if (!success) {
             log_error() << "Error activating Cgroups Gateway, could not set cgroup item "
                         << setting.first << ": " << setting.second;
             break;
         }
     }
 
-    return isSuccess(success);
+    return success;
 }
 
 bool CgroupsGateway::teardownGateway()
--- a/libsoftwarecontainer/src/gateway/devicenode/devicenodegateway.cpp
+++ b/libsoftwarecontainer/src/gateway/devicenode/devicenodegateway.cpp
@@ -31,14 +31,14 @@ DeviceNodeGateway::DeviceNodeGateway() :
 {
 }
 
-ReturnCode DeviceNodeGateway::readConfigElement(const json_t *element)
+bool DeviceNodeGateway::readConfigElement(const json_t *element)
 {
     DeviceNodeParser parser;
     DeviceNodeParser::Device dev;
 
-    if (isError(parser.parseDeviceNodeGatewayConfiguration(element, dev))) {
+    if (!parser.parseDeviceNodeGatewayConfiguration(element, dev)) {
         log_error() << "Could not parse device node configuration";
-        return ReturnCode::FAILURE;
+        return false;
     }
 
     return m_logic.updateDeviceList(dev);
@@ -46,21 +46,10 @@ ReturnCode DeviceNodeGateway::readConfigElement(const json_t *element)
 
 bool DeviceNodeGateway::activateGateway()
 {
-    auto returnValue = applySettings();
-    return isSuccess(returnValue);
-}
-
-bool DeviceNodeGateway::teardownGateway()
-{
-    return true;
-}
-
-ReturnCode DeviceNodeGateway::applySettings()
-{
     auto devlist = m_logic.getDevList();
     if (devlist.empty()) {
         log_info() << "Activate was called when no devices has been configured.";
-        return ReturnCode::FAILURE;
+        return false;
     }
 
     for (auto &dev : devlist) {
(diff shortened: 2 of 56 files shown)


libsoftwarecontainer/src/gateway/cgroups/cgroupsgateway.cpp
```diff
@@ -30,32 +30,32 @@ CgroupsGateway::CgroupsGateway()
...
 
 bool CgroupsGateway::activateGateway()
 {
-    ReturnCode success = ReturnCode::FAILURE;
+    bool success = false;
     auto cgroupSettings = m_parser.getSettings();
     for (auto& setting: cgroupSettings) {
-        success = getContainer()->setCgroupItem(setting.first, setting.second);
-        if (success != ReturnCode::SUCCESS) {
+        success = isSuccess(getContainer()->setCgroupItem(setting.first, setting.second));
```
breakreturn, line 50: I might be misreading this since it's a big diff in total, but this looks like we're adding a line that uses "isSuccess". Is this some other isSuccess than the ReturnCode related one?

libsoftwarecontainer/src/gateway/environment/envgatewayparser.cpp
```diff
@@ -21,23 +21,23 @@
...
     const json_t *element,
     EnvironmentVariable &result,
     const EnvironmentVariables &store)
 {
 
-    if (isError(requireNonEmptyKeyValue(element, "name", result.first))
-     || isError(requireNonEmptyKeyValue(element, "value", result.second))) {
-        return ReturnCode::FAILURE;
+    if (!requireNonEmptyKeyValue(element, "name", result.first)
+     || !requireNonEmptyKeyValue(element, "value", result.second)) {
```
breakreturn: This indentation is actually not according to the guidelines, the '||' should be aligned with the start of the first parameter.

libsoftwarecontainer/src/gateway/devicenode/devicenodegateway.cpp
```diff
@@ -31,36 +31,35 @@ DeviceNodeGateway::DeviceNodeGateway() :
...
     }
 
     return m_logic.updateDeviceList(dev);
 }
 
 bool DeviceNodeGateway::activateGateway()
 {
-    auto returnValue = applySettings();
-    return isSuccess(returnValue);
+    return applySettings();
```
tobsan: Is there a good reason to still keep `applySettings()` instead of moving its code here?

libsoftwarecontainer/src/gateway/gateway.cpp
```diff
@@ -126,13 +126,13 @@ bool Gateway::isActivated()
     return m_state >= GatewayState::ACTIVATED;
 }
 
-ReturnCode Gateway::setEnvironmentVariable(const std::string &variable, const std::string &value)
+bool Gateway::setEnvironmentVariable(const std::string &variable, const std::string &value)
 {
     if (hasContainer()) {
-        return getContainer()->setEnvironmentVariable(variable, value);
+        return isSuccess(getContainer()->setEnvironmentVariable(variable, value));
```
breakreturn, line 132: Same as above, an "isSuccess" which I'm not sure should be here.

breakreturn (changes requested):
Some comments in-line.

Another more general comment is that I would expect the isError and isSucess methods to be removed.

libsoftwarecontainer/src/gateway/gateway.cpp
tobsan, line 132: I guess `ContainerAbstractInterface::setEnvironmentVariable` should also be bool, eliminating the need for `isSuccess` here.

libsoftwarecontainer/src/gateway/network/iptableentry.cpp
```diff
@@ -22,30 +22,31 @@
...
-            return ReturnCode::FAILURE;
+            return false;
         }
     }
 
-    if (ReturnCode::FAILURE == insertCommand(interpretPolicy())) {
-        log_error() << "Unable to set policy " << convertTarget(m_defaultTarget) << " for " << m_type;
-        return ReturnCode::FAILURE;
+    if (!insertCommand(interpretPolicy())) {
+            log_error() << "Unable to set policy " << convertTarget(m_defaultTarget)
```
tobsan: Why the extra indentation here?

libsoftwarecontainer/src/softwarecontainer.cpp
```diff
@@ -178,25 +177,21 @@ ReturnCode SoftwareContainer::startGateways(const GatewayConfiguration &gwConfig
...
-        return result;
+    if (!configureGateways(gwConfig)) {
+        return false;
     }
 
-    result = activateGateways();
-    if (isError(result)) {
-        return result;
+    if (!activateGateways()) {
+        return false;
```
tobsan: You could combine configureGateways() and activateGateways() as follows, since or is lazy.

```
if (!configureGateways() || !activateGateways()) {
    return false;
}
```

libsoftwarecontainer/src/softwarecontainer.cpp
```diff
@@ -444,7 +439,7 @@ ReturnCode SoftwareContainer::bindMount(const std::string &pathOnHost, const std
         throw InvalidOperationError(message);
     }
 
-    return getContainer()->bindMountInContainer(pathOnHost, pathInContainer, readonly);
+    return isSuccess(getContainer()->bindMountInContainer(pathOnHost, pathInContainer, readonly));
```
tobsan, line 444: I guess this should also be a bool.

tobsan (changes requested):
Mostly minor changes, looks good overall!

libsoftwarecontainer/src/gateway/cgroups/cgroupsgateway.cpp
thenor: The reason I need this isSuccess here is since I haven't yet replaced all the ReturnCode occurences (in this case for container).

breakreturn: Ok I see.

libsoftwarecontainer/src/gateway/devicenode/devicenodegateway.cpp
thenor: Not sure. I also thought this looked strange after the change.
Should I move it instead? I can see if apply is used anywhere else.

libsoftwarecontainer/src/gateway/environment/envgatewayparser.cpp
thenor: I see what you mean. I will fix it.

libsoftwarecontainer/src/gateway/gateway.cpp
thenor: Yes, but to avoid making the commit too big (I had already changed in ~20 files) I had to stop somewhere. Is this okay with you, @tobsan @breakreturn?

libsoftwarecontainer/src/gateway/network/iptableentry.cpp
thenor: I will fix it!

libsoftwarecontainer/src/softwarecontainer.cpp
thenor: Sure. Unless we want to add debug?

libsoftwarecontainer/src/softwarecontainer.cpp
thenor: This is also a function in the container, that hasn't been changed yet. As mentioned above, I felt I needed to make an intermediate commit.

libsoftwarecontainer/src/gateway/gateway.cpp
thenor: Same as above, so hope it's okay that I keep it like this until the next commit.

libsoftwarecontainer/src/softwarecontainer.cpp
tobsan: Good point, perhaps best to keep it as is.

thenor: I will keep it as is and add some debug printouts.

libsoftwarecontainer/src/gateway/gateway.cpp
breakreturn: Absolutely.

breakreturn: It's fine with me to break up the change in commits like this if you want.

breakreturn (approved):

tobsan (approved):

PR #344, opened 2017-02-10
Make the SCA testable

kursatkobya:
Add SoftwareContainerFactory class to obtain backend container
Add SoftwareContainerAbstractInterface class to be able to create different SC classes for different purposes
Refactor code regarding changes
Add unit tests for agent class

### DIFF
--- a/agent/component-test/softwarecontaineragent_componenttest.cpp
+++ b/agent/component-test/softwarecontaineragent_componenttest.cpp
@@ -23,6 +23,7 @@
 #include "config/configloader.h"
 #include "config/mainconfigsource.h"
 #include "config/configdefinition.h"
+#include "softwarecontainerfactory.h"
 
 #include <gmock/gmock.h>
 #include <gtest/gtest.h>
@@ -102,8 +103,9 @@ public:
                                                                   ConfigDependencies());
 
         try {
-            sca = std::make_shared<SoftwareContainerAgent>(m_context, config);
-        } catch(SoftwareContainerError &err) {
+            std::shared_ptr<SoftwareContainerFactory> factory = std::shared_ptr<SoftwareContainerFactory> (new SoftwareContainerFactory());
+            sca = std::make_shared<SoftwareContainerAgent>(m_context, config, factory);
+        }  catch(SoftwareContainerError &err) {
             log_error() << "Exception in software agent constructor";
             ASSERT_TRUE(false);
         }
@@ -183,7 +185,7 @@ TEST_F(SoftwareContainerAgentTest, ResumeNonSuspendedContainer) {
     ASSERT_THROW({
         ContainerID id = sca->createContainer(valid_config);
 
-        sca->getContainer(id);
+//        sca->getContainer(id);
         sca->resumeContainer(id);
     }, InvalidOperationError);
 }
--- a/agent/component-test/softwarecontaineragent_componenttest.cpp
+++ b/agent/component-test/softwarecontaineragent_componenttest.cpp
@@ -185,7 +185,6 @@ TEST_F(SoftwareContainerAgentTest, ResumeNonSuspendedContainer) {
     ASSERT_THROW({
         ContainerID id = sca->createContainer(valid_config);
 
-//        sca->getContainer(id);
         sca->resumeContainer(id);
     }, InvalidOperationError);
 }
--- a/agent/src/softwarecontaineragent.cpp
+++ b/agent/src/softwarecontaineragent.cpp
@@ -91,7 +91,6 @@ SoftwareContainerAgent::SoftwareContainerAgent(Glib::RefPtr<Glib::MainContext> m
                                                 sharedMountsDir,
                                                 shutdownTimeout);
 
-    log_debug() << "Created SCA instance :" ;
 }
 
 SoftwareContainerAgent::~SoftwareContainerAgent()
@@ -260,7 +259,7 @@ pid_t SoftwareContainerAgent::execute(ContainerID containerID,
     // Set up a CommandJob for this run in the container
     auto job = container->createCommandJob(cmdLine);
     if (nullptr == job) {
-        std::string errorMessage("Could not create job instance");
+        std::string errorMessage("Could not create job instance for " + cmdLine);
         log_error() << errorMessage;
         throw SoftwareContainerAgentError(errorMessage);
     }
--- a/agent/src/softwarecontaineragent.cpp
+++ b/agent/src/softwarecontaineragent.cpp
@@ -31,9 +31,11 @@
 namespace softwarecontainer {
 
 SoftwareContainerAgent::SoftwareContainerAgent(Glib::RefPtr<Glib::MainContext> mainLoopContext,
-                                               std::shared_ptr<Config> config):
+                                               std::shared_ptr<Config> config,
+                                               std::shared_ptr<SoftwareContainerFactory> factory):
     m_mainLoopContext(mainLoopContext),
-    m_config(config)
+    m_config(config),
+    m_factory(factory)
 {
     m_containerIdPool.push_back(0);
 
@@ -88,6 +90,8 @@ SoftwareContainerAgent::SoftwareContainerAgent(Glib::RefPtr<Glib::MainContext> m
                                                 lxcConfigPath,
                                                 sharedMountsDir,
                                                 shutdownTimeout);
+
+    log_debug() << "Created SCA instance :" ;
 }
 
 SoftwareContainerAgent::~SoftwareContainerAgent()
@@ -218,8 +222,7 @@ ContainerID SoftwareContainerAgent::createContainer(const std::string &config)
 
     // Get an ID and create the container
     ContainerID containerID = findSuitableId();
-    auto container = SoftwareContainerPtr(new SoftwareContainer(containerID,
-                                                                std::move(containerConfig)));
+    auto container = m_factory->createContainer(containerID, std::move(containerConfig));
     log_debug() << "Created container with ID :" << containerID;
 
     m_containers[containerID] = container;
@@ -256,6 +259,12 @@ pid_t SoftwareContainerAgent::execute(ContainerID containerID,
 
     // Set up a CommandJob for this run in the container
     auto job = container->createCommandJob(cmdLine);
+    if (nullptr == job) {
+        std::string errorMessage("Could not create job instance");
+        log_error() << errorMessage;
+        throw SoftwareContainerAgentError(errorMessage);
+    }
+
     job->setOutputFile(outputFile);
     job->setEnvironmentVariables(env);
     job->setWorkingDirectory(workingDirectory);
--- a/agent/src/softwarecontaineragent.h
+++ b/agent/src/softwarecontaineragent.h
@@ -80,7 +80,7 @@ protected:
 class SoftwareContainerAgent
 {
     LOG_DECLARE_CLASS_CONTEXT("SCA", "SoftwareContainerAgent");
-    typedef std::shared_ptr<ContainerAgentInterface> SoftwareContainerPtr;
+    typedef std::shared_ptr<SoftwareContainerAbstractInterface> SoftwareContainerPtr;
 
 public:
     /**
@@ -274,9 +274,6 @@ private:
      */
     ContainerOptionParser m_optionParser;
 
-    /*
-     * Responsible to provide connection to containers.
-     */
     std::shared_ptr<SoftwareContainerFactory> m_factory;
 };
 
--- a/agent/src/softwarecontaineragent.h
+++ b/agent/src/softwarecontaineragent.h
@@ -37,6 +37,7 @@
 #include "capability/defaultconfigstore.h"
 #include "config/config.h"
 #include "containeroptions/containeroptionparser.h"
+#include "softwarecontainerfactory.h"
 
 #include <jsonparser.h>
 #include "commandjob.h"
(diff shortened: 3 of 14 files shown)


agent/component-test/softwarecontaineragent_componenttest.cpp
```diff
@@ -183,7 +185,7 @@ TEST_F(SoftwareContainerAgentTest, ResumeNonSuspendedContainer) {
     ASSERT_THROW({
         ContainerID id = sca->createContainer(valid_config);
 
-        sca->getContainer(id);
+//        sca->getContainer(id);
```
tobsan: Remove this comment, please.

agent/src/softwarecontaineragent.cpp
```diff
@@ -88,6 +90,8 @@ SoftwareContainerAgent::SoftwareContainerAgent(Glib::RefPtr<Glib::MainContext> m
                                                 lxcConfigPath,
                                                 sharedMountsDir,
                                                 shutdownTimeout);
+
+    log_debug() << "Created SCA instance :" ;
```
tobsan: Please don't use "SCA" in printouts, write agent instead.

agent/src/softwarecontaineragent.cpp
```diff
@@ -256,6 +259,12 @@ pid_t SoftwareContainerAgent::execute(ContainerID containerID,
 
     // Set up a CommandJob for this run in the container
     auto job = container->createCommandJob(cmdLine);
+    if (nullptr == job) {
+        std::string errorMessage("Could not create job instance");
```
tobsan: `"Could not create command job for " + cmdLine` would be a better errorMessage

agent/src/softwarecontaineragent.h
```diff
@@ -79,7 +80,7 @@ class SoftwareContainerAgentError : public SoftwareContainerError
 class SoftwareContainerAgent
 {
     LOG_DECLARE_CLASS_CONTEXT("SCA", "SoftwareContainerAgent");
-    typedef std::shared_ptr<SoftwareContainer> SoftwareContainerPtr;
+    typedef std::shared_ptr<ContainerAgentInterface> SoftwareContainerPtr;
```
tobsan: This name is super confusing. Why not SoftwareContainerInterface? It is an interface to SoftwareContainer, isn't it?

agent/src/softwarecontainerfactory.cpp
```diff
@@ -0,0 +1,32 @@
...
+ *
+ * For further information see LICENSE
+ */
+
+#include "softwarecontainerfactory.h"
+
+namespace softwarecontainer {
+std::shared_ptr<ContainerAgentInterface>
+SoftwareContainerFactory::createContainer(const ContainerID id,
+                                  std::unique_ptr<const SoftwareContainerConfig> config)
```
tobsan: indentation

agent/src/softwarecontainerfactory.h
```diff
@@ -0,0 +1,46 @@
...
+#pragma once
+
+#include "softwarecontainer.h"
+#include "config/softwarecontainerconfig.h"
+#include "softwarecontainer-common.h"
+
+namespace softwarecontainer {
+
+/*
+ * This class holds a connection to agent to create containers.
```
tobsan: Not it doesn't, it is a class that creates containers. Its relation to the agent should be described in the agent, not here.

agent/src/softwarecontainerfactory.h
```diff
@@ -0,0 +1,46 @@
...
+ * WHETHER IN AN ACTION OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION,
+ * ARISING OUT OF OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS
+ * SOFTWARE.
+ *
+ * For further information see LICENSE
+ */
+
+/**
+ * @file softwarecontainerfactory.h
+ * @brief Contains the softwarecontainer::AgentFactory class
```
tobsan: This is not the name of the class.

agent/unit-test/softwarecontaineragent_unittest.cpp
```diff
@@ -0,0 +1,213 @@
...
+        return true;
+    }
+
+};
+
+class TestFactory: public SoftwareContainerFactory
+{
+public:
+    std::shared_ptr<ContainerAgentInterface> createContainer
+    (const ContainerID id __attribute__((unused)), std::unique_ptr<const SoftwareContainerConfig> config __attribute__((unused)))
```
tobsan: Indentation, break the line up instead.

agent/unit-test/softwarecontaineragent_unittest.cpp
```diff
@@ -0,0 +1,213 @@
...
+        std::shared_ptr<ContainerAgentInterface> container = std::shared_ptr<ContainerAgentInterface> (new TestContainerInterface());
+        return container;
+     }
+};
+} //namespace
+
+class SoftwareContainerAgentTest: public ::testing::Test
+{
+public:
+    SoftwareContainerAgent *sca;
```
tobsan: Use a smart pointer instead, then you won't need to delete it.

libsoftwarecontainer/include/containeragentinterface.h
```diff
@@ -0,0 +1,130 @@
...
+ * WHETHER IN AN ACTION OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION,
+ * ARISING OUT OF OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS
+ * SOFTWARE.
+ *
+ * For further information see LICENSE
+ */
+
+/**
+ * @file containeragentinterface.h
+ * @brief Contains the softwarecontainer::ContainerAbstractInterface class
```
tobsan: This name does not match.

agent/unit-test/softwarecontaineragent_unittest.cpp
```diff
@@ -0,0 +1,213 @@
...
+        auto id = sca->createContainer(valid_config);
+        sca->suspendContainer(id);
+        sca->resumeContainer(id);
+        sca->getContainer(id);
+        auto containers = sca->listContainers();
+        ASSERT_EQ(1, containers.size());
+
+        sca->shutdownContainer(id);
+    });
+}
```
tobsan: This test is not meaningful imho. Checking that functions we've stubbed does not throw does not help us. We probably want to make sure certain functions gets called when running a certain sequence (and we can stub those I guess).

tobsan (changes requested):
Looks good generally, some naming and checks that should be changed.

agent/src/softwarecontaineragent.h
```diff
@@ -270,6 +273,11 @@ class SoftwareContainerAgent
      * Responsible for parsing any dynamic values given when creating a container
      */
     ContainerOptionParser m_optionParser;
+
+    /*
+     * Responsible to provide connection to containers.
```
breakreturn: I think this could be described as the factory. A factory is a common pattern and further details should be in that class. This comment is a bit confusing at the moment.

agent/src/softwarecontaineragent.h
kursatkobya: Actually in the first idea the name of that class was ContainerAbstractInterface. Yet there is another ContainerAbstractInterface which connect software container to real back-end impelmentation. So not to create confusion i wanted to make it clear that this is interface used by agent. But may be it can be SoftwareContainerAbstractInterface. What do you think

tobsan: I see. In that case I think `SoftwareContainerAbstractInterface` is better, yes.

agent/unit-test/softwarecontaineragent_unittest.cpp
kursatkobya: The problem here even one argument passes the 100char rule. But you are right

agent/unit-test/softwarecontaineragent_unittest.cpp
```diff
@@ -0,0 +1,213 @@
...
+                                     "service-manifest-dir = " + std::string(SERVICE_MANIFEST_DIR_TESTING) + "\n"
+                                     "default-service-manifest-dir = " + std::string(DEFAULT_SERVICE_MANIFEST_DIR_TESTING) + "\n";
+
+    const std::string valid_config = "[{\"enableWriteBuffer\": false}]";
+
+    void SetUp() override
+    {
+        std::unique_ptr<ConfigLoader> loader(new StringConfigLoader(configString));
+        std::unique_ptr<ConfigSource> mainConfig(new MainConfigSource(std::move(loader),
+                                                                      ConfigDefinition::typeMap()));
```
breakreturn, line 134: I guess that should be TypeMap() not typeMap(). Did this compile?

agent/unit-test/softwarecontaineragent_unittest.cpp
breakreturn: Agree. The purpose is to have tests for the sequence, i.e. that some calls on the agent results in the expected call(s) to SoftwareContainer.

agent/unit-test/softwarecontaineragent_unittest.cpp
kursatkobya: I think TypeMap is a type and typeMap is an object to that type.  And yes it compiles. Ref : `static TypeMap typeMap();`

breakreturn: Oh yeah your'e right, my bad.

agent/src/softwarecontainerfactory.cpp
```diff
@@ -0,0 +1,32 @@
...
+
+#include "softwarecontainerfactory.h"
+
+namespace softwarecontainer {
+std::shared_ptr<SoftwareContainerAbstractInterface>
+SoftwareContainerFactory::createContainer(const ContainerID id,
+                                          std::unique_ptr<const SoftwareContainerConfig> config)
+{
+    auto container = std::shared_ptr<SoftwareContainerAbstractInterface>
+                                    (new SoftwareContainer(id, std::move(config)));
```
breakreturn: I think it would be more similar to other parts of the code if the opening parenthesis is on the line above. It's basically the same as when you have a method and need to break the line at the first arg:

longMethodName(
    arg1, arg2);

agent/src/softwarecontainerfactory.h
```diff
@@ -0,0 +1,46 @@
...
+
+namespace softwarecontainer {
+
+/*
+ * This class creates containers.
+ */
+
+class SoftwareContainerFactory {
+public:
+    virtual ~SoftwareContainerFactory() {};
```
breakreturn: No need for the ;

agent/unit-test/softwarecontaineragent_unittest.cpp
```diff
@@ -0,0 +1,213 @@
...
+
+    }
+
+    MOCK_METHOD2(doCreateHelper, void(ContainerID containerID, const SoftwareContainerConfig *config));
+    void createHelper(ContainerID containerID, std::unique_ptr<const SoftwareContainerConfig> config) {
+        doCreateHelper(containerID, config.get());
+    }
+};
+
+*/
```
breakreturn: I guess this can be removed?

agent/unit-test/softwarecontaineragent_unittest.cpp
```diff
@@ -0,0 +1,213 @@
...
+namespace softwarecontainer {
+class TestContainerInterface : public SoftwareContainerAbstractInterface
+{
+    LOG_DECLARE_CLASS_CONTEXT("SCAU", "SoftwareContainerAgent Unit Test");
+public:
+    TestContainerInterface() {};
```
breakreturn: No need for the extra ;

agent/unit-test/softwarecontaineragent_unittest.cpp
```diff
@@ -0,0 +1,213 @@
...
+};
+
+class TestFactory: public SoftwareContainerFactory
+{
+public:
+    TestFactory(std::shared_ptr<SoftwareContainerAbstractInterface> testContainerInterface) :
+        m_container(testContainerInterface)
+    { }
+
+    std::shared_ptr<SoftwareContainerAbstractInterface> createContainer(const ContainerID id __attribute__((unused)),
```
breakreturn, line 96: Will __attribute __() work with clang?

tobsan: I generally don't like the `__attribute__` attributes, I would rather have a compiler warning actually.

agent/unit-test/softwarecontaineragent_unittest.cpp
```diff
@@ -0,0 +1,213 @@
...
+    ASSERT_NO_THROW({
+        auto id = sca->createContainer(valid_config);
+        EXPECT_CALL(*testContainerInterface, suspend());
+        sca->suspendContainer(id);
+        EXPECT_CALL(*testContainerInterface, resume());
+        sca->resumeContainer(id);
+
+        ASSERT_EQ(testContainerInterface, sca->getContainer(id));
+        auto containers = sca->listContainers();
+        ASSERT_EQ(unsigned(1), containers.size());
```
tobsan: You could write this as 1u instead.

tobsan (changes requested):
Mostly minor things. And as I wrote before, I would prefer if we didn't adopt the "SCA" acronym (which goes for commit messages as well).

agent/src/softwarecontainerfactory.cpp
kursatkobya: it should be like this then : 
```
auto container = std::shared_ptr<SoftwareContainerAbstractInterface>(
                           new SoftwareContainer(id, std::move(config)));
```

tobsan (approved):

breakreturn (approved):

PR #345, opened 2017-02-10
Replace ReturnCode with bool

thenor:
Remove all remaining occurrences of ReturnCode and the definitions of the same.

### DIFF
--- a/libsoftwarecontainer/component-test/networkgateway_componenttest.cpp
+++ b/libsoftwarecontainer/component-test/networkgateway_componenttest.cpp
@@ -46,6 +46,8 @@ protected:
         srand(time(NULL));
         uint32_t containerID = rand() % 100;
 
+        ::testing::DefaultValue<bool>::Set(true);
+
         gw = new ::testing::NiceMock<MockNetworkGateway>(containerID);
         SoftwareContainerTest::SetUp();
     }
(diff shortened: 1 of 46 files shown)


thenor:
I have a fix for the build error. Some new isError had been added and I missed them during the rebase. I will wait to fix them in case I get comments.

tobsan (approved):

libsoftwarecontainer/component-test/networkgateway_componenttest.cpp
```diff
@@ -46,6 +46,8 @@ class NetworkGatewayTest : public SoftwareContainerGatewayTest
         srand(time(NULL));
         uint32_t containerID = rand() % 100;
 
+        ::testing::DefaultValue<bool>::Set(true);
```
kursatkobya, line 49: I just wonder why we needed this line ?

kursatkobya (approved):

libsoftwarecontainer/component-test/networkgateway_componenttest.cpp
thenor, line 49: There seems to be a slightly different behavior now that ReturnCode isn't used. I had a failing test case which stopped failing once I added this line, which sets the default return value for mocked functions.

PR #346, opened 2017-02-13
Separate getting-started examples into separate runnable parts

frznlogic:

### DIFF
--- a/doc/chapters/getting-started/examples/run_example.sh
+++ b/doc/chapters/getting-started/examples/run_example.sh
@@ -1,0 +1,15 @@
+#!/bin/sh -xe 
+
+./01_start_sc.sh &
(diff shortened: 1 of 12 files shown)


DunderRoffe (approved):

doc/chapters/getting-started/examples/run_example.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/sh -xe 
+
+./01_start_sc.sh &
+sleep 4
```
tobsan, line 4: Four seconds is a bit much for just starting the agent?

frznlogic: I agree. This part is not finished yet and we need to work on it.

PR #348, opened 2017-02-13
Remove log_ functions from stuff run in container

tobsan:
Currently, due to how DLT works, using ivi-logging with
DLT will disable logging inside containers, and instead
just hang, which is not good. We've cleaned this before,
but missed that it sneaked back in.

### DIFF
--- a/libsoftwarecontainer/src/gateway/network/iptableentry.cpp
+++ b/libsoftwarecontainer/src/gateway/network/iptableentry.cpp
@@ -19,6 +19,7 @@
 
 #include "iptableentry.h"
 #include "gateway/gateway.h"
+#include <iostream>
 
 namespace softwarecontainer {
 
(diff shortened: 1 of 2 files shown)


breakreturn (approved):

libsoftwarecontainer/src/gateway/network/iptableentry.cpp
```diff
@@ -29,20 +30,20 @@ bool IPTableEntry::applyRules()
 
             for (auto proto : rule.protocols) {
                 if (!insertCommand(interpretRuleWithProtocol(rule, proto))) {
-                    log_error() << "Couldn't apply the rule " << rule.target;
+                    std::cout << "Couldn't apply the rule " << rule.target << std::endl;
```
rpannek: Shouldn't you be logging errors into `std::cerr`?

libsoftwarecontainer/src/gateway/network/iptableentry.cpp
```diff
@@ -29,20 +30,20 @@ bool IPTableEntry::applyRules()
...
                     return false;
                 }
             }
 
         } else if (!insertCommand(interpretRule(rule))) {
-            log_error() << "Couldn't apply the rule " << rule.target;
+            std::cout << "Couldn't apply the rule " << rule.target << std::endl;
```
rpannek: Please use std::cerr for error logging.

rpannek (changes requested):

tobsan:
Updated as per your requests @rpannek

jeena (approved):

rpannek (approved):

PR #350, opened 2017-02-13
Service test fixture for creating test output dir

thenor:
Move creation of test output dir to a fixture in conftest.

### DIFF
--- a/servicetest/agent/test_agent.py
+++ b/servicetest/agent/test_agent.py
@@ -29,6 +29,10 @@ CURRENT_DIR = os.path.dirname(os.path.realpath(__file__))
 TESTOUTPUT_DIR = CURRENT_DIR + "/testoutput"
 
 # Provide what the testframework requires #####
+# This function is used by the test framework to know where test specific files should be stored
+def output_dir():
+    return TESTOUTPUT_DIR
+
 
 # This function is used by the 'agent' fixture to know where the log should be stored
 def logfile_path():
@@ -47,18 +51,7 @@ DATA = {
     Container.READONLY: False
 }
 
-@pytest.fixture
-def create_testoutput_dir(scope="module"):
-    """ Create a directory for the generated test files.
-
-        This directory is ignored by git but it's nice to have
-        somewhere locally to store test output to support
-        troubleshooting etc.
-    """
-    if not os.path.exists(TESTOUTPUT_DIR):
-        os.makedirs(TESTOUTPUT_DIR)
-
-# Test suites #
+### Test suites ###
 
 
 @pytest.mark.usefixtures("create_testoutput_dir", "agent")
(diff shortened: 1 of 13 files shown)


servicetest/conftest.py
```diff
@@ -115,3 +115,17 @@ def assert_no_proxy():
...
     assert grep_for_dbus_proxy() != 0, "dbus-proxy is alive when it shouldn't be"
+
+
+@pytest.fixture(scope="module")
+def create_testoutput_dir(request):
+    """ Create a directory for the generated test files.
+
+        This directory is ignored by git but it's nice to have
+        somewhere locally to store test output to support
+        troubleshooting etc.
```
breakreturn: This comment is now a bit off. There is nothing that automatically makes git ignore the directory created by this fixture. Previously it was a comment about the local situation in each test suite directory. Perhaps this comment instead should say that it's recommended by tests to put all their ephemeral test files/output in this dir and have git ignore it. Since that's the intention but it's not necessarily  the case.

breakreturn (changes requested):
Looks good, just a minor comment.

servicetest/agent/test_agent.py
```diff
@@ -29,6 +29,10 @@
 TESTOUTPUT_DIR = CURRENT_DIR + "/testoutput"
 
 # Provide what the testframework requires #####
+# This function is used by the test framewor to know where test specific files should be stored
```
tobsan: typo, should be **framework**

servicetest/conftest.py
```diff
@@ -115,3 +115,17 @@ def assert_no_proxy():
...
+    """
+    testoutput_dir = request.module.testoutput_dir()
+
+    if not os.path.exists(testoutput_dir):
+        os.makedirs(testoutput_dir)
```
tobsan: It would be nice to have some error checking here actually.

tobsan (changes requested):

servicetest/agent/test_agent.py
thenor: And then copy-pasted in all the suites. XD
Will fix.

breakreturn (approved):

servicetest/filesystem/test_filesystem.py
```diff
@@ -41,7 +46,7 @@ def logfile_path():
     Container.READONLY: False
 }
 
-@pytest.mark.usefixtures("dbus_launch", "agent", "assert_no_proxy")
+@pytest.mark.usefixtures("create_testoutput_dir", "dbus_launch", "agent", "assert_no_proxy")
 class TestFileSystem(object):
     """ This suite should do some basic testing of the Filesystem within the
         containers and that the whole chain is working properly.
```
thenor, line 52: This test suite uses "CURRENT_DIR" for the test file (I can't comment on the line, but line 68). Should I take the opportunity to change this?

breakreturn: I think it would be nice if all test suites followed the same pattern w.r.t where the files go and what's in the .gitignore file, so change it if it can be done without any major hassle IMO.

thenor: Ok. Will do.

tobsan (approved):

PR #356, opened 2017-02-15
Fixes some minor issues with the test methodology

frznlogic:

### DIFF
--- a/doc/chapters/getting-started/examples/01_start_sc.sh
+++ b/doc/chapters/getting-started/examples/01_start_sc.sh
@@ -1,1 +1,1 @@
-sudo softwarecontainer-agent 
+sudo -E softwarecontainer-agent
--- a/doc/chapters/getting-started/examples/run_example.sh
+++ b/doc/chapters/getting-started/examples/run_example.sh
@@ -1,7 +1,9 @@
 #!/bin/sh -xe 
 
+eval `dbus-launch --sh-syntax`
+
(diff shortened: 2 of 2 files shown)


doc/chapters/getting-started/examples/09_setcapabilities.sh
```diff
@@ -3,6 +3,5 @@ gdbus call --system \
 --object-path /com/pelagicore/SoftwareContainerAgent \
 --method com.pelagicore.SoftwareContainerAgent.SetCapabilities \
 0 \
-"['com.pelagicore.temperatureservice.gettemperature',
-'com.pelagicore.temperatureservice.settemperature']"
+"['com.pelagicore.temperatureservice.settemperature']"
```
tobsan: So no getTemperature?

frznlogic: Nein!

Seriously though... not sure if it serves any purpose for the example. Do you still want to have it?

tobsan: No, not if nobody uses it

frznlogic: It's an example only, to show how the SetCapabilities call is used inside the Getting Started chapter. The only thing I can imagine it to be good for is as an example on how to pass multiple strings in an array. I will change it  back to end the discussion ;)

PR #358, opened 2017-02-16
Run doc examples from jenkins

frznlogic:

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -125,6 +125,7 @@ node {
 
         stage('Examples') {
             runInVagrant("cd examples && sudo ./run-tests.sh")
+            runInVagrant("cd doc && sudo ./run_examples.sh")
         }
 
         stage('Artifacts') {
--- a/doc/chapters/getting-started/examples/run_example.sh
+++ b/doc/chapters/getting-started/examples/run_example.sh
@@ -1,8 +1,26 @@
 #!/bin/sh -xe 
+# 
+# Prerequisites:
+#  * softwarecontainer is installed together with examples
+#  * temperature service example is installed in
+#    /usr/local/share/softwarecontainer/examples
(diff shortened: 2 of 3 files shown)


doc/chapters/getting-started/examples/run_example.sh
```diff
@@ -2,6 +2,9 @@
 
 eval `dbus-launch --sh-syntax`
 
+sudo cp /usr/local/share/softwarecontainer/examples/temperature/temperature-service.json \
+    /usr/local/etc/softwarecontainer/service-manifest.d
```
tobsan: I would prefer that the example pointed to that directory instead by using -m or -d flags when launching the agent.

Jenkinsfile
```diff
@@ -111,6 +111,10 @@ node {
             runInVagrant("cd servicetest && sudo ./run-tests.sh")
         }
 
+        stage('DocTest') {
```
DunderRoffe: In my mind DocTest refers to a specific type of framework, e.g. https://www.reddit.com/r/cpp/comments/4kibl4/doctest_the_lightest_feature_rich_c_single_header/, which we might adopt in the future. Therefore I think a better name can be found.

DunderRoffe (changes requested):

DunderRoffe (approved):

Jenkinsfile
```diff
@@ -96,6 +91,14 @@ node {
             runInVagrant("cd build && sudo make install")
         }
 
+        // We build the docs first because they are part of ALL
+        stage('User documentation') {
+            runInVagrant("cd build && make user-doc")
```
tobsan: Buliding the user docs after building the ALL target makes no sense.

tobsan (changes requested):

tobsan (approved):

PR #359, opened 2017-02-16
bind-mounting creates paths inside the container

tobsan:
Since various paths inside the container might be mounted over with a
tmpfs (for example /dev), we can't check the path on the rootfs in the
host, we need to do that inside the container.

If anyone wonders why there is a new way of creating parent directories, it is to avoid issues with DLT, since the utilities we have in the file toolkit uses logging macros. Once the DLT issue is fixed that can be removed.

### DIFF
--- a/libsoftwarecontainer/src/container.cpp
+++ b/libsoftwarecontainer/src/container.cpp
@@ -315,6 +315,11 @@ int Container::unlimitCoreDump()
     return 0;
 }
 
+bool Container::setCgroupItem(std::string subsys, std::string value)
+{
+    return m_container->set_cgroup_item(m_container, subsys.c_str(), value.c_str());
+}
+
 int Container::executeInContainerEntryFunction(void *param)
 {
     int canCoreDump = unlimitCoreDump();
@@ -326,9 +331,18 @@ int Container::executeInContainerEntryFunction(void *param)
     return (*function)();
 }
 
-bool Container::setCgroupItem(std::string subsys, std::string value)
+bool Container::executeSync(ExecFunction function,
+                            pid_t *pid,
+                            const EnvironmentVariables &variables,
+                            int stdin_var,
+                            int stdout_var,
+                            int stderr_var)
 {
-    return m_container->set_cgroup_item(m_container, subsys.c_str(), value.c_str());
+    bool execResult = execute(function, pid, variables, stdin_var, stdout_var, stderr_var);
+
+    // Make sure the container execution finishes and was successful
+    int status = waitForProcessTermination(*pid);
+    return execResult && (0 == status);
 }
 
 bool Container::execute(ExecFunction function,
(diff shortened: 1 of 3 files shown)


libsoftwarecontainer/src/container.cpp
```diff
@@ -326,9 +331,18 @@ int Container::executeInContainerEntryFunction(void *param)
...
+                            int stdin_var,
+                            int stdout_var,
+                            int stderr_var)
 {
-    return m_container->set_cgroup_item(m_container, subsys.c_str(), value.c_str());
+    bool execResult = execute(function, pid, variables, stdin_var, stdout_var, stderr_var);
+
+    // Make sure the container execution finishes and was successful
+    int status = waitForProcessTermination(*pid);
+    return execResult && status == 0;
```
kursatkobya: I would prefer `execResult && (0 == status)`

libsoftwarecontainer/src/container.cpp
```diff
@@ -551,82 +565,72 @@ bool Container::bindMountInContainer(const std::string &pathInHost,
...
+    // Check that there is nothing already mounted on the target path
     pid_t pid = INVALID_PID;
-    bool checkIfAlreadyMountpoint = execute([pathInContainer] () {
+    bool checkIfAlreadyMountpoint = executeSync([pathInContainer] () {
         FILE *mountsfile = setmntent("/proc/mounts", "r");
         struct mntent *mounts;
         while ((mounts = getmntent(mountsfile)) != nullptr) {
             std::string mountDir(mounts->mnt_dir);
-            if ((mountDir).compare(pathInContainer) == 0) {
+            if (mountDir.compare(pathInContainer) == 0) {
```
kursatkobya: `0 == mountDir.compare(pathInContainer)` would be better

libsoftwarecontainer/src/container.cpp
```diff
@@ -640,40 +644,74 @@ bool Container::bindMountCore(const std::string &pathInHost,
...
-    std::string pathInContainerOnHost = buildPath(rootFS(), pathInContainer);
+    pid_t pid = INVALID_PID;
+    int status = -1;
+
+    // Utility lambda function to create parent paths of a given path
+    auto createParentDirectories = [this] (std::string path) {
+        std::vector<std::string> paths;
+        std::string currentPath = path;
+        while(true) {
+            if (isDirectory(currentPath)) {
```
DunderRoffe: `while(true)` -> `while(!isDirectory(currentPath))` would be neater

libsoftwarecontainer/src/container.cpp
```diff
@@ -640,40 +644,74 @@ bool Container::bindMountCore(const std::string &pathInHost,
...
+                break;
+            } else {
+                paths.push_back(currentPath);
+                currentPath = parentPath(currentPath);
+            }
+        }
+        for (auto it = paths.rbegin(); it != paths.rend(); it++) {
+            std::string path = *it;
```
DunderRoffe: Would be nice with a C++11 for loop instead:
```cpp
for (std::string path : paths) {
   ...
}
```

DunderRoffe (changes requested):

libsoftwarecontainer/src/container.cpp
```diff
@@ -640,40 +644,74 @@ bool Container::bindMountCore(const std::string &pathInHost,
     std::string filePart = baseName(pathInContainer);
     std::string tempDirInContainer = buildPath(gatewaysDirInContainer(), filePart);
 
-    // The same paths on the host
-    std::string pathInContainerOnHost = buildPath(rootFS(), pathInContainer);
+    pid_t pid = INVALID_PID;
+    int status = -1;
+
+    // Utility lambda function to create parent paths of a given path
+    auto createParentDirectories = [this] (std::string path) {
```
kursatkobya: I would prefer a member function instead of lambda here.

tobsan: Hmmm yes perhaps. But this is only ever used here. I'm open to moving it though.

libsoftwarecontainer/src/container.cpp
tobsan: That would be nice yes, but i want to loop backwards through the vector.

libsoftwarecontainer/src/container.cpp
kursatkobya: I just don't think it is simple and concise enough. But it is just preference; if you feel this is better then its okay.

DunderRoffe: There is a very similar function in `common/filetoolkitwithundo.cpp`. Consider using that or replacing it with this function. No need for multiple versions :)

libsoftwarecontainer/src/container.cpp
```diff
@@ -640,40 +644,73 @@ bool Container::bindMountCore(const std::string &pathInHost,
...
         //
         // First, optionally create any parent directories to the targets
         // We do this on the host to avoid working too much inside the container
         //
+        std::string parentPathInContainer = parentPath(pathInContainer);
+        ExecFunction createParent = std::bind(createParentDirectories, parentPathInContainer);
+
+        if (!executeSync(createParent, &pid)) {
+            log_error() << "Could not create parent directory " << parentPathInContainer
+                        << " in container, status: " << status;
```
dzudek: Where the `status` get populated? Looks like you're going to return -1 always (line 648).

libsoftwarecontainer/src/container.cpp
tobsan: We can't use the method from FileToolKitWithUndo, because that uses logging, and until the DLT is fixed to work well with pid namespaces we can't log with that inside containers.

libsoftwarecontainer/src/container.cpp
tobsan: Good catch, when using `executeSync` the status variable is moved inside that function. Will update.

DunderRoffe (approved):

PR #365, opened 2017-02-17
component-test: use isSuccess/isError in jobs

tobsan:
Also move definition of success/error exit codes into test class

### DIFF
--- a/libsoftwarecontainer/component-test/softwarecontainerlib_componenttest.cpp
+++ b/libsoftwarecontainer/component-test/softwarecontainerlib_componenttest.cpp
@@ -36,10 +36,14 @@
 
 LOG_DECLARE_DEFAULT_CONTEXT(defaultContext, "ff", "dd");
 
+static constexpr const int SUCCESS = 0;
+static constexpr const int FAILURE = 1;
+
 class SoftwareContainerApp : public SoftwareContainerLibTest
 {
 
 public:
+
     void startGateways(const std::string &config, const std::string &gatewayID)
     {
         json_error_t error;
@@ -107,7 +111,6 @@ TEST_F(SoftwareContainerApp, TestWaylandWhitelist) {
     westonJob->start();
     westonJob->wait();
     ASSERT_TRUE(westonJob->isSuccess());
-
 }
 
 /*
@@ -137,9 +140,6 @@ TEST_F(SoftwareContainerApp, EnvVarsSet) {
     ASSERT_TRUE(job->isSuccess());
 }
 
-static constexpr int EXISTENT = 0;
-static constexpr int NON_EXISTENT = 1;
-
 TEST_F(SoftwareContainerApp, FileGatewayReadOnly) {
 
     std::string tempFilename1 = createTempFile();
@@ -147,10 +147,11 @@ TEST_F(SoftwareContainerApp, FileGatewayReadOnly) {
     // They will be mapped to these files, which should not yet exist
     std::string containerPath1 = "/tmp/testFile1";
     auto jobMounted = getSc().createFunctionJob([&] () {
-        return isFile(containerPath1) ? EXISTENT : NON_EXISTENT;
+        return isFile(containerPath1) ? SUCCESS : FAILURE;
     });
     jobMounted->start();
-    ASSERT_EQ(jobMounted->wait(), NON_EXISTENT);
+    jobMounted->wait();
+    ASSERT_TRUE(jobMounted->isError());
 
     // Let's configure the env gateway
     std::string configStr =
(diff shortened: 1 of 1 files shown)


thenor (approved):

libsoftwarecontainer/component-test/softwarecontainerlib_componenttest.cpp
```diff
@@ -487,7 +501,7 @@ TEST_F(SoftwareContainerApp, TestUnixSocket) {
 
                 close(fd2);
 
-                return done == 1 ? EXISTENT : NON_EXISTENT;
+                return done == 1 ? SUCCESS : FAILURE;
```
kursatkobya, line 504: it would be better to map numbers to constexpr as well. 
Also `CONSTEXPRDONE == done` is prefferable

kursatkobya (commented):
Looks good with a comment :
I would prefer defined values instead of some numbers

libsoftwarecontainer/component-test/softwarecontainerlib_componenttest.cpp
```diff
@@ -508,28 +522,29 @@ TEST_F(SoftwareContainerApp, TestUnixSocket) {
         sleep(1);
 
         if ((s = socket(AF_UNIX, SOCK_STREAM, 0)) == -1) {
```
kursatkobya, line 524: i would prefer defined value instead of 

> -1

libsoftwarecontainer/component-test/softwarecontainerlib_componenttest.cpp
```diff
@@ -508,28 +522,29 @@ TEST_F(SoftwareContainerApp, TestUnixSocket) {
...
-            _exit(NON_EXISTENT);
+            _exit(FAILURE);
         }
 
         remote.sun_family = AF_UNIX;
         strncpy(remote.sun_path, tempUnixSocket, strlen(tempUnixSocket) + 1);
         len = strlen(remote.sun_path) + sizeof(remote.sun_family);
         if (connect(s, (struct sockaddr *)&remote, len) == -1) {
```
kursatkobya, line 531: would prefer defined value instead of

libsoftwarecontainer/component-test/softwarecontainerlib_componenttest.cpp
```diff
@@ -508,28 +522,29 @@ TEST_F(SoftwareContainerApp, TestUnixSocket) {
...
-            _exit(NON_EXISTENT);
+            _exit(FAILURE);
         }
 
 
         if (send(s, str, strlen(str), 0) == -1) {
```
kursatkobya, line 536: would prefer defined value instead of -1
also `definedvalue == send(s, str, strlen(str), 0) `would be better

libsoftwarecontainer/component-test/softwarecontainerlib_componenttest.cpp
DunderRoffe: Would not just returning done be enough in this case? If not then the job should be rewritten in such a way that it does.

libsoftwarecontainer/component-test/softwarecontainerlib_componenttest.cpp
```diff
@@ -508,28 +522,29 @@ TEST_F(SoftwareContainerApp, TestUnixSocket) {
         sleep(1);
```
DunderRoffe, line 522: Add comment explaining why there is a sleep here. If possible it would be nice to swap it for a `join(pid)` or something along those lines.

DunderRoffe (commented):
More nitpicking

libsoftwarecontainer/component-test/softwarecontainerlib_componenttest.cpp
tobsan: I would prefer not to touch the tests as such in this PR, actually, but in general I agree.

DunderRoffe: sure

libsoftwarecontainer/component-test/softwarecontainerlib_componenttest.cpp
tobsan: Imo it makes more sense to return known values than relying on that they will match what we get as value for `done`

tobsan:
@kursatkobya and @DunderRoffe: while I would like to use defined values instead of magic ones like -1 and 0, they are not always the same for all calls, so one would have to (almost) define each one.

Plus the point of this was not to change the use of constant values, but if you really insist I can of course update.

kursatkobya (approved):
If you will this is better way then i am okay with that decision

PR #366, opened 2017-02-20
Test setting capability twice

thenor:
Verify that there are no crashes when setting the same
capability twice. Write a service test to verify this.

### DIFF
--- a/servicetest/capabilities/test_caps.py
+++ b/servicetest/capabilities/test_caps.py
@@ -188,7 +188,7 @@ class TestCaps(object):
         try:
             sc.start(DATA)
             with pytest.raises(DBusException) as err:
-               sc.set_capabilities(["test.cap.broken-gw-config"])
+                sc.set_capabilities(["test.cap.broken-gw-config"])
 
         finally:
             sc.terminate()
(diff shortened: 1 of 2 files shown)


servicetest/capabilities/test_caps.py
```diff
@@ -230,3 +230,17 @@ def test_defaults_are_applied(self):
...
+        """ Test that setting the same capability doesn't result in crashes
+        """
+        sc = Container()
+        try:
+            sc.start(DATA)
+            sc.set_capabilities(["test.cap.valid-dbus"])
+
+            with pytest.raises(DBusException) as err:
+                # Set the same capability again
+                sc.set_capabilities(["test.cap.valid-dbus"])
```
DunderRoffe, line 246: Does this really cover the testcase? We should make sure that `err` does not contain the message "NoReply"

DunderRoffe (changes requested):

servicetest/capabilities/test_caps.py
thenor: Oops. Yes. Will fix.

servicetest/capabilities/test_caps.py
```diff
@@ -230,3 +230,17 @@ def test_defaults_are_applied(self):
             assert my_env_var == "TEST_ENV_VALUE"
         finally:
             sc.terminate()
+
+    def test_set_same_cap_twice(self):
+        """ Test that setting the same capability doesn't result in crashes
```
breakreturn: Maybe we can be even a little more clear here, and describe that we are expecting a D-Bus exception of a certain kind and not a crash which results in a timeout error on this end (the client).

breakreturn (changes requested):

servicetest/capabilities/test_caps.py
thenor: Alright. I will update it to be more specific. Just as the comment for the error I expect.

DunderRoffe (approved):

servicetest/capabilities/test_caps.py
```diff
@@ -20,6 +20,7 @@
 
 import os
 import time
+import unittest
```
breakreturn: This could be removed now I guess?

thenor: Yes! Forgot it! Thanks!

breakreturn (approved):

PR #367, opened 2017-02-21
Check exception type

thenor:
In service tests, when expecting and exception, verify
that the exception type is a failure, not a crash/NoReply.

### DIFF
--- a/servicetest/capabilities/test_caps.py
+++ b/servicetest/capabilities/test_caps.py
@@ -173,6 +173,7 @@ class TestCaps(object):
             sc.start(DATA)
             with pytest.raises(DBusException) as err:
                 sc.set_capabilities(["does.not.exist"])
+            assert err.value.get_dbus_name() == Container.DBUS_EXCEPTION_FAILED
 
         finally:
             sc.terminate()
@@ -186,6 +187,7 @@ class TestCaps(object):
             sc.start(DATA)
             with pytest.raises(DBusException) as err:
                sc.set_capabilities(["test.cap"])
+            assert err.value.get_dbus_name() == Container.DBUS_EXCEPTION_FAILED
 
         finally:
             sc.terminate()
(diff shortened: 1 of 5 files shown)


breakreturn (approved):

DunderRoffe (approved):

DunderRoffe:
Looking at the pytest doc it seems that the assert of message can be set as
 ```python
with pytest.raises (DBusException, message="organisation.freedesktop...")
  do some crazy stuff
```

**Comment by tobsan** [comment]
Is it just me, or does the indentation of the asserts feel off?

**Comment by thenor** [comment]
I understand your point, @tobsan. It's not possible to have it in the same scope as the "with", since the command which results in the exception triggers the clause to terminate.

**Comment by breakreturn** [comment]
"with" does not create a new scope. I think this is how you'd do this.

**Comment by thenor** [comment]
@DunderRoffe After having tested this it seems it is not possible. So keeping it as is.

File: servicetest/capabilities/test_caps.py [hunk 1]
```diff
@@ -245,7 +248,7 @@ def test_set_same_cap_twice(self, agent):
                 # Set the same capability again
                 sc.set_capabilities(["test.cap.valid-dbus"])
             # The expected failure is 'Failed', not 'NoReply'
-            assert err.value.get_dbus_name() == "org.freedesktop.DBus.Error.Failed"
+            assert err.value.get_dbus_name() == sc.DBUS_EXCEPTION_FAILED
```
breakreturn: I think all of these are better accessed like "Container.DBUS_EXCEPTION_FAILED" to make it clear that it's intended to be a "constant" on the class, not the object.

thenor: Alright. Will fix it.

servicetest/testframework/lib.py
```diff
@@ -160,6 +160,9 @@ class Container():
     HOST_PATH = "host-path"
     READONLY = False
 
+    # Other static class members used by tests
```
breakreturn: I think it would be better to comment what this is for.

PR #368, opened 2017-02-21
Move from debian to ubuntu

frznlogic:
This is necessary to be able to test the overlayfs used in the writebuffers

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -85,16 +85,19 @@ Vagrant.configure(2) do |config|
 
     # Install dependencies via git
     config.vm.provision "shell", privileged: false,
-        args: ["dlt-daemon", "https://github.com/GENIVI/dlt-daemon.git"],
+        env: {"SRC_DIR" => "dlt-daemon", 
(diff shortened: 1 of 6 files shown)


DunderRoffe (approved):

breakreturn (approved):

Vagrantfile
```diff
@@ -56,7 +56,7 @@ Vagrant.configure(2) do |config|
     config.vm.provision "shell", inline: "apt-get update"
 
     # Install build dependencies
-    config.vm.provision "shell", path: "cookbook/system-config/setup-ccache-and-icecc.sh"
+#    config.vm.provision "shell", path: "cookbook/system-config/setup-ccache-and-icecc.sh"
```
thenor, line 59: Why not remove this line, it is commented out.

frznlogic: I think we might want to enable it fairly soon again if we can get icecc stable. This is a bit  unrelated to the story tbh, but icecc has become unreliable after the changes to the network for some reason.

thenor: Alright, might be best to add a comment, since there is a reason, but perhaps a bit too cumbersome to make an update only for this.

PR #369, opened 2017-02-22
Adding support for config file in the service tests

tobsan:
See the individual commits for more info.

### DIFF
--- a/agent/CMakeLists.txt
+++ b/agent/CMakeLists.txt
@@ -72,7 +72,17 @@ if(ENABLE_NETWORKGATEWAY)
 endif()
 
 # Configure and install the general config file
-configure_file(softwarecontainer-config.in softwarecontainer-config @ONLY)
+configure_file(${CMAKE_CURRENT_SOURCE_DIR}/softwarecontainer-config.in
+               ${CMAKE_CURRENT_BINARY_DIR}/softwarecontainer-config
+               @ONLY)
+
+# Copy the configuration to the service test directory, so that we can use that
+# as a base for when we test with different configs.
+configure_file(${CMAKE_CURRENT_BINARY_DIR}/softwarecontainer-config
+               ${CMAKE_SOURCE_DIR}/servicetest/testframework/softwarecontainer-config
+               @ONLY)
+
+# Install the configuration file
 install(FILES "${CMAKE_CURRENT_BINARY_DIR}/softwarecontainer-config" DESTINATION ${SYS_CONFIG_DIR})
 
 #
--- a/servicetest/capabilities/test_caps.py
+++ b/servicetest/capabilities/test_caps.py
@@ -26,6 +26,7 @@ from testframework import Container
 from testframework import Capability
 from testframework import StandardManifest
 from testframework import DefaultManifest
+from testframework import ConfigFile
 
 from dbus.exceptions import DBusException
 
@@ -118,6 +119,15 @@ def service_manifests():
 def mounted_path_in_host():
     return CURRENT_DIR
 
+def agent_config():
+    """ The agent fixture calls this function when it creates the config file to be
+        used in this module. It expects a ConfigFile object.
+    """
+    return ConfigFile(TESTOUTPUT_DIR + "/agent-config",
+            {
+                "SoftwareContainer": {"use-session-bus": "true"}
+            }
+    )
 
 @pytest.mark.usefixtures("testhelper", "dbus_launch", "create_testoutput_dir", "agent", "assert_no_proxy")
 class TestCaps(object):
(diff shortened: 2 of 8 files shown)


agent/CMakeLists.txt
```diff
@@ -72,7 +72,14 @@ if(ENABLE_NETWORKGATEWAY)
 endif()
 
 # Configure and install the general config file
-configure_file(softwarecontainer-config.in softwarecontainer-config @ONLY)
+configure_file(${CMAKE_CURRENT_SOURCE_DIR}/softwarecontainer-config.in
+               ${CMAKE_CURRENT_BINARY_DIR}/softwarecontainer-config
+               @ONLY)
+# Copy the configuration to the service test directory
```
breakreturn: I think this comment should say why we do this, the actual configure/move of file is already visible in the actual code, but it's not easy to understand why.

servicetest/capabilities/test_caps.py
```diff
@@ -118,6 +119,15 @@ def service_manifests():
 def mounted_path_in_host():
     return CURRENT_DIR
 
+def agent_config():
+    """ The agent fixture calls this function when it creates the config file to be
+        used in this module. It expects a ConfigFile object.
+    """
+    return ConfigFile(TESTOUTPUT_DIR + "/agent-config",
+            {
+                "SoftwareContainer": { "bridge-device": "lxcbr5" }
```
breakreturn: It's unusual in our code to have spaces after braces.

servicetest/testframework/lib.py
```diff
@@ -32,9 +32,105 @@
...
+        })
+    """
+
+    def __init__(self, location, full_config):
+        """ Create a config object from a given dict of dicts.
+        """
+        self.__filename = location
+        self.__groups = full_config
+
+    def filename(self):
```
breakreturn: This actually returns the full path including filename right? A name change might be appropriate in that case.

servicetest/testframework/lib.py
```diff
@@ -32,9 +32,105 @@
...
+        result = ""
+        for group in self.__groups:
+            result += "[" + group + "]\n"
+            for item in self.__groups[group]:
+                result += item + " = " + self.__groups[group][item] + "\n"
+
+        return result
+
+    def __default_config(self):
+        """ Read the softwarecontainer-config prepared by cmake, and return its values as a digt.
```
breakreturn: "you're a digt!"

servicetest/testframework/lib.py
```diff
@@ -32,9 +32,105 @@
...
+        group = None
+        config_path = os.path.join(os.path.dirname(__file__), "softwarecontainer-config")
+        with open(config_path, "r") as fh:
+            while True:
+                line = fh.readline()
+                if not line:
+                    return conf
+                else:
+                    line = line.strip()
+                    if len(line) == 0 or line[0] == '\n' or line[0] == '#':
```
breakreturn: Maybe add a short comment here what this means, i.e. what are we ignoring? Empty lines and newlines-only are easy to guess why, but it's not obvious that the '#' is related to the top comment about what's extracted or not.

Like: "Skip empty lines and comments in the file"

servicetest/testframework/lib.py
```diff
@@ -32,9 +32,105 @@
...
+            The file, and the dict, will only contain values that are mandatory options, and as
+            such has no default values within SoftwareContainer. These are the values that one
+            has to have in a config file for the agent to start at all. Any other config values
+            have reasonable defaults and are therefore commented out in the config file.
+        """
+        conf = dict()
+        group = None
+        config_path = os.path.join(os.path.dirname(__file__), "softwarecontainer-config")
+        with open(config_path, "r") as fh:
+            while True:
```
breakreturn: There might be a reason for this loop construct, but maybe it would be easier to read if we just read the complete content once, split it by newline and then process the strings in that list?

breakreturn (changes requested):
Looks good. Some comments and questions.

kursatkobya (approved):
Couldn't look in it deeply but it looks nice

tobsan:
Updated after comments from @breakreturn

servicetest/testframework/lib.py
```diff
@@ -32,9 +32,110 @@
...
+
+            The file, and the dict, will only contain values that are mandatory options, and as
+            such has no default values within SoftwareContainer. These are the values that one
+            has to have in a config file for the agent to start at all. Any other config values
+            have reasonable defaults and are therefore commented out in the config file.
+        """
+        config_path = os.path.join(os.path.dirname(__file__), "softwarecontainer-config")
+
+        # Open the file, read it and close it
+        fh = open(config_path, "r")
```
breakreturn: Nitpicking here so do as you please, but it's more common in our code to use "with". I guess you want to avoid it because it looks weird to use the variable file_contents later outside of the "with"? It's perfectly OK afaik in Python since there is no new scope in a "with".

PR #372, opened 2017-02-28
Gateway base class now supports quick-launch

breakreturn:
Unit tests are updated to reflect this. Note that
no actual gateways are modified to support this
yet.

### DIFF
--- a/libsoftwarecontainer/component-test/softwarecontainerlib_componenttest.cpp
+++ b/libsoftwarecontainer/component-test/softwarecontainerlib_componenttest.cpp
@@ -22,6 +22,7 @@
 #include <sys/un.h>
 #include <sys/mman.h>
 #include <stdlib.h>
+#include <cassert>
 
 #include "softwarecontainer_test.h"
 
@@ -44,18 +45,20 @@ class SoftwareContainerApp : public SoftwareContainerLibTest
 
 public:
 
-    void startGateways(const std::string &config, const std::string &gatewayID)
+    bool startGateways(const std::string &config, const std::string &gatewayID)
     {
         json_error_t error;
         json_t *configJson = json_loads(config.c_str(), 0, &error);
-        ASSERT_FALSE(configJson == nullptr);
+        assert(configJson != nullptr);
 
         GatewayConfiguration gwConf;
         gwConf.append(gatewayID, configJson);
 
-        getSc().startGateways(gwConf);
+        bool result = getSc().startGateways(gwConf);
 
         json_decref(configJson);
+
+        return result;
     }
 
     bool bindMountInContainer(const std::string src, const std::string dst, bool readOnly)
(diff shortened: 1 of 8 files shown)


thenor (approved):

libsoftwarecontainer/src/gateway/gateway.cpp
```diff
@@ -21,11 +21,30 @@
 
 namespace softwarecontainer {
 
+
+Gateway::Gateway(const std::string &id,
+                 std::shared_ptr<ContainerAbstractInterface> container,
+                 bool quickLaunchEnabled) :
+    m_id(id),
+    m_container(container),
+    m_quickLaunchEnabled(quickLaunchEnabled),
```
DunderRoffe: I think `reconfigurable` better describes the meaning of the member

libsoftwarecontainer/unit-test/gateway_unittest.cpp
```diff
@@ -28,7 +28,21 @@ using namespace softwarecontainer;
...
+    MOCK_METHOD1(readConfigElement, bool(const json_t *element));
+    MOCK_METHOD0(activateGateway, bool());
+    MOCK_METHOD0(teardownGateway, bool());
+};
+
+class MockGatewayQuickLaunchEnabled : public Gateway
+{
+public:
+    MockGatewayQuickLaunchEnabled() :
+        Gateway(std::string("quick-launch-mock"), nullptr, true)
```
DunderRoffe: As mentioned above, I think `reconfigure` is better than `quick-launch`.

libsoftwarecontainer/unit-test/gateway_unittest.cpp
```diff
@@ -186,16 +202,52 @@ TEST_F(GatewayTest, ConfigEnablesActivateEnablesTeardown) {
...
+/*
+ * Test that double configuration and activation is not possible for
+ * gatways that are not quick-launch enabled.
+ */
+TEST_F(GatewayTest, CanNotConfigureAndActivateTwiceWhenNotQuickLaunchEnabled) {
+    loadConfig(validConf);
+    ASSERT_TRUE(gw.setConfig(jsonConf));
+    ASSERT_TRUE(gw.activate());
+    ASSERT_THROW(gw.setConfig(jsonConf), GatewayError);
+    ASSERT_THROW(gw.activate(), GatewayError);
```
DunderRoffe, line 235: The fourth assert here is a bit redundant since the state is not changed. It becomes very similar to `CanNotActivateTwiceWhenNotQuickLaunchEnabled`

DunderRoffe (changes requested):
Overall I like the solution but I found a few nitpicks.

libsoftwarecontainer/src/gateway/gateway.cpp
```diff
@@ -21,11 +21,30 @@
...
+{
+    return m_id;
+}
+
 bool Gateway::setConfig(const json_t *config)
 {
-    if (m_state == GatewayState::ACTIVATED) {
-        log_error() << "Can not configure a gateway that is already activated: " << id();
-        throw GatewayError("Gateway already activated");
+    if (m_state == GatewayState::ACTIVATED && !m_quickLaunchEnabled) {
```
kursatkobya: `GatewayState::ACTIVATED == m_state` would be better presentation

libsoftwarecontainer/src/gateway/gateway.cpp
breakreturn: I thought about this. The concept of quick-launch is a bit abstract for the Gateway interface, I agree with that. But the interface also says that one can configure and activate, and this change is about re-configuring and re-activating, so calling it "reconfigurable" will not match up well to the interface. Do you have a suggestion for a name that covers this?

libsoftwarecontainer/src/gateway/gateway.cpp
breakreturn: To reduce the risk of an accidental assignment? In principle I agree, it's just that the whole code base (basically) does it like this. I didn't plan to go through the gateway code in hunt for this type of improvement. Do you think we could perhaps have that as a separate change?

libsoftwarecontainer/unit-test/gateway_unittest.cpp
```diff
@@ -186,16 +202,52 @@ TEST_F(GatewayTest, ConfigEnablesActivateEnablesTeardown) {
...
     ASSERT_TRUE(gw.setConfig(jsonConf));
     ASSERT_TRUE(gw.activate());
     ASSERT_THROW(gw.activate(), GatewayError);
 }
 
 /*
+ * Test that double activation is not possible for gatways that are
+ * quick-launch enabled.
+ */
+TEST_F(GatewayTest, CanNotActivateTwiceWhenIsQuickLaunchEnabled) {
```
kursatkobya: This test name is not explanatory enough imo. It sounds like we cannot activate this twice because quicklaunch is enabled. But actually we cannot activate it because we did not configure in between.

libsoftwarecontainer/unit-test/gateway_unittest.cpp
kursatkobya: I think test is trying to express even if you configure the gateway after activation you cannot activate again if its QuickLaunchEnabled flag is not set. 

In this mindset i assume the test actually wanted to show activate will throw an exception.

kursatkobya (approved):
Nice work. Approved with minor comments.

libsoftwarecontainer/unit-test/gateway_unittest.cpp
breakreturn: But the two tests do test two different behaviors. I had help by this test when working with the changes, so I thought it would be good to have to support future refactoring or changes in behavior. Just testing that activate fails does not define the behavior to the same extent that also testing activate fails even after a configure.

libsoftwarecontainer/src/gateway/gateway.cpp
kursatkobya: I was thinking that it could make sense to write like this when writing new code and to change when we touch a code particle. Yet it is not a mandatory or so important thing.

DunderRoffe (approved):

libsoftwarecontainer/component-test/softwarecontainerlib_componenttest.cpp
```diff
@@ -44,18 +45,20 @@ class SoftwareContainerApp : public SoftwareContainerLibTest
 
 public:
 
-    void startGateways(const std::string &config, const std::string &gatewayID)
+    bool startGateways(const std::string &config, const std::string &gatewayID)
     {
         json_error_t error;
         json_t *configJson = json_loads(config.c_str(), 0, &error);
-        ASSERT_FALSE(configJson == nullptr);
+        assert(configJson != nullptr);
```
kursatkobya, line 52: doesn't ASSERT_NE working here ? As far as i remember we used to use that function instead of this

breakreturn: This change is because gtest doesn't seem to handle ASSERT_* macros outside of a TEST_* macro very well. And also, I don't think we want to test that the config is valid etc., since that's not the point of the tests (the tests contain the config themselves), we just want to assert that the test has the required prerequisites to work correctly.

PR #374, opened 2017-03-03
User docs check CPU usage

thenor:
Add troubleshooting info on how to check the CPU usage of a process
within a container.

### DIFF
--- a/doc/chapters/api/index.rst
+++ b/doc/chapters/api/index.rst
@@ -22,7 +22,7 @@ Error Handling
 All methods in this API can and will return D-Bus errors on failure. All calls
 that don't return a D-Bus error have worked successfully. Any call that returns
 a D-Bus error has failed. All calls that change a state and raise a D-Bus error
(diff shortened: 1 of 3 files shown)


doc/chapters/api/index.rst
```diff
@@ -80,7 +82,7 @@ Launches the specified application/code in the container.
 |
 
 :Return Values:
-        :pid: ``int32`` PID of the process run inside the container.
+        :pid: ``int32`` the external PID of the process run inside the container.
```
breakreturn: Maybe "PID of process running in the container, as seen by the host" or something like that? "External" might be confusing for some readers I think.

breakreturn (changes requested):

doc/chapters/api/index.rst
thenor: Alright. I read it and felt it needed something. Your suggestion might be even better. I will change it.

breakreturn (approved):

PR #378, opened 2017-03-09
Fix writebuffers and make sure service tests work again. 

frznlogic:


common/filetoolkitwithundo.cpp
```diff
@@ -111,16 +111,20 @@ bool FileToolkitWithUndo::bindMount(const std::string &src,
                                     bool readOnly,
                                     bool enableWriteBuffer)
 {
-    unsigned long flags = MS_BIND;
+    unsigned long flags =  0; //MS_BIND;
```
DunderRoffe: Is this a miss or are there a reason to keep the comment?

common/filetoolkitwithundo.cpp
```diff
@@ -111,16 +111,20 @@ bool FileToolkitWithUndo::bindMount(const std::string &src,
...
 
     if (!existsInFileSystem(src)) {
         log_error() << src << " does not exist on the host, can not bindMount";
         return false;
     }
 
+    if (!existsInFileSystem(dst)) {
+        log_error() << src << " does not exist on the host, can not bindMount";
+        return false;
+    }
```
DunderRoffe: code duplication

common/filetoolkitwithundo.cpp
```diff
@@ -170,15 +178,15 @@ bool FileToolkitWithUndo::overlayMount(const std::string &lower,
                                        const std::string &dst)
 {
     std::string fstype = "overlay";
-    unsigned long flags = MS_BIND;
+    unsigned long flags = 0; //MS_BIND;
```
DunderRoffe: Is this a miss or are there a reason to keep the comment?

common/recursivedelete.cpp
```diff
@@ -0,0 +1,81 @@
...
+#include <ftw.h>
+#include <fstream>
+#include <iostream>
+#include <string>
+
+#include <glibmm.h>
+
+namespace softwarecontainer {
+
+std::string deleteRoot;
```
DunderRoffe: There should be a better solution than using global variables

DunderRoffe (changes requested):

common/filetoolkitwithundo.cpp
frznlogic: Both. I'm removing it.

common/filetoolkitwithundo.cpp
frznlogic: Both. I'm removing it ;).

common/filetoolkitwithundo.cpp
frznlogic: Actually no... code copy, but checking for src in the first test, then dst, but forgot to replace src with dst in the log message. I've fixed that. I hope that is good enough?

common/recursivedelete.cpp
frznlogic: Actually it's dead. Removed it.

common/recursivedelete.h
```diff
@@ -0,0 +1,65 @@
...
+    bool del(std::string dir);
+
+private:
+    RecursiveDelete();
+    ~RecursiveDelete();
+
+    /**
+     * @brief m_deleteLock lock used to protect global variables in the delete() method.
+     */
+    std::mutex m_deleteLock;
```
DunderRoffe: I guess this should be removed as well now that the global variables are removed.

DunderRoffe (changes requested):

DunderRoffe (approved):

common/recursivedelete.h
frznlogic: Done!

breakreturn (approved):
Looks good, didn't comment on what Viktor said about duplication, but approving assuming that will be fixed.

PR #379, opened 2017-03-09
Fixed issue with leaking GW configs

breakreturn:
The problem with GW configs leaking between caps
when using multiple containers is fixed. Relevant
service test is enabled again. Some additional
debug outpuut added as well.

### DIFF
--- a/agent/src/capability/baseconfigstore.cpp
+++ b/agent/src/capability/baseconfigstore.cpp
@@ -105,6 +105,7 @@ void BaseConfigStore::readCapsFromFile(const std::string &filePath)
     }
 
     // Can't use json_decref on fileroot, it removes objects too early
+    // TODO: So when do we actually do decref on this?
     parseCapabilities(capabilities);
 }
 
@@ -142,7 +143,7 @@ void BaseConfigStore::parseCapabilities(json_t *capabilities)
             throw CapabilityParseError(errorMessage);
         }
 
-        log_debug() << "Found capability \"" << capName << "\", parsing gateways...";
+        log_debug() << "Found capability \"" << capName << "\"";
 
         parseGatewayConfigs(capName, gateways);
     }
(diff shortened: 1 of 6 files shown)


DunderRoffe (changes requested):
Looks good but would be nice to keep debug stuff and comments as separate commits

thenor (approved):

common/gatewayconfig.cpp
```diff
@@ -77,8 +88,8 @@ bool GatewayConfiguration::append(const std::string &id, json_t *sourceArray)
         }
         json_decref(backupArray);
     } else {
-        m_configMap[id] = sourceArray;
-        json_incref(sourceArray);
+        m_configMap[id] = json_deep_copy(sourceArray);
+        // we do not need to decref the sourceArray since we did a deep copy
```
frznlogic, line 92: Is this comment is needed for something?

agent/src/capability/baseconfigstore.cpp
```diff
@@ -142,7 +143,7 @@ void BaseConfigStore::parseCapabilities(json_t *capabilities)
             throw CapabilityParseError(errorMessage);
         }
 
-        log_debug() << "Found capability \"" << capName << "\", parsing gateways...";
+        log_debug() << "Found capability \"" << capName << "";
```
kursatkobya: I guess you missed one " at the end
`log_debug() << "Found capability \"" << capName << "\"";`

frznlogic (approved):

breakreturn:
@DunderRoffe I've split up the changes into two commits.

agent/src/capability/baseconfigstore.cpp
breakreturn: Yeah, will fix.

DunderRoffe (approved):

common/gatewayconfig.cpp
breakreturn: It's a comment for why we don't do anything with the source that we got from the caller. The reader be all like "hey, why no code here?", and we be like "cause we copy deep".

PR #385, opened 2017-03-14
Agent: add checks that DBus setup works properly

tobsan:
Makes sure that we can connect to the bus, and that we can
own the softwarecontainer name on that bus. In case of error,
the agent quits.

Also some cleanup in the generated code.

### DIFF
--- a/agent/src/dbus/softwarecontaineragent_dbus_stub.cpp
+++ b/agent/src/dbus/softwarecontaineragent_dbus_stub.cpp
@@ -62,27 +62,34 @@ Glib::ustring interfaceXml0 = R"XML_DELIMITER(
 </node>
 )XML_DELIMITER";
 
-com::pelagicore::SoftwareContainerAgent::SoftwareContainerAgent() : connectionId(0), registeredId(0), m_objectPath("/com/pelagicore/SoftwareContainerAgent"), m_interfaceName("com.pelagicore.SoftwareContainerAgent") {
-
-    ProcessStateChanged_signal.connect(sigc::mem_fun(this, &SoftwareContainerAgent::ProcessStateChanged_emitter));
-
+com::pelagicore::SoftwareContainerAgent::SoftwareContainerAgent() :
+    connectionId(0),
+    registeredId(0),
+    m_objectPath("/com/pelagicore/SoftwareContainerAgent"),
+    m_interfaceName("com.pelagicore.SoftwareContainerAgent")
+{
+    ProcessStateChanged_signal.connect(
+        sigc::mem_fun(this, &SoftwareContainerAgent::ProcessStateChanged_emitter)
+    );
 }
-void com::pelagicore::SoftwareContainerAgent::connect (
+
+void com::pelagicore::SoftwareContainerAgent::connect(
     Gio::DBus::BusType busType,
     std::string name)
 {
     try {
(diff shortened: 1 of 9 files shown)


agent/src/main.cpp
```diff
@@ -216,11 +216,15 @@ int main(int argc, char **argv)
...
-        std::shared_ptr<ContainerUtilityInterface> utility = std::shared_ptr<ContainerUtilityInterface> (new ContainerUtilityInterface());
+        std::shared_ptr<SoftwareContainerFactory> factory =
+            std::shared_ptr<SoftwareContainerFactory> (new SoftwareContainerFactory());
+        std::shared_ptr<ContainerUtilityInterface> utility =
+            std::shared_ptr<ContainerUtilityInterface> (new ContainerUtilityInterface());
 
         ::softwarecontainer::SoftwareContainerAgent agent(mainContext, config, factory, utility);
-        std::unique_ptr<SoftwareContainerAgentAdaptor> adaptor(new SoftwareContainerAgentAdaptor(agent, useSessionBus));
+        std::unique_ptr<SoftwareContainerAgentAdaptor> adaptor(
+            new SoftwareContainerAgentAdaptor(ml, agent, useSessionBus)
```
breakreturn: Maybe there should be a comment what the agent adaptor needs the mainloop for? The reader can ofc look at the implementation but I'd like to see that already here if I read this.

breakreturn (approved):
Minor comment, do as you wish with that. Approving regardless.

agent/src/dbus/softwarecontaineragent_dbus_stub.h
```diff
@@ -91,14 +92,23 @@ class SoftwareContainerAgent {
            const Glib::ustring& property_name,
            const Glib::VariantBase& value);
 
-private:
-bool emitSignal(const std::string& propName, Glib::VariantBase& value);
+    /*
+     * We emit this signal when we have a lost/acquired a name on dbus,
```
thenor: It's a list of three signals so "this" doesn't feel right to me. Just a nitpicky comment, but might as well do it right. ;)

kursatkobya (approved):
good addition with nice format corrections

agent/src/dbus/softwarecontaineragent_dbus_stub.cpp
```diff
@@ -227,51 +242,51 @@ void com::pelagicore::SoftwareContainerAgent::on_method_call(const Glib::RefPtr<
...
+    paramsList.push_back(Glib::Variant<guint32 >::create((exitCode)));;
 
-m_connection->emit_signal(
-              "/com/pelagicore/SoftwareContainerAgent",
-              "com.pelagicore.SoftwareContainerAgent",
-              "ProcessStateChanged",
-              Glib::ustring(),
-              Glib::Variant<std::vector<Glib::VariantBase> >::create_tuple(paramsList));
-      }
+    m_connection->emit_signal(
```
thenor, line 278: Is this a tab?

tobsan: I think my editor converts tabs to spaces for me, plus I think everything was spaces already.

EDIT: That's a no :)

frznlogic (approved):

PR #386, opened 2017-03-14
Enable reconfiguration of the DeviceNode GW

kursatkobya:
Add isConfigured variable to devicenodeparser.h
Add component tests for reconfiguring DeviceNodeGateway


libsoftwarecontainer/src/gateway/devicenode/devicenodegateway.h
```diff
@@ -70,6 +70,13 @@ class DeviceNodeGateway :
      */
     virtual bool teardownGateway() override;
 
+    /**
+     * @brief Questions whether the device is configured or not
+     *
+     * return true or false regarding configuration status
+     */
+    virtual bool isDeviceConfigured(const std::string deviceName);
```
tobsan: Shouldn't this be private instead of public?

libsoftwarecontainer/src/gateway/devicenode/devicenodegateway.cpp
```diff
@@ -51,38 +51,41 @@ bool DeviceNodeGateway::activateGateway()
...
-                log_error() << "Could not 'chmod " << dev.mode
-                            << "' the mounted device " << dev.name;
+            // If the parent directory does not already exist, we create it.
+            if (!isDirectory(deviceParent) && !createDirectory(deviceParent)) {
+                log_error() << "Could not create parent directory for device " << deviceParent;
                 return false;
             }
+
+            // Mount device in container
+            getContainer()->mountDevice(dev.name);
```
tobsan: Shouldn't there be some error checking here?

libsoftwarecontainer/component-test/devicenodegateway_componenttest.cpp
```diff
@@ -97,3 +97,60 @@ TEST_F(DeviceNodeGatewayTest, TestOverwriteDeviceFails) {
...
+            "/dev/tty0",
+            true},
+        testSetup{
+            "[{\"name\": \"/dev/tty1\", \"mode\":  622}]",
+            "[{\"name\": \"/dev/tty1\", \"mode\":  622}]",
+            "/dev/tty1",
+            true}
+));
+
+TEST_P(DynamicConfigurationTests, DynamicDeviceNodeConfiguration) {
```
breakreturn: Needs a comment what the test is about etc.

breakreturn (changes requested):

libsoftwarecontainer/component-test/devicenodegateway_componenttest.cpp
```diff
@@ -97,3 +97,60 @@ TEST_F(DeviceNodeGatewayTest, TestOverwriteDeviceFails) {
...
+
+/*
+ * This data is fed to the DynamicConfiguration tests
+ */
+INSTANTIATE_TEST_CASE_P(DeviceModeConfigurations, DynamicConfigurationTests, ::testing::Values(
+        testSetup{
+            "[{\"name\": \"/dev/tty0\", \"mode\":  622}]",
+            "[{\"name\": \"/dev/tty0\", \"mode\":  755}]",
+            "/dev/tty0",
+            false},
```
breakreturn: For readability, could you add a space before the opening braces and break the line for the closing braces and fix the indentation so the config strings are indented one level and the braces matches on the outer level?

Like:

testSetup {
    "string",
    "string"
},
testSetup {

etc.

libsoftwarecontainer/component-test/devicenodegateway_componenttest.cpp
```diff
@@ -97,3 +97,60 @@ TEST_F(DeviceNodeGatewayTest, TestOverwriteDeviceFails) {
     checkMode.wait();
     ASSERT_TRUE(checkMode.isSuccess());
 }
+
+struct testSetup
+{
+    const std::string firstConfiguration;
+    const std::string secondConfiguration;
```
tobsan, line 104: It would be nice to be able to use more than two configuration stages in the tests, since the implementation is not limited to two afaict.

tobsan (changes requested):
Looks good, just had some comments. I like the general idea to mark a device as up-to-date or not.

libsoftwarecontainer/component-test/devicenodegateway_componenttest.cpp
breakreturn: This is a good point, the quick-launch feature might not require more than two sequences at the moment, but our support for this in general is not so far limited to two and I don't think this should be either.

libsoftwarecontainer/src/gateway/devicenode/devicenodegateway.h
kursatkobya: This function is needed by gtest to examine the idea of "only update device when it is needed".

libsoftwarecontainer/src/gateway/devicenode/devicenodegateway.cpp
kursatkobya: this part is not new code, but sure.

libsoftwarecontainer/src/gateway/devicenode/devicenodegateway.cpp
```diff
@@ -51,38 +51,45 @@ bool DeviceNodeGateway::activateGateway()
...
 
     for (auto &dev : devlist) {
-        log_info() << "Mapping device " << dev.name;
+        if (!dev.isConfigured) {
+            log_info() << "Mapping device " << dev.name;
 
-        std::string devicePathInContainerOnHost = buildPath(getContainer()->rootFS(), dev.name);
-        std::string deviceParent = parentPath(devicePathInContainerOnHost);
+            std::string devicePathInContainerOnHost = buildPath(getContainer()->rootFS(), dev.name);
+            std::string deviceParent = parentPath(devicePathInContainerOnHost);
```
tobsan: I just realized that this whole check for `deviceParent` etc only works because we are lucky. The /dev/ directory inside the container is mounted as a tmpfs, and as such, we can't see it from the host side.

@kursatkobya can you update the test to try to mount `/dev/input/event0` into the container? I suspect this should fail (I should've taken care of this when I handled the general bind-mounting issue) because `/dev/input` does not exist inside the container, and the `createDirectory` calls will only create it on the host (not inside the tmpfs where `/dev` in the container resides).

If that is the case, we should update to try to create any needed parent directories inside the container (with a functionjob, I guess).

libsoftwarecontainer/src/gateway/devicenode/devicenodegateway.cpp
```diff
@@ -51,38 +51,29 @@ bool DeviceNodeGateway::activateGateway()
     }
 
     for (auto &dev : devlist) {
-        log_info() << "Mapping device " << dev.name;
-
-        std::string devicePathInContainerOnHost = buildPath(getContainer()->rootFS(), dev.name);
-        std::string deviceParent = parentPath(devicePathInContainerOnHost);
-
-        // Already existing files can't be converted into directories
-        if (existsInFileSystem(deviceParent) && !isDirectory(deviceParent)) {
```
frznlogic, line 60: Are these checks performed somewhere else or why are we removing them?

libsoftwarecontainer/src/gateway/devicenode/devicenodegateway.h
frznlogic: Could it be made protected instead in that case?

frznlogic (approved):

libsoftwarecontainer/src/gateway/devicenode/devicenodegateway.cpp
kursatkobya: Those lines are removed, because we are no longer support creating and mounting new devices.

tobsan: or rather, when we add an existing device, LXC handles creation of any parent directories.

libsoftwarecontainer/src/gateway/devicenode/devicenode.cpp
```diff
@@ -0,0 +1,94 @@
...
+}
+
+Device::Device(const Device &dev) :
+        name(dev.name),
+        mode(dev.mode),
+        isConfigured(dev.isConfigured)
+{
+}
+
+bool Device::parse(const json_t *element)
```
tobsan: I would prefer if this function was a static factory function, so that it creates a Device object on successful parse.

libsoftwarecontainer/src/gateway/devicenode/devicenode.h
```diff
@@ -0,0 +1,69 @@
...
+
+
+namespace softwarecontainer {
+
+class Device
+{
+    LOG_DECLARE_CLASS_CONTEXT("DN", "Device Node");
+
+public:
+    Device(std::string _name = "", int _mode = -1, bool _condigured = false);
```
tobsan: I don't think we need defaults for these, do we?

libsoftwarecontainer/src/gateway/devicenode/devicenode.h
```diff
@@ -0,0 +1,69 @@
...
+
+    /**
+     * @brief Activates this device by running mknod and chmod commands which are run in the
+     * container.
+     *
+     * @return true upon success, false otherwise
+     */
+    bool activate(std::shared_ptr<ContainerAbstractInterface> container);
+
+    std::string name;
```
tobsan: This should be private with getter/setter if needed.

libsoftwarecontainer/src/gateway/devicenode/devicenode.h
```diff
@@ -0,0 +1,69 @@
...
+    /*
+     * @brief Holds device mode which consist 3 integer number between 0 and 7 which represents
+     * permissions for owner, groups and all users
+     * */
+    int mode;
```
tobsan: This should be private with getter/setter if needed.

libsoftwarecontainer/src/gateway/devicenode/devicenode.h
```diff
@@ -0,0 +1,69 @@
...
+
+    /*
+    * @brief Holds the flag that whether the device is configured or not
+    */
+    bool isConfigured;
```
tobsan: This should be private with getter/setter if needed.

libsoftwarecontainer/src/gateway/devicenode/devicenodelogic.cpp
```diff
@@ -18,33 +18,40 @@
...
+std::vector<std::shared_ptr<Device>>::iterator
 DeviceNodeLogic::findDeviceByName(const std::string name)
 {
-    std::vector<DeviceNodeParser::Device>::iterator item =
+    std::vector<std::shared_ptr<Device>>::iterator item =
             std::find_if(m_devList.begin(), m_devList.end(),
-            [&] (DeviceNodeParser::Device const &d) { return d.name == name; });
+            [&] (std::shared_ptr<Device> const &d) { return d->name == name; });
 
     return item;
```
tobsan: I think it's a bit weird to return an iterator to a shared_ptr. Can't you return the shared_ptr directly instead?

libsoftwarecontainer/src/gateway/devicenode/devicenodelogic.cpp
```diff
@@ -18,33 +18,40 @@
...
+        std::shared_ptr<Device> newDevice(new Device(dev));
+        newDevice->isConfigured = false;
+        m_devList.push_back(std::move(newDevice));
     } else {
-        item->mode = calculateDeviceMode(item->mode, dev.mode);
+        auto mode = calculateDeviceMode(((*item))->mode, dev.mode);
+        if (mode != (*item)->mode) {
+            (*item)->isConfigured = false;
+            (*item)->mode = mode;
+        }
```
tobsan: If you return the shared_ptr directly, from findDeviceByName, this code would look way better.

libsoftwarecontainer/unit-test/devicenode_unittest.cpp
```diff
@@ -0,0 +1,211 @@
...
+    ASSERT_TRUE(dev.activate(container));
+    ASSERT_TRUE(dev.isConfigured);
+}
+
+
+
+struct testSetup
+{
+    Device firstConfiguration;
+    Device secondConfiguration;
```
tobsan, line 155: As I said earlier, it would be nice to be able to test more than two configuration runs. Perhaps make it a vector instead?

tobsan (changes requested):
I really like the re-design of this gateway. However, it looks super weird to hold shared_ptr of Device in DeviceNodeLogic, when the only pointer operations are done inside the logic class itself. That feels completely backwards.

I would propose that the Device class itself can do the whitelisting operations now done in DeviceNodeLogic, or that the lookup of devices in the logic class is changed if possible.

libsoftwarecontainer/src/gateway/devicenode/devicenode.h
kursatkobya: except iscondigured it would be better to have defaults imo.

libsoftwarecontainer/src/gateway/devicenode/devicenode.cpp
kursatkobya: we can overload constructor but i think this way is easier to read.

libsoftwarecontainer/unit-test/devicenode_unittest.cpp
kursatkobya: We can create another test to examine this. But what would have been the perks of having it ? Since this feature is provided by inherited Gateway class, it is actually on/off option. Either a gateway can be activated once or no limitation.

libsoftwarecontainer/src/gateway/devicenode/devicenode.cpp
```diff
@@ -0,0 +1,138 @@
...
+
+#include "devicenode.h"
+#include "functionjob.h"
+
+namespace softwarecontainer {
+
+Device::Device(std::string _name, int _mode) :
+        name(_name),
+        mode(_mode),
+        isConfigured(false)
```
tobsan: Perhaps nitpicking, but you should use m_name, m_mode and m_isConfigured for the members. Then you won't need _name etc.

libsoftwarecontainer/src/gateway/devicenode/devicenode.cpp
```diff
@@ -0,0 +1,138 @@
...
+
+int Device::getMode()
+{
+    return mode;
+}
+
+bool Device::getIsconfigured ()
+{
+    return isConfigured;
+}
```
tobsan: I would prefer if the getters had the name of the variable, so mode(), name() etc.

libsoftwarecontainer/unit-test/devicenode_unittest.cpp
tobsan: I just mean, instead of testing one or two activations, we could test one or more activations.

libsoftwarecontainer/unit-test/devicenode_unittest.cpp
```diff
@@ -0,0 +1,249 @@
...
+
+/*
+ * Verify whitelisting in the case when mode is set to more permissive
+ */
+TEST_P(DeviceModeTests, UpdateDeviceMode) {
+    auto testparams = GetParam();
+    dev.setMode(testparams.itemMode);
+    dev.calculateDeviceMode(testparams.newMode);
+    ASSERT_EQ(testparams.expectedMode, dev.getMode());
+}
```
tobsan, line 249: These tests perform the same testing as the component test that's added in this PR does, right? Then I would vote for removing the component test. We should add a service test, but that should not be part of this PR imo.

tobsan (changes requested):
Some new comments.

libsoftwarecontainer/unit-test/devicenode_unittest.cpp
kursatkobya: The component tests additionally performs mount operations. So i think we should remove them after adding some related service tests.

tobsan (approved):

PR #387, opened 2017-03-15
Test more error cases possible for Service Manifests

thenor:


thenor:
This change is moved to another pull request which includes restructuring of ConfigStore.

PR #388, opened 2017-03-16
Restructure BaseConfigStore

thenor:
Restructure baseconfigstore.cpp to make the method naming more logical
and for simplifying testability.


agent/src/capability/servicemanifestfileloader.cpp
```diff
@@ -0,0 +1,144 @@
...
+        log_error() << errorMessage;
+        throw ServiceManifestPathError(errorMessage);
+    } else if (isDirectory(m_source)) {
+        log_debug() << "Path to service manifests is a dir";
+        loadServiceManifestDir();
+    } else if (isFile(m_source)) {
+        log_debug() << "Path to service manifest is a file";
+        loadServiceManifestFile(m_source);
+    } else {
+        errorMessage = "Path to service manifest does not exist: \"" + m_source + "\"";
```
tobsan: Actually, this is not true. It could be a socket or a fifo (just not a directory or file). I would rewrite this error message.

agent/src/capability/servicemanifestfileloader.cpp
```diff
@@ -0,0 +1,144 @@
...
+    }
+    return m_content; // TODO: is this a good way to handle the default?
+}
+
+void ServiceManifestFileLoader::loadServiceManifestDir()
+{
+    std::string errorMessage;
+
+    if (isDirectoryEmpty(m_source)) {
+        log_warning() << "Path to configuration files is empty: " << m_source;
```
tobsan: Can we say "directory is empty" instead?

agent/src/capability/servicemanifestfileloader.cpp
```diff
@@ -0,0 +1,144 @@
...
+    json_error_t error;
+    json_t *fileroot = json_load_file(m_source.c_str(), 0, &error);
+
+    if (nullptr == fileroot) {
+        errorMessage = "Could not parse the Service Manifest: "
+            + filePath + ":" + std::to_string(error.line) + " : " + std::string(error.text);
+        log_error() << errorMessage;
+        throw ServiceManifestParseError(errorMessage);
+    } else if (!json_is_object(fileroot)) {
+        errorMessage = "The Service Manifest does not contain a json object: "
```
tobsan: Can we change this to say that its root node has to be an object (it CAN syntactically be an array).

agent/src/capability/servicemanifestfileloader.cpp
```diff
@@ -0,0 +1,144 @@
...
+
+/*
+ * Help functions
+ */
+
+std::vector<std::string> ServiceManifestFileLoader::fileList()
+{
+    std::vector<std::string> files;
+
+    DIR *dirFile = opendir(m_source.c_str());
```
tobsan: Consider using the Dir class from Glibmm instead: https://developer.gnome.org/glibmm/stable/classGlib_1_1Dir.html

Should make traversing easier etc and provides a higher abstraction level than the C system API.

agent/src/capability/servicemanifestloader.h
```diff
@@ -0,0 +1,60 @@
...
+public:
+    // Enforce inheriting classes to initialize service manifest source member
+    ServiceManifestLoader() = delete;
+    ServiceManifestLoader(const std::string &source) : m_source(source) {}
+
+    virtual ~ServiceManifestLoader() {
+        // TODO: can we decref these?
+        // for (json_t *item : m_content) {
+        //     json_decref(item);
+        // }
```
tobsan: Yes you can (and should) as long as the subclasses doesn't do it in their destructors.

http://stackoverflow.com/questions/654428/what-is-the-order-in-which-the-destructors-and-the-constructors-are-called-in-c

https://isocpp.org/wiki/faq/multiple-inheritance#mi-vi-dtor-order

agent/src/capability/servicemanifestloader.h
```diff
@@ -0,0 +1,60 @@
...
+    }
+
+    virtual std::vector<json_t *> loadContent() = 0; // TODO what to call this
```
tobsan: I think the name is ok.

tobsan (changes requested):
Mostly good, some wording etc, and it would be nice to remove all old things that are now commented out.

agent/src/capability/servicemanifestfileloader.cpp
thenor: Alright. I will rewrite it to something more general.

agent/src/capability/servicemanifestfileloader.cpp
thenor: This may have been a copy-paste error.

agent/src/capability/servicemanifestfileloader.cpp
thenor: I will rephrase.

agent/src/capability/servicemanifestfileloader.cpp
thenor: I uploaded an update a bit too quickly. I started by just adding out-commented code, so I would like to fix the code to use glib instead.

agent/src/capability/baseconfigstore.cpp
```diff
@@ -21,100 +21,33 @@
...
+    json_t *capabilities = json_object_get(serviceManifest, "capabilities");
     if (nullptr == capabilities) {
-        errorMessage = "Could not parse the \"capability\" object in file: " + filePath;
+        errorMessage = "Could not parse the \"capability\" object";
         log_error() << errorMessage;
         throw CapabilityParseError(errorMessage);
     }
     if (!json_is_array(capabilities)) {
-        errorMessage = "The \"capability\" object is not an array, in file: " + filePath;
+        errorMessage = "The \"capability\" object is not an array";
```
tobsan: "Capability key does not point to an array" would make more sense.

agent/src/capability/servicemanifestfileloader.cpp
```diff
@@ -0,0 +1,143 @@
...
+        log_error() << errorMessage;
+        throw ServiceManifestPathError(errorMessage);
+    } else if (isDirectory(m_source)) {
+        log_debug() << "Path to service manifests is a dir";
+        loadServiceManifestDir();
+    } else if (isFile(m_source)) {
+        log_debug() << "Path to service manifest is a file";
+        loadServiceManifestFile(m_source);
+    } else {
+        errorMessage = "The path to the service manifest(s) is not valid: \"" + m_source + "\"";
```
tobsan: Actually, I think we can be explicit about that it has to be a directory or a file.

tobsan (changes requested):
Just minor comments on wording, looks good!

agent/src/capability/servicemanifestfileloader.cpp
```diff
@@ -0,0 +1,143 @@
...
+        log_error() << errorMessage;
+        throw ServiceManifestPathError(errorMessage);
+    }
+    return m_content; // TODO: is this a good way to handle the default?
```
breakreturn: Can this code ever be reached?

agent/src/capability/servicemanifestloader.h
```diff
@@ -0,0 +1,59 @@
...
+    ServiceManifestLoader() = delete;
+    ServiceManifestLoader(const std::string &source) : m_source(source) {}
+
+    virtual ~ServiceManifestLoader() {
+        for (json_t *item : m_content) {
+            json_decref(item);
+        }
+    }
+
+    virtual std::vector<json_t *> loadContent() = 0; // TODO what to call this
```
breakreturn: About the TODO - you don't need to call it anything, this is a pure virtual and we should expect the reader to know that since it's a common thing.

Also, can't this method throw exceptions? If so, we should state that  somewhere so the user of this interface knows what to expect.

breakreturn (changes requested):

agent/src/softwarecontaineragent.cpp
```diff
@@ -73,8 +74,13 @@ SoftwareContainerAgent::SoftwareContainerAgent(Glib::RefPtr<Glib::MainContext> m
...
 
-    m_filteredConfigStore = std::make_shared<FilteredConfigStore>(serviceManifestDir);
-    m_defaultConfigStore  = std::make_shared<DefaultConfigStore>(defaultServiceManifestDir);
+    std::unique_ptr<ServiceManifestLoader> defaultLoader(new ServiceManifestFileLoader(defaultServiceManifestDir));
+    std::unique_ptr<ServiceManifestLoader> loader(new ServiceManifestFileLoader(serviceManifestDir));
+
+    m_filteredConfigStore = std::make_shared<FilteredConfigStore>(std::move(loader));
+    m_defaultConfigStore  = std::make_shared<DefaultConfigStore>(std::move(defaultLoader));
+    // m_filteredConfigStore = std::make_shared<FilteredConfigStore>(serviceManifestDir);
+    // m_defaultConfigStore  = std::make_shared<DefaultConfigStore>(defaultServiceManifestDir);
```
breakreturn: Remove?

agent/unit-test/configstore_unittest.cpp
```diff
@@ -30,89 +31,165 @@
...
+
+public:
+    // Constructor needs to init parent with the service manifest source string
+    ServiceManifestStringLoader(const std::string &source) :
+        ServiceManifestLoader(source)
+    {}
+    ~ServiceManifestStringLoader()
+    {
+        for (json_t *item : m_content) {
+            json_decref(item);
```
breakreturn: Isn't the base class doing this?

agent/src/capability/servicemanifestloader.h
```diff
@@ -0,0 +1,59 @@
...
+
+#include "softwarecontainer-common.h"
+#include "configstoreerror.h"
+
+#include <jsonparser.h>
+
+namespace softwarecontainer {
+
+/*
+ * Abstract interface to a service manifest loader.
```
breakreturn: If the base class is responsible for managing the freeing of memory, I think the description needs to be extended. It's not strictly an abstract interface any longer and the user needs to know who's responsible for what.

breakreturn (changes requested):

agent/src/capability/servicemanifestfileloader.cpp
thenor: Since the "if" is not returning anything then yes, it can be reached.

agent/src/capability/servicemanifestloader.h
thenor: Yes. Good that you pointed out that it can throw. I forgot about writing the code documentation.
It can throw, but that is up to the inheriting class, not this base class. So I guess mentioning that it throws here is not correct, but I do need to write it in the inheriting classes' h files.

agent/src/softwarecontaineragent.cpp
thenor: Yes, clean up is needed. Thanks for reminding me.

agent/src/capability/servicemanifestfileloader.cpp
breakreturn: Yeah missed that bit :)

agent/src/capability/servicemanifestloader.h
thenor: Ok. I will need to add something to describe this then.

agent/src/capability/servicemanifestloader.h
breakreturn: Shouldn't the exception be part of the interface though? The user uses this interface, and should not know about what concrete implementation there is.

agent/unit-test/configstore_unittest.cpp
thenor: You are right. I moved it to the base class and forgot to remove it here. I will remove this.

agent/src/capability/servicemanifestfileloader.cpp
thenor: Alright.

agent/src/capability/baseconfigstore.cpp
thenor: Alright. Changing it.

agent/src/capability/servicemanifestloader.h
thenor: Ok. I have an update. Let's see if it is what you had in mind. The inheriting classes are expected to throw ConfigStoreError (or an error inheriting from that class).

agent/src/capability/servicemanifestloader.h
thenor: I have added more documentation for the methods, but nothing here. I will add something here also.

agent/src/capability/servicemanifestloader.h
```diff
@@ -0,0 +1,75 @@
...
+        }
+    }
+
+    /*
+     * @brief Loads the json content of the service manifest(s) based on the source
+     * sent to the constructor.
+     *
+     * @return a vector of json_t corresponding to the content of the service manifest
+     * @throws ConfigStoreError depending on the implementation of the inheriting
+     * classes
```
thenor: I don't like this phrase myself, but didn't know what was a good way of writing it.

breakreturn: I don't understand. This is the interface that tells the user what will happen and it tells anyone inheriting this class what happens so they could implement it accordingly. An interface should not allow the implementing class to choose, it should follow what the interface dictates.

breakreturn (changes requested):

agent/src/capability/servicemanifestloader.h
thenor: Yes, of course. I will change the text.

breakreturn (approved):

agent/src/capability/servicemanifestfileloader.h
```diff
@@ -0,0 +1,90 @@
...
+
+#include <jansson.h>
+
+#include "softwarecontainer-common.h"
+#include "servicemanifestloader.h"
+#include "configstoreerror.h"
+
+namespace softwarecontainer {
+
+class ServiceManifestFileLoader : public ServiceManifestLoader
```
frznlogic, line 31: I'm missing class documentation. What is the purpose of the class? 

Also, as a curiosity thing, are we expecting to  get service manifests from something else than files in the future or whats up with the abstract interface?

agent/src/capability/servicemanifestloader.h
```diff
@@ -0,0 +1,73 @@
...
+
+namespace softwarecontainer {
+
+/**
+ * Abstract interface to a service manifest loader.
+ *
+ * The loader will either be a production loader that loads a json file
+ * from a file on the system, or a test loader that loads json from a string.
+ */
+class ServiceManifestLoader
```
frznlogic: Do we have any naming convention on how to name Abstract classes?

frznlogic (changes requested):
How do we test this code now? I can't really find any tests but a lot of files have been removed?

Also, two minor comments on naming convention for abstract classes and some documentation etc.

Otherwise looks good imho.

agent/src/capability/servicemanifestfileloader.h
thenor: The biggest reason for the change was for testability. The other abstract class is inside the unit-test suite (also included in this PR) configstore_unittest.

I do agree that I could add more class documentation though.

agent/src/capability/servicemanifestloader.h
thenor: Is that something that should be included in the documentation of this class? Or are you just wondering? I have only gone with what felt logical for me at the time.

thenor:
@frznlogic There are unit-tests which test the logic and using the string loader (included in the unit test suite), but also some unit tests testing the file loader. The service tests also use the file loader, so if something goes wrong it should show in them (hopefully).

agent/src/capability/servicemanifestloader.h
frznlogic: I'm wondering and asking :). If we do, we should rename it, but as @breakreturn already said out of topic, we don't have any rules I think.

frznlogic:
Ok, as long as you have code coverage on the functionality I'm happy with it ;)

PR #390, opened 2017-03-20
DeviceNodeGateway: Add service test

kursatkobya:
A parameterized service test is added for DeviceNodeGateway to examine
Quick-launch support by testing multiple activation on the gateway.


tobsan (approved):

servicetest/devicenode/test_devicenode_gateway.py
```diff
@@ -0,0 +1,188 @@
...
+            of capabilities.
+        """
+        try:
+            sc = Container()
+            sc.start(DATA)
+
+            sc.set_capabilities([firstcap])
+            sc.set_capabilities([secondcap])
+        finally:
+            sc.terminate()
```
breakreturn, line 189: I don't see any assert anywhere, so I'm assuming the test is considered "pass" when there is no exception? This should be explained in the comment if that is  the case.

servicetest/devicenode/test_devicenode_gateway.py
```diff
@@ -0,0 +1,188 @@
...
+    """
+    return [manifest]
+
+# Test suites
+
+
+@pytest.mark.usefixtures("testhelper", "agent")
+class TestDeviceNode(object):
+    """ This suite tests the behavior of DeviceNodeGateway within given set of capabilities.
+        It uses the DeviceNodeGateway to assert behavior.
```
breakreturn: This comment doesn't help me much as a reader. It's clear that it tests the DeviceNodeGateway, but not what behavior is expected or how that is verified.

servicetest/devicenode/test_devicenode_gateway.py
```diff
@@ -0,0 +1,188 @@
...
+
+        The tests use the Agent D-Bus interface to enable gateway configurations,
+        for a specified list of capabilities.
+    """
+
+    @pytest.mark.parametrize(("firstcap", "secondcap"), [
```
breakreturn: It's difficult to understand how these combinations are meant to interact. In this case I'm guessing none of them should cause an exception and that's it? Otherwise, it might be a good idea to structure this so it's possible for the reader to understand "capA and capB in that order should mean that..."

servicetest/devicenode/test_devicenode_gateway.py
```diff
@@ -0,0 +1,188 @@
...
+
+# External utility imports
+import os
+import time
+
+# Test framework and local test helper imports
+from testframework import Capability
+from testframework import StandardManifest
+from testframework import Container
+from dbus.exceptions import DBusException
```
breakreturn: Unused import.

servicetest/devicenode/test_devicenode_gateway.py
```diff
@@ -0,0 +1,188 @@
...
+# ARISING OUT OF OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS
+# SOFTWARE.
+#
+# For further information see LICENSE
+
+import pytest
+
+# External utility imports
+import os
+import time
```
breakreturn: Unused import

breakreturn (changes requested):

servicetest/devicenode/test_devicenode_gateway.py
kursatkobya: it was used to examine exception mechanism and to be sure that test fails on exception But i should remove it now.

servicetest/devicenode/test_devicenode_gateway.py
kursatkobya: All capabilities has comments at the place that they have declared. Maybe i should add another comment here as well

servicetest/devicenode/test_devicenode_gateway.py
kursatkobya: This is general comment to indicate this is general test suit, detailed information about particular test is presented in methods comment.

breakreturn: But in what way does it "use DeviceNodeGateway to assert behavior"? I know enough of the context to interpret this, but I'm not sure I will be able to do that in the future. I think it can be written more clearly, e.g. how the behavior is examined and verified. Or skip the sentence completely perhaps?

servicetest/devicenode/test_devicenode_gateway.py
```diff
@@ -0,0 +1,194 @@
...
+        The tests use the Agent D-Bus interface to enable gateway configurations,
+        for a specified list of capabilities.
+    """
+
+    """ Followinng decorator presents parametrization of argument list to test function.
+        each argument consist two capabilitties to applied on gateway. The test is responsible
+        for appliying each of them sequentially.
+
+        To see differences and aims of parameters please see comments on declaration
+    """
```
breakreturn: Ok, I see. Yes the reference to the definition of the caps is enough I guess, should be clear enough. I think it's enough to reference that and skip the first paragraph of this comment, it only states what a parametrization decorator is and does, and that should be known by the reader already.

breakreturn (approved):

PR #393, opened 2017-03-21
CONTRIBUTING: add reference to commit guidelines

tobsan:

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -9,6 +9,9 @@ We are happy to accept contributions. To accept a contribution, we require two t
   https://www.clahub.com/agreements/Pelagicore/softwarecontainer
   - We don't like CLAs but we have to have one. We'll remove it as soon as we can.
 
+We aim to follow the [https://wiki.gnome.org/Git/CommitMessages](GNOME guidelines) for commit
+messages.
+
 Help wanted
 ===========
 If you don't know what to start with, there are some general topics that we want to see more


kursatkobya (approved):
Maybe using "we are following" instead of "We try to" would be better for being more strict for upcoming commits

tobsan:
@kursatkobya yes, I thought about that at first, and we should be strict, but let's be honest: we're gonna screw it up from time to time as well.

CONTRIBUTING.md
```diff
@@ -9,6 +9,9 @@ We are happy to accept contributions. To accept a contribution, we require two t
   https://www.clahub.com/agreements/Pelagicore/softwarecontainer
   - We don't like CLAs but we have to have one. We'll remove it as soon as we can.
 
+We try to follow the [https://wiki.gnome.org/Git/CommitMessages](GNOME guidelines) for commit
```
thenor: I do agree with Kursat. My suggestion is to just rephrase this to "We aim to follow".

thenor (changes requested):

thenor (approved):

rpannek:
Why is the "CONTRIBUTING" in this commit all upper case? I read the linked https://wiki.gnome.org/Git/CommitMessages which explicitly says: "You can add a tag in lower cases with a colon before the rest." So we did it wrong in the first commit ;)

rpannek:
And btw in markdown you do `[link text](http://example.com)` not the other way around like in this commit.

tobsan:
@rpannek as I said: we're gonna screw it up from time to time. See #394 for updates :)

jeremiah:
I think that mentioning that SC is an open source project and you can review its license might be good. That is often what I look for when evaluating a project to use.

PR #394, opened 2017-03-21
contributing.md: wording updates + various

tobsan:

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -2,15 +2,18 @@ How to contribute
 =================
 We are happy to accept contributions. To accept a contribution, we require two things:
 
-* A developer certificate of origin: see http://developercertificate.org/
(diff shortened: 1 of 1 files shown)


rpannek:
Is the tag supposed to be a list of all the filenames of the changed files? I think we need to rethink the whole tag thing in the headline, I'm not sure we even need to force us to use it in cases like this one.

tobsan:
@rpannek no, the tag can be anything, but in this case the change only applies to one file, so it is reasonable to say that imho. In general the tag is supposed to mark what part of the code base was changed (if one chooses to use a tag, it is optional), like "agent: blablabla" or "common: blablabla" or even "dbusgateway: blablabla"

rpannek:
Your other examples sound reasonable, but I personally really dislike putting the name of the file there (especially with the suffix).

tobsan:
Updated the PR after comments from @jeremiah

rpannek (approved):

PR #395, opened 2017-03-21
user docs: describe error cases on the api calls

tobsan:
Describes what errors might happen on the various api calls, and what
the sources of those errors might be.

### DIFF
--- a/doc/chapters/api/index.rst
+++ b/doc/chapters/api/index.rst
@@ -3,150 +3,247 @@
 API
 ***
 
-This chapter documents the D-Bus API for interacting with the
-SoftwareContainerAgent. This is mostly useful for integrators.
+This chapter documents the D-Bus API for interacting with the SoftwareContainerAgent. This is mostly
+useful for integrators, or anyone writing a launcher for running apps inside SoftwareContainer
+containers.
 
 .. _dbus-api:
 
 D-Bus API
 =========
 
-D-Bus API is an IPC interface to call SoftwareContainer agent methods. The API
-provides following object path and interface.
+The D-Bus API is an interface to call SoftwareContainerAgent methods. The API provides the following
+object path and interface.
 
-:Object Path: /com/pelagicore/SoftwareContainer
-:Interface: com.pelagicore.SoftwareContainerAgent
+* Object Path: /com/pelagicore/SoftwareContainerAgent
+* Interface: com.pelagicore.SoftwareContainerAgent
 
 Error Handling
 --------------
-All methods in this API can and will return D-Bus errors on failure. All calls
-that don't return a D-Bus error have worked successfully. Any call that returns
-a D-Bus error has failed. All calls that change a state and raise a D-Bus error
-shall have their side effects documented in this chapter.
+All methods in this API can and will return D-Bus errors on failure. All calls that don't return a
+D-Bus error have worked successfully. Any call that returns a D-Bus error has failed. All calls that
+change a state and raise a D-Bus error shall have their side effects documented in this chapter.
+
+For each method, we describe the various error types and their potential sources. Some of these
+errors have a note saying that they put the container in an `invalid state`. An invalid state means
+that no other operations can be reliably done, and SoftwareContainer will disallow any operations on
+a container that is in an invalid state, except for destroying them.
 
 Methods
 -------
 
 BindMount
----------
-Binds a path on the host to the container. Any missing parent directories in
-the path will be created.
+~~~~~~~~~
+Binds a path on the host to the container. Any missing parent directories in the path will be
+created.
 
-:Parameters:
-        :containerID: ``int32`` The ID obtained by CreateContainer method.
-        :pathInHost: ``string`` absolute path in the host.
-        :pathInContainer: ``string`` the absolute path to mount to in the container.
-        :readOnly: ``bool`` indicates whether the directory is read-only or not.
+Parameters
+##########
+* containerID: ``int32`` The ID obtained by CreateContainer method.
+* pathInHost: ``string`` absolute path in the host.
+* pathInContainer: ``string`` the absolute path to mount to in the container.
+* readOnly: ``bool`` indicates whether the directory is read-only or not.
 
-|
+Error sources
+#############
+* Invalid ID: No matching container exists.
+* Bad state: The container is not started or ready
+* Bad path
+    * The host path does not exist
+    * The destination path is not absolute
+    * The destination path is already mounted to
+* Filesystem issues
+    * Could not create parent directories
+    * Could not create the target directory (if mounting a directory)
+    * Could not touch the target file (if mounting a file)
+* Mount issues
+    * Mount failed (:linuxman:`mount(2)` failed)
+    * Couldn't move the mount
+    * Failed to re-mount read-only
+
+**Note:** Currently, failing mounts are not rolled back.
 
 Create
-------
+~~~~~~
 Creates a container with given configuration.
 
-:Parameters:
-        :config: ``string`` config file in json format.
+Parameters
+##########
+* config: ``string`` config file in json format.
 
-                Example config JSON::
+Example config JSON::
 
-                [{"enableWriteBuffer": true}]
+[{"enableWriteBuffer": true}]
 
-|
+Return Values
+#############
+* containerID: ``int32`` ID of created SoftwareContainer.
 
-:Return Values:
-        :containerID: ``int32`` ID of created SoftwareContainer.
+Error sources
+#############
+* Network setup issues
+    * Bridge device not available and unable to set it up
+    * Given netmask can not be used to assign IP address to container
+    * See :ref:`Network setup<network-setup>`
+
+* LXC issues
+    * Couldn't prepare file system while initializing container
+    * Container ID already being in use (caused by lingering old containers)
+    * LXC config couldn't be loaded
+    * LXC could not create container
+
+If the error about ID already being in use shows up, one can try to restart the
+softwarecontainer-agent, since it does a cleanup of old containers on startup.
 
 Destroy
--------
-Tears down all active gateways related to container and shuts down the
-container with all reserved sources.
+~~~~~~~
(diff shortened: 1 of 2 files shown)


doc/chapters/api/index.rst
```diff
@@ -32,102 +32,187 @@ BindMount
...
-
-|
+Parameters
+##########
+* containerID: ``int32`` The ID obtained by CreateContainer method.
+* pathInHost: ``string`` absolute path in the host.
+* pathInContainer: ``string`` the absolute path to mount to in the container.
+* readOnly: ``bool`` indicates whether the directory is read-only or not.
+
+Error sources
```
breakreturn, line 47: I don't think this chapter reads nicely. The opening paragraphs are a bit off, they look like each sentence was added by a different person. But more to the point is that they don't really provide the information needed to understand e.g. what an "error source" is. I'm guessing it's "a reason something failed"? Is "Bad state" a special case? It looks  like it since it's followed by a colon and the other items in the list have sub-items. Is there a connection between these items and the error message the user receives? I'm not saying that the message should be duplicated here (maintenance problem), but there should be some hint of how to connect an "error source" to the user experience.

doc/chapters/api/index.rst
```diff
@@ -32,102 +32,187 @@ BindMount
...
+* Invalid state: The container is not ready or is suspended
+* Gateway errors: One or more of the gateways might have failed to shut down
+* LXC issues:
+    * LXC could not shutdown the container
+    * LXC could not force stop the container (if shutdown fails)
+    * LXC could not destroy the container
 
-|
+Failing to destroy the container leads to it being put into an invalid state. One should
+(ironically) destroy such a container.
```
breakreturn: This is ironic but not very helpful. I think we should skip the "ironically" and mention what possibilities the user have in this case, or perhaps better, link to some part of the docs that describe what an unrecoverable state means (e.g. "restart SC").

doc/chapters/api/index.rst
```diff
@@ -32,102 +32,187 @@ BindMount
...
+Return value
+############
+* pid: ``int32`` PID of process running in the container, as seen by the host.
+
+Error sources
+#############
+* Bad state: The container is not ready or is suspended
+* Capability error: couldn't set default capabilities (if ``SetCapabilities`` has not already been
+  called)
+* LXC error: The call to ``attach()`` failed.
```
breakreturn: I think this is too detailed information about the interaction with LXC. It's a maintenance burden potentially (an unnecessary one). What about "the underlying LXC operation failed" or something like that?

doc/chapters/api/index.rst
```diff
@@ -32,102 +32,187 @@ BindMount
...
+##########
+* containerID: ``int32`` The ID obtained by CreateContainer method.
+
+Error sources
+#############
+* Bad state: The container is not suspended
+* LXC error: Couldn't resume the container
 
-|
+Failure to resume a container leads to it being put in an invalid state. One should destroy any such
```
breakreturn: Maybe we should mention state in the beginning of the page, e.g what an invalid state is and what that means, as it will be the same in all cases.

breakreturn (changes requested):

breakreturn:
While you're at it, could you please also change the opath description in the beginning of the page to be correct?

doc/chapters/api/index.rst
tobsan: The reason bad state has a colon is because there is only one possible issue with the container state, whereas with the path there are many things that could go wrong. So it is not a special case, it's just that having a single sub-item made little sense to me. But I can change it for consistency.

tobsan: And yes, these items are connected, although loosely, to the error message one gets. It is intentionally not closer connected that this, since that could lead to a maintenance problem.

Moreover, I think this list is more exhaustive than the error messages one can get, since some errors are wrapped in others etc. I think it makes sense to keep them in just to give the user a hint of what might go wrong.

doc/chapters/api/index.rst
tobsan: Yes, good point.

tobsan:
@breakreturn updated after your comments.

breakreturn (approved):

doc/chapters/api/index.rst
```diff
@@ -3,131 +3,230 @@
...
+containers.
 
 .. _dbus-api:
 
 D-Bus API
 =========
 
-D-Bus API is an IPC interface to call SoftwareContainer agent methods. The API
-provides following object path and interface.
+The D-Bus API is an interface to call SoftwareContainer agent methods. The API provides following
```
thenor: For consistency write either "SoftwareContainerAgent" or "SoftwareContainer agent".

thenor: Add a "the" in the last sentence: "The API provides _the_ following"

doc/chapters/api/index.rst
```diff
@@ -3,131 +3,230 @@
...
+For each method, we describe the various error types and their potential sources. Some of these
+errors have a note saying that they put the container in an `invalid state`. An invalid state means
+that no other operations can be reliably done, and SoftwareContainer will disallow any operations on
+a container that is in an invalid state, except for destroying them.
 
 Methods
 -------
 
 BindMount
 ---------
```
thenor: Do these two headers end up on the same level? It looks like the same level. Is it possible to get them on different levels?

doc/chapters/api/index.rst
```diff
@@ -3,131 +3,230 @@
...
+* Filesystem issues
+    * Could not create parent directories
+    * Could not create the target directory (if mounting a directory)
+    * Could not touch the target file (if mounting a file)
+* Mount issues
+    * Mount failed (:linuxman:`mount(2)` failed)
+    * Couldn't move the mount
+    * Failed to re-mount read-only
+
+Currently, failing mounts are not rolled back.
```
thenor: Is this part obvious enough? Do we want to add maybe a "**Note:**" in front of it? Or do we just highlight our own shortcomings then?

doc/chapters/api/index.rst
tobsan: Looks like it, yeah. Will have a look.

doc/chapters/api/index.rst
```diff
@@ -3,131 +3,230 @@
...
+* containerID: ``int32`` ID of created SoftwareContainer.
 
-:Return Values:
-        :containerID: ``int32`` ID of created SoftwareContainer.
+Error sources
+#############
+* Network setup issues
+    * Bridge device not available and unable to set it up
+    * Given netmask can not be used to assign IP address to container
+    * See :ref:`Network setup<network-setup>`
```
thenor, line 87: I don't think it would hurt to add this type of reference to setup instructions in more sections.

doc/chapters/api/index.rst
```diff
@@ -3,131 +3,230 @@
...
 
-:Parameters:
-        :containerID: ``int32`` The ID obtained by CreateContainer method.
+If the error about ID already being in use shows up, one can try to restart the
+softwarecontainer-agent, since it does a cleanup of old containers on startup.
 
-|
+Destroy
+-------
+Tears down all active gateways related to container and shuts down the container with all reserved
```
thenor: "related to container" -> "related to the container"
You are missing a few "the" in the sentences. When it's for making the sentences brief (in the bullet lists) I think it's ok.

tobsan:
@thenor updated after your comments.

doc/chapters/api/index.rst
```diff
@@ -3,150 +3,247 @@
...
+
+Error sources
+#############
+* Invalid ID: No matching container exists.
+* Bad state: The container is not ready or is suspended
+* Capability error: couldn't set default capabilities (if ``SetCapabilities`` has not already been
+  called)
+* LXC error: The underlying LXC method call failed.
+
+Note that even if the ``Execute`` call works fine, that doesn't mean the command that is being run
```
thenor: I think you can re-use the same format as above with "**Note:** Even if..."

doc/chapters/api/index.rst
```diff
@@ -3,150 +3,247 @@
...
+* Invalid ID: No matching container exists.
+* Bad state: The container is not suspended
+* LXC error: Couldn't resume the container
 
-SetCapabilities
----------------
-Applies the given list of capability names to the container. Capabilities are
-mapped to gateway configurations and applied to each gateway for which they
-map a configuration.
+Failure to resume a container leads to it being put in an invalid state.
```
thenor: Please add "**Note:**" here also.

doc/chapters/api/index.rst
```diff
@@ -3,150 +3,247 @@
...
+Parameters
+##########
+* containerID: ``int32`` The ID obtained by CreateContainer method.
+* capabilities: ``array<string>`` of capability names
+
+Error sources
+#############
+* Invalid ID: No matching container exists.
+* Bad state: The container is not ready
+* The given array of capabilities is empty (semantically not an error though)
```
thenor, line 208: Is it possible to initialize/activate the container with only default caps? I know we have discussed it, but can't remember if we added/changed this behavior or not.

thenor (commented):
I have some comments, so I'll just wait to approve until you have time to read them and respond.

doc/chapters/api/index.rst
tobsan: Yes, but then one doesn't call SetCapabilities at all, and the Execute call will check if any caps has been set. If not, it will set the default ones.

doc/chapters/api/index.rst
tobsan: Yes, but I couldn't really find anything else to reference... :/

thenor: Oh. Maybe need to add if it is missing instead then. Perhaps not in this pull request but perhaps it is information that's good to have in the user docs.

doc/chapters/api/index.rst
thenor: Alright.

thenor (approved):

PR #396, opened 2017-03-21
container: check return value of mount

tobsan:


libsoftwarecontainer/src/container.cpp
```diff
@@ -550,6 +550,7 @@ bool Container::destroy(unsigned int timeout)
     if (m_enableWriteBuffer)
     {
         log_debug() << "Unmounting the overlay rootfs";
+        // TODO: We should check the return value here.
```
kursatkobya: can we check and return false instead of adding TODO

kursatkobya (approved):

PR #398, opened 2017-03-22
user docs: Describe systemd integration of SoftwareContainer

kursatkobya:
All layers that use SoftwareContainer define their own service files.
Therefore the systemd file in the repository is removed and a small
instructions guide is added to the integration guidelines.


doc/chapters/integration-guidelines/10-systemd-service.rst
```diff
@@ -0,0 +1,22 @@
+systemd.service
+===============
+
+As it is mentioned in :ref:`Getting Started <getting-started>` section the user API is implemented by
+the SoftwareContainer Agent as a D-bus service. The API is used to start, configure, and control
+containers. Thus in order to use SoftwareContainers the agent should be run first. To be able to start
```
tobsan: I would phrase this as "SoftwareContainer containers" instead of "SoftwareContainers"

doc/chapters/integration-guidelines/10-systemd-service.rst
```diff
@@ -0,0 +1,22 @@
...
+the agent and watch its process/system logs it would be wise decision to use systemd.service.
```
tobsan: Or rather, to use it "as a systemd service"

doc/chapters/integration-guidelines/10-systemd-service.rst
```diff
@@ -0,0 +1,22 @@
...
+To add the agent to init system, the integrator should prepare a configuration file whose name ends
+in ``.service``. This file encodes information about a process controlled and supervised by systemd.
+The ``.service`` files are generally kept in the ``/lib/systemd/system`` directory.
+
+Example
+-------
+The following sample will create a container where a write buffer will be
+created for all filesystems in that container
+
+.. literalinclude:: examples/softwarecontainer-agent.service
```
tobsan: I'm not sure we really want an example here, due to it creating a maintenance burden. How about just describing that there is special syntax for dbus services, and any runtime dependencies that the agent has?

tobsan (changes requested):
Mostly wording issues.

doc/chapters/integration-guidelines/10-systemd-service.rst
kursatkobya: freedesktop calls it systemd.service, thats why i wanted to stick with that wording. (ref:https://www.freedesktop.org/software/systemd/man/systemd.service.html)

tobsan: Well, the manpage is systemd.service because it applies to any systemd service. The man page consistently says just "service" when talking about a systemd service.

tobsan (approved):

doc/chapters/integration-guidelines/10-systemd-service.rst
```diff
@@ -0,0 +1,19 @@
+systemd.service
+===============
+
+As it is mentioned in :ref:`Getting Started <getting-started>` section the user API is implemented by
```
thenor: "mentioned in ..." -> "mentioned in _the_"

doc/chapters/integration-guidelines/10-systemd-service.rst
```diff
@@ -0,0 +1,19 @@
...
+the SoftwareContainer Agent as a D-bus service. The API is used to start, configure, and control
```
thenor: I just reviewed @tobsan's PR and in that part of the User Docs we use "SoftwareContainerAgent", so I think we should do the same here.

doc/chapters/integration-guidelines/10-systemd-service.rst
```diff
@@ -0,0 +1,19 @@
...
+containers. Thus in order to use SoftwareContainer containers, the agent should be run first. To be able to start
+the agent and watch its process/system logs it would be wise decision to use it as a systemd service.
+
+systemd is an init system used in Linux distributions to bootstrap the user space. As of 2015, a large
```
thenor: We could perhaps write "systemd" as \`\`systemd\`\`

thenor (approved):

PR #399, opened 2017-03-22
user docs: describe prerequisities to API methods

tobsan:
Adds a section under each method on the D-Bus API that describes what
prerequisities has to be fulfilled for a successful call to that method.

### DIFF
--- a/doc/chapters/api/index.rst
+++ b/doc/chapters/api/index.rst
@@ -18,6 +18,12 @@ object path and interface.
 * Object Path: /com/pelagicore/SoftwareContainerAgent
 * Interface: com.pelagicore.SoftwareContainerAgent
 
+Prerequisities
+--------------
+The prerequisities for each method describe what state the SoftwareContainer system needs to be in
+for a call to that method to be valid. For most methods, the only prerequisite is a successfully
+created container, which is the same as getting a container ID returned from the Create method.
+
 Error Handling
 --------------
 All methods in this API can and will return D-Bus errors on failure. All calls that don't return a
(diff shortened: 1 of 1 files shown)


doc/chapters/api/index.rst
```diff
@@ -18,6 +18,13 @@ object path and interface.
 * Object Path: /com/pelagicore/SoftwareContainerAgent
 * Interface: com.pelagicore.SoftwareContainerAgent
 
+Prerequisities
+--------------
+The prerequisities for each method describe the absolute minimum needed for a call to that method to
```
kursatkobya: > the absolute minimum needed what actually ?

tobsan: The prerequisities is mostly that one has a container, but that is the minimum. It doesn't hinder using other API methods before calling the next one. Perhaps it makes little sense, in which case I should probably remove it.

doc/chapters/api/index.rst
```diff
@@ -39,11 +46,16 @@ created.
...
 ##########
-* containerID: ``int32`` The ID obtained by CreateContainer method.
+* containerID: ``int32`` The ID obtained by Create method.
 * pathInHost: ``string`` absolute path in the host.
 * pathInContainer: ``string`` the absolute path to mount to in the container.
 * readOnly: ``bool`` indicates whether the directory is read-only or not.
 
+Prerequisities
+##############
+* A container that has been successfully created, such that the Create call returned a containerID.
```
kursatkobya: i would like more concise wording like : "Create method called and containerID is gathered"

doc/chapters/api/index.rst
```diff
@@ -79,6 +91,10 @@ Return Values
 #############
 * containerID: ``int32`` ID of created SoftwareContainer.
 
+Prerequisities
+##############
+None
```
kursatkobya, line 95: can we mention about lxc dependency here ?

kursatkobya (approved):

doc/chapters/api/index.rst
tobsan: There is no dependency for the user to have done anything before calling this method. To have LXC installed etc is not what these headings are about.

doc/chapters/api/index.rst
thenor: I would rephrase it a bit.
"The prerequisites for each method specifies the mandatory parameters necessary for a successful method call"
Also I'm not a fan of the next sentence. I have to read it many times to understand it. Perhaps rephrase that also.
E.g. "For most calls the only prerequisite is a successfully created container, which is the same as getting a container ID returned from the Create method"

PR #400, opened 2017-03-24
Fix bindmounts when using writebuffers

frznlogic:
The writebuffers wont work together with files, fifos and sockets. Disable
this for the future. Add service tests to make sure it works as expected in
the future

### DIFF
--- a/common/filetoolkitwithundo.cpp
+++ b/common/filetoolkitwithundo.cpp
@@ -127,7 +127,7 @@ bool FileToolkitWithUndo::bindMount(const std::string &src,
 
     log_debug() << "Bind-mounting " << src << " in " << dst << ", flags: " << flags;
 
-    if(enableWriteBuffer) {
+    if(enableWriteBuffer && isDirectory(src)) {
         std::string upperDir = tempDir("/tmp/sc-bindmount-upper-XXXXXX");
         std::string workDir = tempDir("/tmp/sc-bindmount-work-XXXXXX");
         fstype.assign("overlay");
--- a/servicetest/filesystem/test_filesystem.py
+++ b/servicetest/filesystem/test_filesystem.py
@@ -19,10 +19,8 @@ import pytest
 
 import os
 import time
-import subprocess
-import fileapp
-import time
 import platform
+import socket
 
 from testframework import Container
 
@@ -30,9 +28,12 @@ CURRENT_DIR = os.path.dirname(os.path.realpath(__file__))
 TESTOUTPUT_DIR = CURRENT_DIR + "/testoutput"
 TESTFILE = 'testfile.txt'
 
-# This function is used by the test framework to know where test specific files should be stored
+
+# This function is used by the test framework to know where test specific
+# files should be stored
 def output_dir():
     return TESTOUTPUT_DIR
+
 
 # This function is used by the 'agent' fixture to know where the log should
 # be stored
@@ -45,6 +46,7 @@ DATA = {
     Container.HOST_PATH: CURRENT_DIR,
     Container.READONLY: False
 }
+
 
 @pytest.mark.usefixtures("create_testoutput_dir", "dbus_launch", "agent", "assert_no_proxy")
 class TestFileSystem(object):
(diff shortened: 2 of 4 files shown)


common/filetoolkitwithundo.cpp
```diff
@@ -127,7 +127,7 @@ bool FileToolkitWithUndo::bindMount(const std::string &src,
 
     log_debug() << "Bind-mounting " << src << " in " << dst << ", flags: " << flags;
 
-    if(enableWriteBuffer) {
+    if(enableWriteBuffer && isDirectory(src)) {
```
rpannek, line 130: So this means that the bindmounted files will not use overlayfs at all? If so then this should be mentioned in the documentation I guess.

frznlogic: Files can't really use overlayfs, but yes. I'll add some documentation I guess.

frznlogic: Documentation added in userdocs.

rpannek (approved):

DunderRoffe (commented):
No prefix in the commit short log.
Otherwise, it looks good to me.

frznlogic:
Added the commit short log prefix.

servicetest/filesystem/test_filesystem.py
```diff
@@ -101,3 +103,113 @@ def test_write_buffer_flag(self, flag):
...
             ca.terminate()
+
+    @pytest.mark.xfail("platform.release() <= \"3.18.0\"")
+    @pytest.mark.parametrize("flag", [True, False])
+    def test_bindmount_files(self, flag):
+        """
+        Test that files are bindmountable as expected.
+        :param flag: enable or disable writeBuffer.
+        :return: Nothing
+        """
```
breakreturn: This comment doesn't follow the style of most other comments, take a look at the other modules and make it more like them please. I also don't think there's any use in stating that the test methods returns nothing since that's always the case afaik. No user calls these methods.

servicetest/filesystem/test_filesystem.py
```diff
@@ -101,3 +103,113 @@ def test_write_buffer_flag(self, flag):
...
+            os.remove(absoluteTestFile)
+
+    @pytest.mark.xfail("platform.release() <= \"3.18.0\"")
+    @pytest.mark.parametrize("flag", [True, False])
+    def test_bindmount_fifos(self, flag):
+        """
+        Test that fifos are bindmountable as expected.
+        :param flag: enable or disable writeBuffer.
+        :return: Nothing
+        """
```
breakreturn: Same as above.

servicetest/filesystem/test_filesystem.py
```diff
@@ -101,3 +103,113 @@ def test_write_buffer_flag(self, flag):
...
+            os.remove(absoluteTestFile)
+
+    @pytest.mark.xfail("platform.release() <= \"3.18.0\"")
+    @pytest.mark.parametrize("flag", [True, False])
+    def test_bindmount_sockets(self, flag):
+        """
+        Test that sockets are bindmountable as expected.
+        :param flag: enable or disable writeBuffer.
+        :return: Nothing
+        """
```
breakreturn: Same as above.

servicetest/filesystem/test_filesystem.py
```diff
@@ -101,3 +103,113 @@ def test_write_buffer_flag(self, flag):
...
+            f.close()
+
+        ca = Container()
+        if flag is True:
+            DATA[Container.CONFIG] = '[{"enableWriteBuffer": true}]'
+        else:
+            DATA[Container.CONFIG] = '[{"enableWriteBuffer": false}]'
+
+        try:
+            success = ca.start(DATA)
```
breakreturn: "success" is never used. Same goes for the other test methods below.

breakreturn (changes requested):

kursatkobya (commented):
It could be nice to revisit comments below test_write_buffer_flag method. Since the writebuffer flag is working now, those comments should be changed imo.

kursatkobya (approved):

frznlogic:
@breakreturn Fixed all the questionmarks. Anything else that you can see?

breakreturn (approved):

PR #402, opened 2017-03-30
Filesystem testing

frznlogic:


servicetest/filesystem/test_filesystem_strace.py
```diff
@@ -0,0 +1,140 @@
...
+    return "strace"
+
+DATA = {
+    Container.CONFIG: '[{"enableWriteBuffer": false}]',
+    Container.BIND_MOUNT_DIR: "/gateways/app",
+    Container.HOST_PATH: CURRENT_DIR,
+    Container.READONLY: False
+}
+
+def get_logblock(file, startline, stopline):
```
tobsan, line 56: This function could use some comments.

servicetest/filesystem/test_filesystem_strace.py
```diff
@@ -0,0 +1,140 @@
...
+
+        for line in input:
+            if line.strip() == stopline:
+                break
+            buf += line
+
+        return buf
+
+
+def get_strace_created_files(buf):
```
tobsan, line 75: This function could use some comments.

servicetest/filesystem/test_filesystem_strace.py
```diff
@@ -0,0 +1,140 @@
...
+        if re.search('^mkdir', line):
+            p = re.compile('^mkdir\(\"(.*)\",')
+            files.append(p.match(line).group(1))
+
+    return files
+
+@pytest.mark.usefixtures("create_testoutput_dir", "dbus_launch", "agent", "assert_no_proxy")
+class TestFileSystemStrace(object):
+    """ This suite should do some basic testing of the Filesystem within the
+        containers and that the whole chain is working properly.
```
tobsan: This comment could be expanded a bit I think.

servicetest/filesystem/test_filesystem_strace.py
```diff
@@ -0,0 +1,140 @@
...
+        time.sleep(0.5)
+        with open(logfile_path(), mode="a") as f:
+            f.write("\n" + endword + "\n")
+
+        buf = get_logblock(logfile_path(), startword, endword)
+
+        files = get_strace_created_files(buf)
+
+
+        with open("gobbles.txt", "w") as w:
```
tobsan: Nice naming :D :dancing_men:

servicetest/testframework/lib.py
```diff
@@ -380,14 +380,22 @@ def __init__(self,
...
                  default_caps_dir=None,
-                 check_connection=True):
+                 check_connection=True,
+                 exec_prefix=None):
         if log_file_path is None:
             self.__log_file = subprocess.STDOUT
         else:
-            self.__log_file = open(log_file_path, "w")
+            if os.path.exists(log_file_path):
+                os.unlink(log_file_path)
```
tobsan, line 389: Is this really desirable? Don't we want to append on existing logs as well? Since the agent is class scoped, one could have the same logfile for several test classes in the same module, for example.

tobsan (changes requested):

servicetest/filesystem/test_filesystem_strace.py
```diff
@@ -0,0 +1,140 @@
...
+            buf += line
+
+        return buf
+
+
+def get_strace_created_files(buf):
+    files = []
+
+    for line in StringIO.StringIO(buf):
+        print line
```
breakreturn: Is it intentional to keep this? Looks like something kept from when writing the test.

servicetest/filesystem/test_filesystem_strace.py
```diff
@@ -0,0 +1,140 @@
...
+        for line in input:
+            if line.strip() == stopline:
+                break
+            buf += line
+
+        return buf
+
+
+def get_strace_created_files(buf):
+    files = []
```
breakreturn: I'd prefer "file = list()"

servicetest/filesystem/test_filesystem_strace.py
```diff
@@ -0,0 +1,140 @@
...
+
+    for line in StringIO.StringIO(buf):
+        print line
+        if re.search('^open(.*)O_CREAT', line):
```
breakreturn: I think it would be useful with comments of what we're matching.

servicetest/filesystem/test_filesystem_strace.py
breakreturn: I don't like these sort of names. I sometimes find lingering test files, log files, output etc. that are very difficult to understand where they come from. Something that supports the understanding that this is from this test would be better.

servicetest/testframework/lib.py
frznlogic: Hmm, the logs grow kind of quickly... the old behavior was to delete the file (default behavior of the write mode). 
I could add a flag though that defaults to deleting the file and if you want to keep it in a specific test...? but I don't see the point of it?

breakreturn: If this might lead to us not having the full log from the whole run then I don't think we should do this.

servicetest/filesystem/test_filesystem_strace.py
frznlogic: Thanks:P Gobbles approved!

[http://vignette1.wikia.nocookie.net/southpark/images/1/15/Gobbles.png/revision/latest?cb=20130331225014](url)

servicetest/testframework/lib.py
```diff
@@ -412,7 +420,11 @@ def check_connection(self):
             # Wait for the softwarecontainerStarted message to appear on the
             # msgQueue, this is evoked when softwarecontainer-agent is ready to
             # perform work. If we timeout tear down what we have started so far.
-            while self.__rec.msg_queue().get(block=True, timeout=3) != "softwarecontainerStarted":
+            if self.__exec_prefix:
```
breakreturn, line 423: We should comment why we need different timeouts and what they are for.

breakreturn (changes requested):

servicetest/filesystem/test_filesystem_strace.py
frznlogic: The pull request is marked incomplete, it's not intended to keep that, I just wanted it accessible after running the test :)

servicetest/filesystem/test_filesystem_strace.py
frznlogic: @breakreturn The entire file will be removed when the PR is done :).

servicetest/filesystem/test_filesystem_strace.py
frznlogic: Done.

servicetest/filesystem/test_filesystem_strace.py
frznlogic: Done

servicetest/testframework/lib.py
frznlogic: Added comment

servicetest/testframework/lib.py
frznlogic: The whole run is kept "as is", but we don't want to keep the old runs imho. Or do we misunderstand each other ?

servicetest/filesystem/test_filesystem_strace.py
frznlogic: Expanded.

servicetest/testframework/lib.py
tobsan: It depends on what one means with old runs. We create a new agent for each test class, so if we have more than one test class in the same module, we only get the logs from the last one.

frznlogic: But that was already the case before my change? Would you like me to change this somehow?

breakreturn: It's OK for me to keep it as it was before to avoid scope creep in your current work.

common/filetoolkitwithundo.cpp
```diff
@@ -128,8 +129,14 @@ bool FileToolkitWithUndo::bindMount(const std::string &src,
     log_debug() << "Bind-mounting " << src << " in " << dst << ", flags: " << flags;
 
     if(enableWriteBuffer && isDirectory(src)) {
-        std::string upperDir = tempDir("/tmp/sc-bindmount-upper-XXXXXX");
-        std::string workDir = tempDir("/tmp/sc-bindmount-work-XXXXXX");
+        std::string upperDir , workDir;
+        if (tmpContainerRoot == "") {
```
tobsan: A comment about why this is a relevant check would be nice.

servicetest/testframework/lib.py
tobsan: Ah yes, "w" truncates the file, right? Then I guess nvm...

servicetest/filesystem/test_filesystem_strace.py
```diff
@@ -0,0 +1,191 @@
...
+    The tests are performed by prefixing the softwarecontainer-agent binary
+    with strace and logging the output into a logfile and then parsing the
+    logfile for the files and directories being created. We can then check the
+    paths and make sure they are valid.
+    """
+
+    @pytest.mark.xfail("platform.release() <= \"3.18.0\"")
+    def test_bindmount_sockets_strace(self):
+        """ Test that sockets are bindmountable as expected.
+        """
```
tobsan: Is there a specific reasons sockets are extra interesting to test?

tobsan (commented):
Mostly questions here.

servicetest/testframework/lib.py
frznlogic: Yes:)

servicetest/filesystem/test_filesystem_strace.py
frznlogic: Not really any specific reason as far as I know, I wasn't sure if they behaved differently than other types of files so I wanted to have another test for them :)

servicetest/filesystem/test_filesystem_strace.py
```diff
@@ -0,0 +1,191 @@
...
+def get_logblock(file, startline, stopline):
+    """ This gets a block of text/logs between the startline and stopline
+    from a file and returns the text as a string to the caller. Does not
+    include startline or stopline in the actual textblock passed to the caller.
+
+    :param file: The file to read
+    :param startline: The first line to look for
+    :param stopline: The last line to look for
+    :return: Returns a textbuffer
+    """
```
breakreturn: Please make indentation and comment style in the same way as in the other parts of the code (the other modules).

Same goes for all comments in this file I think.

servicetest/filesystem/test_filesystem_strace.py
```diff
@@ -0,0 +1,191 @@
...
+        # Directory patterns that are acceptable to create inside.
+        patterns = ['/tmp/container/SC-0',
+                    '/usr/var/lib/lxc/SC-0',
+                    '/sys/fs/cgroup/',
+                    '/run/lxc/']
+
+        # Find all instances in files matching any of the patterns and create
+        # a fileset containing all "OK" files
+        fileset = list()
+        for file in set(files):
```
DunderRoffe: `file` is a predefined function in python 2.x and should not be used as a variable name.
https://docs.python.org/2.7/library/functions.html#file

servicetest/filesystem/test_filesystem_strace.py
```diff
@@ -0,0 +1,191 @@
...
+                    fileset.append(file)
+
+        # Append specific hard to match files that will be "created" for
+        # various reasons
+        fileset.append('/')
+        fileset.append('/run/')
+        fileset.append('/tmp/stdout')
+
+        # Remove the fileset files from the files list
+        for file in fileset:
```
DunderRoffe: `file` is a predefined function in python 2.x and should not be used as a variable name.
https://docs.python.org/2.7/library/functions.html#file

servicetest/filesystem/test_filesystem_strace.py
```diff
@@ -0,0 +1,191 @@
...
+
+    @pytest.mark.xfail("platform.release() <= \"3.18.0\"")
+    def test_bindmount_sockets_strace(self):
+        """ Test that sockets are bindmountable as expected.
+        """
+        startword = "======STARTSTARTSTART====="
+        endword = "======ENDENDEND====="
+
+        with open(logfile_path(), mode="a") as f:
+            f.write("\n" + startword + "\n")
```
DunderRoffe, line 119: Do we not have a log handler? if not we should really consider getting one.

servicetest/filesystem/test_filesystem_strace.py
```diff
@@ -0,0 +1,186 @@
...
+        fileset.append('/tmp/stdout')
+
+        # Remove the fileset files from the files list
+        for file in fileset:
+            f = filter(lambda a: a != file, files)
+            files = f
+
+        # And assert that the list is empty. If anything is left, they are
+        # created in a bad location.
+        assert files == []
```
DunderRoffe: There are no newline at end of file

DunderRoffe (changes requested):

servicetest/filesystem/test_filesystem_strace.py
frznlogic: We don't at the moment, let's consider it for the future.

DunderRoffe (approved):

tobsan (approved):

breakreturn (approved):

PR #403, opened 2017-03-30
Rollback on failed unmount operations

kursatkobya:
CreateDir: Add a new class to track to hold directory creation backlog.
FileToolkitWithUndo: Use smart pointer instead of raw pointer.
Rollback on failed unmount operations
unittests for CreateDir class

Note : one tests marked to be skipped. It is necessary to mark it as "DISABLED" due to current CI conditions.

### DIFF
--- a/common/createdir.cpp
+++ b/common/createdir.cpp
@@ -24,6 +24,25 @@
 
 namespace softwarecontainer {
 
+CreateDir::~CreateDir()
+{
+    bool success = true;
+
+    while (!m_rollbackCleaners.empty()) {
+        auto cuHandler = m_rollbackCleaners.back();
+        m_rollbackCleaners.pop_back();
+
+        if (!cuHandler.clean()) {
+            success = false;
+        }
+    }
+
+    if(!success) {
+        log_error() << "One or more cleanup handlers returned error status, please check the log";
+    }
+}
+
+
 bool CreateDir::createParentDirectory(const std::string path)
 {
     log_debug() << "Creating parent directories for " << path;
@@ -54,58 +73,29 @@ bool CreateDir::createDirectory(const std::string path)
     }
 
     if (!pathInList(path)){
-        m_rollbackCleaners.emplace_back(new DirectoryCleanUpHandler(path));
+        m_rollbackCleaners.emplace_back(path);
     }
     return true;
 }
 
-void CreateDir::clear()
-{
-    m_rollbackCleaners.clear();
-    m_tempFileCleaners.clear();
-}
-
-bool CreateDir::rollBack()
-{
-    bool success = true;
-    // Clean up created directories
-    for (auto &cleanupHandler:m_rollbackCleaners) {
-        if (!cleanupHandler->clean()) {
-            log_error() << cleanupHandler->queryName() << " could not be cleaned!";
-            success = false;
-        }
-    }
-
-    // Clean up created temp directories
-    for (auto &cleanupHandler:m_tempFileCleaners) {
-        if (!cleanupHandler->clean()) {
-            log_error() << cleanupHandler->queryName() << " could not be cleaned!";
-            success = false;
-        }
-    }
-
-    clear();
-    return success;
-}
-
-std::string CreateDir::tempDir(std::string templ)
+std::string CreateDir::createTempDirectoryFromTemplate(std::string templ)
 {
     char *dir = const_cast<char*>(templ.c_str());
     dir = mkdtemp(dir);
     if (dir == nullptr) {
-        std::string message = "Failed to create buffered Directory: " + std::string(strerror(errno));
+        std::string message = "Failed to create directory from template: " + std::string(strerror(errno));
         log_warning() << message;
         throw SoftwareContainerError(message);
     }
 
-    m_tempFileCleaners.emplace_back(new DirectoryCleanUpHandler(std::string(dir)));
+    m_rollbackCleaners.emplace_back(std::string(dir));
     return std::string(dir);
 }
 
 bool CreateDir::pathInList(const std::string path)
 {
     for (auto &element : m_rollbackCleaners) {
-        if (element->queryName() == path) {
+        if (element.queryName() == path) {
             return true;
         }
     }
--- a/common/createdir.cpp
+++ b/common/createdir.cpp
@@ -18,6 +18,8 @@
  */
 
 #include "createdir.h"
+#include "softwarecontainererror.h"
+
 #include <sys/stat.h>
 
 namespace softwarecontainer {
@@ -91,8 +93,9 @@ std::string CreateDir::tempDir(std::string templ)
     char *dir = const_cast<char*>(templ.c_str());
     dir = mkdtemp(dir);
     if (dir == nullptr) {
-        log_warning() << "Failed to create buffered Directory: " << strerror(errno);
-        return nullptr;
+        std::string message = "Failed to create buffered Directory: " + std::string(strerror(errno));
+        log_warning() << message;
+        throw SoftwareContainerError(message);
     }
 
     m_tempFileCleaners.emplace_back(new DirectoryCleanUpHandler(std::string(dir)));
--- a/common/createdir.cpp
+++ b/common/createdir.cpp
@@ -1,0 +1,80 @@
+/*
+ * Copyright (C) 2017 Pelagicore AB
+ *
+ * Permission to use, copy, modify, and/or distribute this software for
+ * any purpose with or without fee is hereby granted, provided that the
+ * above copyright notice and this permission notice appear in all copies.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL
+ * WARRANTIES WITH REGARD TO THIS SOFTWARE INCLUDING ALL IMPLIED
+ * WARRANTIES OF MERCHANTABILITY AND FITNESS. IN NO EVENT SHALL THE AUTHOR
+ * BE LIABLE FOR ANY SPECIAL, DIRECT, INDIRECT, OR CONSEQUENTIAL DAMAGES
+ * OR ANY DAMAGES WHATSOEVER RESULTING FROM LOSS OF USE, DATA OR PROFITS,
+ * WHETHER IN AN ACTION OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION,
+ * ARISING OUT OF OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS
+ * SOFTWARE.
+ *
+ * For further information see LICENSE
+ */
+
+#include "createdir.h"
+#include <sys/stat.h>
+
+namespace softwarecontainer {
+
+bool CreateDir::createParentDirectory(const std::string path)
+{
+    log_debug() << "Creating parent directories for " << path;
+    std::string parent = parentPath(path);
+
+    if (!createDirectory(parent)) {
+        log_error() << "Could not create directory " << parent;
+        return false;
+    }
+
+    return true;
+}
+
+bool CreateDir::createDirectory(const std::string path)
+{
+    log_debug() << "createDirectory(" << path << ") called";
+    if (isDirectory(path)) {
+        return true;
+    }
+
+    if(!createParentDirectory(path)) {
+        log_error() << "Couldn't create parent directory for " << path;
+        return false;
+    }
+
+    if (mkdir(path.c_str(), S_IRWXU | S_IRWXG | S_IRWXO) == -1) {
+        log_error() << "Could not create directory " << path << ": " << strerror(errno);
+        return false;
+    }
+
+    m_cleanupHandlers.emplace_back(new DirectoryCleanUpHandler(path));
+    return true;
+}
+
+void CreateDir::clear()
+{
+    m_cleanupHandlers.clear();
+}
+
+bool CreateDir::rollBack()
+{
+    bool success = true;
+    // Clean up all created directories
+    for (auto &cleanupHandler:m_cleanupHandlers) {
+        if (!cleanupHandler->clean()) {
+            log_error() << cleanupHandler->queryName() << " could not be cleaned!";
+            success = false;
+        }
+    }
+
+    clear();
+    return success;
+}
+
+
+}
--- a/common/createdir.cpp
+++ b/common/createdir.cpp
@@ -37,7 +37,6 @@ bool CreateDir::createParentDirectory(const std::string path)
 
 bool CreateDir::createDirectory(const std::string path)
 {
-    log_debug() << "createDirectory(" << path << ") called";
     if (isDirectory(path)) {
         return true;
     }
@@ -52,20 +51,31 @@ bool CreateDir::createDirectory(const std::string path)
         return false;
     }
 
-    m_cleanupHandlers.emplace_back(new DirectoryCleanUpHandler(path));
+    if (!pathInList(path)){
+        m_rollbackCleaners.emplace_back(new DirectoryCleanUpHandler(path));
+    }
     return true;
 }
 
 void CreateDir::clear()
 {
-    m_cleanupHandlers.clear();
+    m_rollbackCleaners.clear();
+    m_tempFileCleaners.clear();
 }
 
 bool CreateDir::rollBack()
 {
     bool success = true;
-    // Clean up all created directories
-    for (auto &cleanupHandler:m_cleanupHandlers) {
+    // Clean up created directories
+    for (auto &cleanupHandler:m_rollbackCleaners) {
+        if (!cleanupHandler->clean()) {
+            log_error() << cleanupHandler->queryName() << " could not be cleaned!";
+            success = false;
+        }
+    }
+
+    // Clean up created temp directories
+    for (auto &cleanupHandler:m_tempFileCleaners) {
         if (!cleanupHandler->clean()) {
             log_error() << cleanupHandler->queryName() << " could not be cleaned!";
             success = false;
@@ -76,5 +86,27 @@ bool CreateDir::rollBack()
     return success;
 }
 
+std::string CreateDir::tempDir(std::string templ)
+{
+    char *dir = const_cast<char*>(templ.c_str());
+    dir = mkdtemp(dir);
+    if (dir == nullptr) {
+        log_warning() << "Failed to create buffered Directory: " << strerror(errno);
+        return nullptr;
+    }
+
+    m_tempFileCleaners.emplace_back(new DirectoryCleanUpHandler(std::string(dir)));
+    return std::string(dir);
+}
+
+bool CreateDir::pathInList(const std::string path)
+{
+    for (auto &element : m_rollbackCleaners) {
+        if (element->queryName() == path) {
+            return true;
+        }
+    }
+    return false;
+}
 
 }
--- a/common/createdir.h
+++ b/common/createdir.h
@@ -48,6 +48,26 @@ public:
      *
      * */
     void clear();
+
+    /**
+     * @brief tempDir Creates a uniquely named temporary directory according to templatePath.
+     *
+     * @warning The temporary path will be destroyed when the instance of FileToolkitWithUndo
+     *  is destroyed.
+     *
+     * @param templ a template Path used to create the path of the temporary directory, including
+     *  XXXXXX which will be replaced with a unique ID for the temporary directory
+     *
+     * @return A string path pointing to the newly created temporary directory.
+     */
+    std::string tempDir(std::string templatePath);
+
+    /**
+     * @brief m_tempFileCleaners holds cleanup handler for each created temporary directory.
+     * The main purpose of this is to rollback in any errors while bind-mounting. On success
+     * m_tempFileCleaners will be moved to cleanup handlers of FileToolkitWithUndo
+     */
+    std::vector<std::unique_ptr<DirectoryCleanUpHandler>> m_tempFileCleaners;
 private :
     /**
      * @brief createParentDirectory Recursively tries to create the directory pointed to by path.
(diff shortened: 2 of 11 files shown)


common/filetoolkitwithundo.cpp
```diff
@@ -180,15 +123,33 @@ bool FileToolkitWithUndo::overlayMount(const std::string &lower,
...
+        || !m_create.createDirectory(upper)
+        || !m_create.createDirectory(work)
+        || !m_create.createDirectory(dst))
     {
+        m_create.rollBack();
         log_error() << "Failed to create lower/upper/work directory for overlayMount. lowerdir=" <<
                        lower << ", upperdir=" << upper << ", workdir=" << work;
         return false;
     }
+    m_create.clear();
```
tobsan: I don't really like this solution at all. Wouldn't it make more sense if one could have a local FileToolKitWithUndo instance that also has a cleanup handler API?

```
FileToolkKitWithUndo ft;
if (!ft.createDirectory(lower) || ...) {
   // Something went wrong
} else 
    m_cleanupHandlers.emplace_back(ft);
}
```

tobsan: In the went wrong step, the local ft would go out of scope and would automatically cleanup as part of its dtor.

common/filetoolkitwithundo.h
```diff
@@ -127,14 +109,9 @@ class FileToolkitWithUndo
...
-private :
-    /**
-     * @brief createParentDirectory Recursively tries to create the directory pointed to by path.
-     * @param path The directory path to be created.
-     * @return true on success, false on failure
-     */
-    bool createParentDirectory(const std::string &path);
+    std::vector<std::unique_ptr<CleanUpHandler>> m_cleanupHandlers;
+
+    CreateDir m_create;
```
tobsan: Separating the directory creation code like this feels unnatural, and causes code duplication later on (either clean, or add the stuff done with m_create to a cleanuphandler anyway)

common/unit-test/createdir_unittest.cpp
```diff
@@ -0,0 +1,66 @@
...
+{
+public:
+
+    void SetUp() override
+    {
+    }
+
+    CreateDir cd;
+
+};
```
tobsan: It would make more sense to test parts of the FileToolkitWithUndo instead, in line with my other comments.

common/unit-test/createdir_unittest.cpp
```diff
@@ -0,0 +1,66 @@
...
+    chmod("/tmp/test/error/in/nested", 444);
+    ASSERT_FALSE(cd.createDirectory("/tmp/test/error/in/nested/directory/error"));
+    ASSERT_TRUE(cd.rollBack());
+}
+
+TEST_F(CreateDirTest, tempError)
+{
+    EXPECT_THROW(cd.tempDir("tmp/sc-unappropriate/template"), SoftwareContainerError);
+    ASSERT_TRUE(cd.rollBack());
+}
```
tobsan: These tests are good though, this is the type of stuff we want to test with the cleanuphandlers.

libsoftwarecontainer/src/container.cpp
```diff
@@ -110,11 +110,16 @@ bool Container::initialize()
 {
     if (m_state < ContainerState::PREPARED) {
         std::string gatewayDir = gatewaysDir();
-        if (!createDirectory(gatewayDir)) {
+        if (!m_create.createDirectory(gatewayDir)) {
+            m_create.rollBack();
```
tobsan: This change makes little sense as any parent directories would have been cleaned anyway when the container dtor is called (and this is kinda fatal error).

libsoftwarecontainer/src/container.cpp
```diff
@@ -110,11 +110,16 @@ bool Container::initialize()
...
             log_error() << "Could not create gateway directory "
                         << gatewayDir << ": " << strerror(errno);
             return false;
         }
+        m_create.clear();
+        if (!pathInList(gatewayDir)) {
+            m_cleanupHandlers.emplace_back(new DirectoryCleanUpHandler(gatewayDir));
+        }
```
tobsan: Here is another example of how this change creates double checks instead of using the existing FileToolkitWithUndo.

libsoftwarecontainer/src/container.cpp
```diff
@@ -620,14 +627,40 @@ bool Container::bindMountInContainer(const std::string &pathInHost,
...
     }
 
-    return bindMountCore(pathInHost, pathInContainer, tempPath, readOnly);
+    if (!bindMountCore(pathInHost, pathInContainer, tempPath, readOnly)) {
+        // bindMountInContainer operation is not successful. Clean mounted tempPath.
+        MountCleanUpHandler rollback{tempPath};
+        rollback.clean();
+        // clean createdDirectories as well
+        m_create.rollBack();
+        return false;
```
tobsan: This is going in the right direction though, but it could be done nicer.

libsoftwarecontainer/src/container.cpp
```diff
@@ -620,14 +627,40 @@ bool Container::bindMountInContainer(const std::string &pathInHost,
...
+            m_cleanupHandlers.emplace_back(std::move(tempFileCleaner));
+        }
+        // clear rollback history
+        m_create.clear();
+    } else {
+        // add cleanup for tempPath file
+        m_cleanupHandlers.emplace_back(new FileCleanUpHandler(tempPath));
+    }
+    m_cleanupHandlers.emplace_back(new MountCleanUpHandler(tempPath));
+    return true;
```
tobsan: This feels largely like we're copying parts of the code from the FileToolKitWithUndo and making it more specific. Do we have to?

libsoftwarecontainer/src/softwarecontainer.cpp
```diff
@@ -440,11 +440,16 @@ void SoftwareContainer::checkWorkspace()
...
+        if(!m_create.createDirectory(rootDir)) {
+            m_create.rollBack();
             std::string message = "Failed to create container root directory";
             log_error() << message;
             throw SoftwareContainerError(message);
         }
+        m_create.clear();
+        if (!pathInList(rootDir)) {
+            m_cleanupHandlers.emplace_back(new DirectoryCleanUpHandler(rootDir));
+        }
```
tobsan: This part of the code has nothing to do with failed mounts.

libsoftwarecontainer/src/container.cpp
tobsan: Also this has nothing to do with failing mounts.

common/createdir.h
```diff
@@ -0,0 +1,94 @@
...
+     * @param path The directory path to be created.
+     * @return true on success, false on failure
+     */
+    bool createParentDirectory(const std::string path);
+
+    /**
+     * @brief m_createDirectoryCleanupHandlers holds cleanup handler for each created directory.
+     * The main purpose of this is to rollback in any errors.
+     */
+    std::vector<std::unique_ptr<DirectoryCleanUpHandler>> m_rollbackCleaners;
```
tobsan: Is there a reason for using pointers here and not objects or references directly?

common/createdir.h
```diff
@@ -0,0 +1,94 @@
...
+     * @return A string path pointing to the newly created temporary directory.
+     */
+    std::string tempDir(std::string templatePath);
+
+    /**
+     * @brief m_tempFileCleaners holds cleanup handler for each created temporary directory.
+     * The main purpose of this is to rollback in any errors while bind-mounting. On success
+     * m_tempFileCleaners will be moved to cleanup handlers of FileToolkitWithUndo
+     */
+    std::vector<std::unique_ptr<DirectoryCleanUpHandler>> m_tempFileCleaners;
```
tobsan: Should this member really be public?

tobsan (changes requested):
I find the CreateDir class completely superfluous. 

Furthermore, this PR does not solve the issues that might arise from a failing bind mount inside the container. What if the mount-move fails? What if a re-mount to read-only fails? These cases does not seem to be addressed here.

common/filetoolkitwithundo.cpp
kursatkobya: You mean in overlayMount method of FileToolkitWithUndo class should create a inception FileToolkKitWithUndo class. I don't believe that would be nicer way. It does not make more sense to me, no.

libsoftwarecontainer/src/container.cpp
kursatkobya: ok it will be removed.

common/createdir.cpp
```diff
@@ -0,0 +1,116 @@
...
+    }
+    return success;
+}
+
+std::string CreateDir::tempDir(std::string templ)
+{
+    char *dir = const_cast<char*>(templ.c_str());
+    dir = mkdtemp(dir);
+    if (dir == nullptr) {
+        std::string message = "Failed to create buffered Directory: " + std::string(strerror(errno));
```
tobsan: This isn't necessarily a buffered directory, it's just a tempdir.

common/filetoolkitwithundo.h
```diff
@@ -127,14 +109,8 @@ class FileToolkitWithUndo
...
-    std::vector<CleanUpHandler *> m_cleanupHandlers;
-private :
-    /**
-     * @brief createParentDirectory Recursively tries to create the directory pointed to by path.
-     * @param path The directory path to be created.
-     * @return true on success, false on failure
-     */
-    bool createParentDirectory(const std::string &path);
+    std::vector<std::unique_ptr<CleanUpHandler>> m_cleanupHandlers;
+    std::vector<std::unique_ptr<CreateDir>> m_createDirList;
```
tobsan: It would be good with some comments about why these are separated.

common/unit-test/createdir_unittest.cpp
tobsan: Comments on all these test cases would be good.

libsoftwarecontainer/src/container.cpp
```diff
@@ -622,12 +625,23 @@ bool Container::bindMountInContainer(const std::string &pathInHost,
                 log_error() << "Could not create file " << tempPath;
                 return false;
             }
-            m_cleanupHandlers.push_back(new FileCleanUpHandler(tempPath));
```
tobsan, line 625: Why is this line removed? If I read this correctly, this was the cleanup handler for a file that was touched in case we didn't mount to a dir but to a file. So why not cleanup that file just as we clean up the directories? Especially in the case where things go bad.

libsoftwarecontainer/src/container.cpp
```diff
@@ -622,12 +625,23 @@ bool Container::bindMountInContainer(const std::string &pathInHost,
...
         }
     }
 
-    return bindMountCore(pathInHost, pathInContainer, tempPath, readOnly);
+    if (!bindMountCore(pathInHost, pathInContainer, tempPath, readOnly)) {
+        // bindMountInContainer operation is not successful. Clean mounted tempPath.
+        MountCleanUpHandler rollback{tempPath};
+        rollback.clean();
```
tobsan, line 634: So we create a `MountCleanUpHandler` with `tempPath` and then instantly clear it of that path?

common/createdir.cpp
```diff
@@ -0,0 +1,116 @@
...
+    if (!pathInList(path)){
+        m_rollbackCleaners.emplace_back(path);
+    }
+    return true;
+}
+
+void CreateDir::clear()
+{
+    m_rollbackCleaners.clear();
+}
```
tobsan: This call is only used once in the code (and I'm not sure it should). Consider removing it.

tobsan (changes requested):
Looks better, see further comments.

common/createdir.h
```diff
@@ -0,0 +1,90 @@
...
+     * @param path Path of directory to be created
+     * @return true on success, false on failure
+     */
+    bool createDirectory(const std::string path);
+
+    /**
+     * @brief rollBack calls clean method for all members of m_createDirectoryCleanupHandlers.
+     *
+     * This function is created as a helper to callers of createDirectory functions. It is strongly
+     * recommended that this functions should be called if createDirectory function returns false.
```
breakreturn: I think comments like this also should be summarized, in the context of how to use this class, as a class comment above the class declaration.

common/createdir.h
```diff
@@ -0,0 +1,90 @@
...
+class CreateDir {
+    LOG_DECLARE_CLASS_CONTEXT("CD", "Create Directory");
+public:
+    ~CreateDir();
+
+    /**
+     * @brief createDirectory Create a directory, and if successful append it
+     *  to a list of dirs to be deleted in the dtor. Since nestled dirs will
+     *  need to be deleted in reverse order to creation insert to the beginning
+     *  of the list.
```
breakreturn: Unnecessary implementation details here. Could we just say that "Creates a directory. If a directory is successfully created, it is removed when this object is destroyed"?

libsoftwarecontainer/src/container.cpp
kursatkobya: Yes, this is a line just for cleaning mount. Instead we can umount2 here, but i think it would be quite ugly.

common/createdir.h
```diff
@@ -0,0 +1,90 @@
...
+     * This function is created as a helper to callers of createDirectory functions. It is strongly
+     * recommended that this functions should be called if createDirectory function returns false.
+     *
+     * @return true on success, false on failure
+     */
+    bool rollBack();
+
+    /*@brief clear removes all elements from the m_createDirectoryCleanupHandlers
+     *
+     * */
```
breakreturn: Formatting

common/createdir.h
```diff
@@ -0,0 +1,90 @@
...
+    void clear();
+
+    /**
+     * @brief tempDir Creates a uniquely named temporary directory according to templatePath.
+     *
+     * @warning The temporary path will be destroyed when the instance of FileToolkitWithUndo
+     *  is destroyed.
```
breakreturn: Which instance is this? When is it destroyed?

libsoftwarecontainer/src/container.cpp
tobsan: Ah yes, I read it as a CreateDir object. My bad.

common/createdir.h
```diff
@@ -0,0 +1,90 @@
...
+private :
+    /**
+     * @brief createParentDirectory Recursively tries to create the directory pointed to by path.
+     * @param path The directory path to be created.
+     * @return true on success, false on failure
+     */
+    bool createParentDirectory(const std::string path);
+
+    /**
+     * @brief m_createDirectoryCleanupHandlers holds cleanup handler for each created directory.
```
breakreturn: Is this comment about an old naming for the below member? I think it's enough to comment on the member and let it be implicit that the comment is about the member directly below the comment, to avoid mistakes like this.

libsoftwarecontainer/src/container.cpp
kursatkobya: Here this is moved to end of the function. See the snippet

```
if (!pathIsDirectory) {
    m_cleanupHandlers.emplace_back(new FileCleanUpHandler(tempPath));
 }
```

breakreturn (changes requested):

common/createdir.h
kursatkobya: Those detailes are inherited from previous implementation, will change as you requested.

common/createdir.h
```diff
@@ -0,0 +1,83 @@
...
+     */
+    bool rollBack();
+
+    /**
+     * @brief tempDir Creates a uniquely named directory according to templatePath.
+     *
+     * @warning The created directory will be destroyed when the object is destroyed or rollBack
+     * method is called.
+     *
+     * @param templ a template Path used to create the path of the temporary directory, including
```
breakreturn: Typo in the param

common/createdir.h
```diff
@@ -0,0 +1,83 @@
...
+     *  XXXXXX which will be replaced with a unique ID for the temporary directory
+     *
+     * @throws SoftwareContainerError if template path is not appropriate.
+     * @return A string path pointing to the newly created temporary directory.
+     */
+    std::string tempDir(std::string templatePath);
```
breakreturn: When does one use createDirectory and tempDir? Does tempDir create a directory?

common/createdir.h
```diff
@@ -0,0 +1,83 @@
...
+private :
+    /**
+     * @brief createParentDirectory Recursively tries to create the directory pointed to by path.
+     * @param path The directory path to be created.
+     * @return true on success, false on failure
+     */
+    bool createParentDirectory(const std::string path);
+
+    /**
+     * @brief m_rollbackCleaners holds cleanup handler for each created directory.
```
breakreturn: Why mention the variable name here again? It makes it easy to miss when updating the code.

breakreturn (changes requested):

common/unit-test/createdir_unittest.cpp
tobsan: None of these tests actually test that `rollBack()` removes the files, or that they do not already exist before running the tests.

Also, this is the only place where the `rollBack()` function is used as a public member. The tests could be rewritten so that `rollBack()` doesn't have to be public. Not really important though...

breakreturn: @tobsan Isn't rollback() one of the main features? I'm probably misreading your comment...

libsoftwarecontainer/src/container.cpp
```diff
@@ -622,12 +625,24 @@ bool Container::bindMountInContainer(const std::string &pathInHost,
...
+        return false;
+    }
 
+    // bindMountInContainer succeed. Add cleanup for mounted tempPath
+    m_createDirList.push_back(std::move(createDirInstance));
+    if (!pathIsDirectory) {
+        m_cleanupHandlers.emplace_back(new FileCleanUpHandler(tempPath));
+    }
+
+    m_cleanupHandlers.emplace_back(new MountCleanUpHandler(tempPath));
```
tobsan, line 644: You could probably rewrite this so that you use the same MountCleanupHandler as mentioned above (the one that cleans instantly), and then depending on result clean or save to `m_cleanupHandlers`.

common/unit-test/createdir_unittest.cpp
tobsan: @breakreturn yes, but rollback is done automatically upon destruction of the object.

common/createdir.h
kursatkobya: tempDir and createDirectory are independent two functions. tempDir creates uniquely named temporary directory from template. 
Meanwhile createDirectory just creates a directory in given path. 
For more information one can look into bindMount function in FileToolkitWithUndo class.

common/unit-test/createdir_unittest.cpp
breakreturn: @tobsan Yeah, but I thought we needed that for some reason? Why would it otherwise exist at all on the public interface?

common/createdir.h
breakreturn: So "tempDir" should probably be named in a similar fashion as "createDirectory" is then? Otherwise the user will be forced to wonder why one says "create" and the other not, while both logically should create something.

And what do you mean with look into FileToolkitWithUndo? How would the user know that?

libsoftwarecontainer/src/container.cpp
kursatkobya: I don't see any perk of doing like this. I prefer this way if it is not disruptive for you.

common/unit-test/createdir_unittest.cpp
tobsan: We need the functionality, but it can be triggered by just letting the object die, which will happen around the same time anyway in most cases. But it doesn't hurt to keep it.

kursatkobya: We had one test examines `rollBack`. But as you know it is disabled for now due to root user issue on unit tests. Yet `rollBack` function can be removed as you mentioned.

breakreturn: I think we should have as few methods as possible visible to the user and as crisp and clear interface possible as well. More methods means more ways to do the same thing, means more tests, means more ways to fail etc. I'd just like to see one clear explanation of how to use this class, if it's "might be needed but often happens at the same time", I think there is a risk it's not thought through.

common/createdir.h
kursatkobya: I didn't create those naming but i believe it is not the worst naming. Still i will change it to createTemplateDir which make more sense. 

These functions are not for users of SoftwareContainer but for developers of it. So if FileToolkitWithUndo does not ring any bells then probably the one doesn't need to have any  further look in this class!

common/createdir.h
```diff
@@ -0,0 +1,82 @@
...
+     */
+    bool createDirectory(const std::string path);
+
+    /**
+     * @brief createTempDirectoryFromTemplate creates a uniquely named directory according to
+     * templatePath.
+     *
+     * @warning The created directory will be destroyed when the object is destroyed
+     *
+     * @param templatePath a template Path used to create the path of the temporary directory,
```
breakreturn: What's the template? What should one pass in here?

common/createdir.h
```diff
@@ -0,0 +1,82 @@
...
+     * @throws SoftwareContainerError if template path is not appropriate.
+     * @return A string path pointing to the newly created temporary directory.
+     */
+    std::string createTempDirectoryFromTemplate(std::string templatePath);
+
+private :
+    /**
+     * @brief createParentDirectory Recursively tries to create the directory pointed to by path.
+     * @param path The directory path to be created.
+     * @return true on success, false on failure
```
breakreturn: Comment formatting is inconsistent with whitespace (compare this with the other comments).

common/createdir.h
kursatkobya: The last six characters of template must be XXXXXX and these are replaced with a string that makes the directory name unique. This one is mentioned within comment in paramline area "@param templatePath a template Path used to create the path of the temporary directory,
     * including "XXXXXX" which will be replaced with a unique ID for the temporary directory"

common/unit-test/createdir_unittest.cpp
```diff
@@ -0,0 +1,81 @@
...
+ */
+TEST_F(CreateDirTest, DISABLED_creationError)
+{
+    {
+        CreateDir tempCd;
+        ASSERT_TRUE(tempCd.createDirectory("/tmp/test/error/in/nested"));
+        chmod("/tmp/test/error/in/nested", 444);
+        ASSERT_FALSE(tempCd.createDirectory("/tmp/test/error/in/nested/directory/error"));
+    }
+    ASSERT_FALSE(isDirectory("/tmp/test/error/in/nested"));
```
tobsan: Imho this type of check should be in all the tests to make sure that things are rolled back. Or it can be part of the fixture class.

tobsan (changes requested):

tobsan (approved):

PR #404, opened 2017-03-31
examples: Add mandatory system config to temp

thenor:
The temperature example failed due to an incorrectly formatted
service manifest.
Added an empty D-Bus system config, instead of omitting it entirely.


thenor:
The solution proposed in this pull request doesn't actually solve any problem. A previous change we made has most likely already solved the problem.

PR #409, opened 2017-04-04
User Docs: Add note to describe malformed parameters for EnvironmentGateway.

kursatkobya:
Environment GW currently does not check for malformed keys and values.
This addition to documentation will help user to understand malformed
arguments and refrain using them.

### DIFF
--- a/doc/chapters/gateways/10-environment-gateway.rst
+++ b/doc/chapters/gateways/10-environment-gateway.rst
@@ -74,4 +74,24 @@ to e.g. ``/tmp/test``, the varaiable value would now be set to ``/tmp/test:/some
 If ``SOME_ENVIRONMENT_VARIABLE`` had not been previously set, the value would now be
 set to ``:/some/path``.
 
+Important Note
+--------------
 
+Environment variables are internally stored as an array of pointers to strings in Linux based
+systems. This array is terminated by a NULL pointer. Each name and value pairs are matched with "="
+sign and stored together. For instance the first example configuration above will be stored as
+``SOME_ENVIRONMENT_VARIABLE=SOME_VALUE`` in environment table. Thus, please note that it would be
+necessary to **refrain from setting environment variable names or values anything that includes**
+
+- the equal sign ``'='``
+- embedded NULL characters ``'\0'``.
+
+Setting some particular variables may cause unexpected behavior on SoftwareContainer and applications
+running in a container. **Please be very cautious when you are setting following variables**
+
+- XDG_RUNTIME_DIR
+- DBUS_SESSION_BUS_ADDRESS
+- DBUS_SYSTEM_BUS_ADDRESS
+- PULSE_SERVER
+- HOME
+- PATH


doc/chapters/gateways/10-environment-gateway.rst
```diff
@@ -74,4 +74,14 @@ to e.g. ``/tmp/test``, the varaiable value would now be set to ``/tmp/test:/some
 If ``SOME_ENVIRONMENT_VARIABLE`` had not been previously set, the value would now be
 set to ``:/some/path``.
 
+Important Note
+--------------
 
+Environment variables are internally stored as an array of pointers to strings in Linux based
+systems. This array is terminated by a NULL pointer. Each name and value pairs are matched with "="
```
tobsan: I don't think this is a relevant point. We can add strings containing the NUL character to the array, right?

doc/chapters/gateways/10-environment-gateway.rst
```diff
@@ -74,4 +74,14 @@ to e.g. ``/tmp/test``, the varaiable value would now be set to ``/tmp/test:/some
...
+sign and stored together. For instance the first example configuration above will be stored as
+``SOME_ENVIRONMENT_VARIABLE=SOME_VALUE`` in environment table. Thus, please note that it would be
+necessary to **refrain from setting environment variable names or values anything that includes**
+
+- the equal sign ``'='``
+- embedded NULL characters ``'\0'``.
```
tobsan, line 87: I like this!

Can we please also add a section on that it is a bad idea setting for example XDG_RUNTIME_DIR or DBUS_SESSION_BUS_ADDRESS if one at the same time expect the dbus gateway or wayland gateway to work?

tobsan (changes requested):

doc/chapters/gateways/10-environment-gateway.rst
kursatkobya: Yeah but when the system is reading this from environ, it will think as it is the end of environment table. I can refer http://linuxdocs.org/HOWTOs/Secure-Programs-HOWTO/environment-variables.html for more information. Please check part 4.2.2. paragraph 2

doc/chapters/gateways/10-environment-gateway.rst
kursatkobya: Of course should i also add thing about other variables which are used setuid or library path or debug path.  For instance PATH, LD_LIBRARY_PATH, LD_DEBUG, LD_PRELOAD.

doc/chapters/gateways/10-environment-gateway.rst
tobsan: Ah, you mean like if an app uses an environment variable? I read it as if SoftwareContainer stored it in a certain way (and who cares about that?!). But yes, then this makes sense.

thenor: It should be possible to refer to this linux manual for clarity.

kursatkobya: it is possible but i think we do not need to do it. Cause the sentence should give concise reasoning that why we should not use those special characters. If the sentence is not enought then i should develop it i guess.

tobsan (approved):

PR #410, opened 2017-04-04
Remove dbus-c++ from everywhere (no longer used)

tobsan:
Some mentions etc of dbus-c++ were left lingering after we did the
switch to glibmm's dbus bindings. This removes the last of it.


frznlogic (approved):
A question on the cmake_modules, we don't want to do something similar to the vagrant-cookbook for the cmake_modules and break them into a separate repository and reuse them? We'd wind up with another dependency but ....

kursatkobya (approved):

tobsan:
@frznlogic Actually it's the other way around. Those macros were part of a proprietary repo and we split them out to be open, but this was the only OSS repo where they were used...

PR #411, opened 2017-04-05
cgroupsgateway: add support for cpu.shares

thenor:
Add support in CGroupsGateway for configuring cpu.shares.
Add unit tests to verify correct behavior also during whitelisting.


libsoftwarecontainer/src/gateway/cgroups/cgroupsgateway.h
```diff
@@ -112,11 +112,24 @@ class JSonError : public CgroupsGatewayError
     }
 };
 
+class InvalidValueError : public CgroupsGatewayError
+{
```
kursatkobya: maybe we can combine InvalidValueError & HexFormatError as a invalidValueInput

libsoftwarecontainer/src/gateway/cgroups/cgroupsparser.cpp
```diff
@@ -105,6 +105,36 @@ void CGroupsParser::parseCGroupsGatewayConfiguration(const json_t *element)
                 return;
             }
         }
+
+    } else if (("cpu.shares" == settingKey)) {
+        long int newValue;
```
kursatkobya: probably we don't need to use long here
an integer would be sufficient

libsoftwarecontainer/src/gateway/cgroups/cgroupsgateway.h
thenor: Yes, I was thinking of this also. Couldn't decide. I guess the error text describes in more detail anyway.

libsoftwarecontainer/src/gateway/cgroups/cgroupsparser.cpp
```diff
@@ -105,6 +105,36 @@ void CGroupsParser::parseCGroupsGatewayConfiguration(const json_t *element)
...
+        try {
+            newValue = std::stol(settingValue);
```
kursatkobya: and so stoi can be used instead

libsoftwarecontainer/src/gateway/cgroups/cgroupsparser.cpp
thenor: You may be right. There doesn't seem to be a limit, but I didn't know if we wanted to limit the user, but I do agree that an int should be sufficient.

libsoftwarecontainer/src/gateway/cgroups/cgroupsparser.cpp
```diff
@@ -105,6 +105,36 @@ void CGroupsParser::parseCGroupsGatewayConfiguration(const json_t *element)
...
+        }
+        // if the new value is smaller/equal than the old value, then we don't save the new value.
+        if (m_settings.count(settingKey) != 0) {
+            if (std::stol(m_settings[settingKey]) >= newValue) {
+                log_debug() << "Value of cpu.shares is less or equal to already set value";
+                return;
+            }
+        }
+        settingValue = std::to_string(newValue);
+        log_debug() << "Value for cpu.shares: " << settingValue;
```
kursatkobya, line 135: instead of using variable just for debug, it would be better to convert directly  like 
`log_debug() << "Value for cpu.shares: " << std::to_string(newValue);`

kursatkobya (commented):
Look good so far.
Need to be reviewed again after servicetests.
Actually i don't think cgroups needs any component tests. Maybe component test should be removed.

libsoftwarecontainer/src/gateway/cgroups/cgroupsparser.cpp
thenor: Ok. I will use int instead. :)

libsoftwarecontainer/src/gateway/cgroups/cgroupsparser.cpp
thenor: I also need the settingValue for setting the variable further down. So it's not just for the printout.

libsoftwarecontainer/src/gateway/cgroups/cgroupsparser.cpp
```diff
@@ -105,6 +105,36 @@ void CGroupsParser::parseCGroupsGatewayConfiguration(const json_t *element)
...
+            std::string errorMessage = "The value for cpu.shares is too high";
+            log_debug() << errorMessage;
+            throw InvalidValueError(errorMessage);
+        } catch (std::invalid_argument &err) {
+            std::string errorMessage = "The value for cpu.shares is not an integer";
+            log_debug() << errorMessage;
+            throw InvalidValueError(errorMessage);
+        }
+
+        if (newValue < 2) {
```
tobsan: Why does it have to be at least 2? The shares are relative to each other so any integer value should be fine, no?

libsoftwarecontainer/src/gateway/cgroups/cgroupsparser.cpp
```diff
@@ -105,6 +105,36 @@ void CGroupsParser::parseCGroupsGatewayConfiguration(const json_t *element)
...
+            std::string errorMessage = "Value of cpu.shares must be at least 2";
+            log_debug() << errorMessage;
+            throw InvalidValueError(errorMessage);
+        }
+        // if the new value is smaller/equal than the old value, then we don't save the new value.
+        if (m_settings.count(settingKey) != 0) {
+            if (std::stol(m_settings[settingKey]) >= newValue) {
```
tobsan: If choosing integer, then I guess `std::stoi` should be used here later on as well. Btw, I think the check should be the other way: if the `newValue` is larger, we save it. Giving a smaller value isn't even worthy of a debug output imho :)

thenor:
I talked to @breakreturn about perhaps removing the component test and just keeping service tests and unit tests. So unless @tobsan or other reviewers have strong opinions against that, then I can remove the component test suite for cgroups.

tobsan (changes requested):
Just small things. If we have proper unit tests and service tests we probably don't need component tests.

libsoftwarecontainer/src/gateway/cgroups/cgroupsgateway.h
thenor: @tobsan @breakreturn Perhaps one of you has an opinion about this. It doesn't really make sense, as @kursatkobya points out, to have two different Errors, since they both indicate the same type of error (invalid input).

thenor: @tobsan answered offline that he thinks should combine, so I will do that.

libsoftwarecontainer/src/gateway/cgroups/cgroupsparser.cpp
thenor: According to https://access.redhat.com/documentation/en-US/Red_Hat_Enterprise_Linux/6/html/Resource_Management_Guide/sec-cpu.html : "The value specified in the cpu.shares file must be 2 or higher. "

libsoftwarecontainer/src/gateway/cgroups/cgroupsparser.cpp
thenor: I can remove the debug printout, but not sure how you mean to change it to check that newValue is higher than the one set. I need a return for the case when it is <= newValue, since that implies we won't reach the setting on the variable.

servicetest/cgroups/test_cgroups.py
```diff
@@ -214,3 +235,40 @@ def test_netcls_cgroup_set(self):
                 assert value == int(TEST_NETCLS_VALUE, base=16)
         finally:
             sc.terminate()
+
+    def test_cpu_shares_cgroup_set(self):
+        """ TODO: add text
```
kursatkobya: more comment would be nice

servicetest/cgroups/test_cgroups.py
```diff
@@ -214,3 +235,40 @@ def test_netcls_cgroup_set(self):
...
+
+            time.sleep(0.5)
+            with open("/sys/fs/cgroup/cpu/lxc/" + containerID + "/cpu.shares", "r") as fh:
+                value = int(fh.read())
+                assert value == int(CPU_SHARES_LOW_VALUE)
+        finally:
+            sc.terminate()
+
+    def test_cpu_shares_cgroup_whitelisting(self):
+        """ TODO: add text
```
kursatkobya: more comment would be nice

kursatkobya (approved):

servicetest/cgroups/test_cgroups.py
thenor: Oh thanks for reminding me. I forgot. Good thing I wrote "TODO".

breakreturn (approved):

tobsan (approved):

PR #413, opened 2017-04-06
workspace: Move checkWorkspace

frznlogic:
This commit moves the checkWorkspace method from the SoftwareContainer to the
ContainerUtilityInterface and the usage into the SoftwareContainerAgent class
as the workspace is connected to the agent, not each container. A new check
was added to the SoftwareContainer for the containerRoot (previously
workspace).

### DIFF
--- a/agent/component-test/softwarecontaineragent_componenttest.cpp
+++ b/agent/component-test/softwarecontaineragent_componenttest.cpp
@@ -104,8 +104,8 @@ public:
                                                                   ConfigDependencies());
 
         try {
-            auto factory = std::make_shared<SoftwareContainerFactory> ();
-            auto utility = std::make_shared<ContainerUtilityInterface> (config);
+            auto factory = std::make_shared<SoftwareContainerFactory>();
+            auto utility = std::make_shared<ContainerUtilityInterface>(config);
             sca = std::make_shared<SoftwareContainerAgent>(m_context, config, factory, utility);
         }  catch(SoftwareContainerError &err) {
             log_error() << "Exception in software agent constructor";
--- a/agent/component-test/softwarecontaineragent_componenttest.cpp
+++ b/agent/component-test/softwarecontaineragent_componenttest.cpp
@@ -104,8 +104,8 @@ public:
                                                                   ConfigDependencies());
 
         try {
-            std::shared_ptr<SoftwareContainerFactory> factory = std::shared_ptr<SoftwareContainerFactory> (new SoftwareContainerFactory());
-            std::shared_ptr<ContainerUtilityInterface> utility = std::shared_ptr<ContainerUtilityInterface> (new ContainerUtilityInterface());
+            auto factory = std::make_shared<SoftwareContainerFactory> ();
+            auto utility = std::make_shared<ContainerUtilityInterface> (config);
             sca = std::make_shared<SoftwareContainerAgent>(m_context, config, factory, utility);
         }  catch(SoftwareContainerError &err) {
             log_error() << "Exception in software agent constructor";
--- a/agent/src/containerutilityinterface.cpp
+++ b/agent/src/containerutilityinterface.cpp
@@ -19,9 +19,17 @@
 
 #include "containerutilityinterface.h"
 
+#include "softwarecontainer-common.h"
+#include "createdir.h"
+
+
 #include <lxc/lxccontainer.h>
 
 namespace softwarecontainer {
+ContainerUtilityInterface::ContainerUtilityInterface(std::shared_ptr<Config> config)
+    : m_config(std::move(config))
+{
+}
 
 void ContainerUtilityInterface::removeOldContainers()
 {
@@ -69,5 +77,22 @@ void ContainerUtilityInterface::removeOldContainers()
     delete containerList;
     delete containerNames;
 }
+
+void ContainerUtilityInterface::checkWorkspace()
+{
+    const std::string rootDir = m_config->getStringValue("SoftwareContainer", "shared-mounts-dir");
+    if (!isDirectory(rootDir)) {
+        log_debug() << "Container root " << rootDir << " does not exist, trying to create";
+        std::unique_ptr<CreateDir> createDirInstance{new CreateDir()};
+        if(!createDirInstance->createDirectory(rootDir)) {
+            std::string message = "Failed to create container root directory";
+            log_error() << message;
+            throw SoftwareContainerError(message);
+        }
+
+        m_createDirList.push_back(std::move(createDirInstance));
+    }
+}
+
 
 } //namespace
--- a/agent/src/containerutilityinterface.cpp
+++ b/agent/src/containerutilityinterface.cpp
@@ -83,7 +83,7 @@ void ContainerUtilityInterface::checkWorkspace()
     const std::string rootDir = m_config->getStringValue("SoftwareContainer", "shared-mounts-dir");
     if (!isDirectory(rootDir)) {
         log_debug() << "Container root " << rootDir << " does not exist, trying to create";
-        std::unique_ptr<CreateDir> createDirInstance{new CreateDir()};
+        std::unique_ptr<CreateDir> createDirInstance(new CreateDir());
         if(!createDirInstance->createDirectory(rootDir)) {
             std::string message = "Failed to create container root directory";
             log_error() << message;
--- a/agent/src/containerutilityinterface.h
+++ b/agent/src/containerutilityinterface.h
@@ -70,7 +70,7 @@ class ContainerUtilityInterface : private FileToolkitWithUndo
     LOG_DECLARE_CLASS_CONTEXT("CUI", "Container Utility Interface");
 public:
     ContainerUtilityInterface(std::shared_ptr<Config> config);
-    ~ContainerUtilityInterface() {}
+    virtual ~ContainerUtilityInterface() {}
 
     /**
      * @brief This method cleans unused old containers before agent starts up
--- a/agent/src/containerutilityinterface.h
+++ b/agent/src/containerutilityinterface.h
@@ -25,6 +25,10 @@
 
 #include "softwarecontainer-log.h"
 #include "softwarecontainererror.h"
+#include "config/config.h"
+#include "filetoolkitwithundo.h"
+
+#include <memory>
 
 namespace softwarecontainer {
 
@@ -61,15 +65,24 @@ protected:
  * This class contains utility functions
  */
 
-class ContainerUtilityInterface {
+class ContainerUtilityInterface : private FileToolkitWithUndo
+{
     LOG_DECLARE_CLASS_CONTEXT("CUI", "Container Utility Interface");
 public:
-    virtual ~ContainerUtilityInterface() {}
+    ContainerUtilityInterface(std::shared_ptr<Config> config);
+    ~ContainerUtilityInterface() {}
 
     /**
      * @brief This method cleans unused old containers before agent starts up
      */
-    virtual void removeOldContainers(void);
+    void removeOldContainers(void);
+    /**
+     * @brief Check that the workspace exists.
+     */
+    void checkWorkspace(void);
+
+private:
+    std::shared_ptr<Config> m_config;
 };
 
 } //namespace
--- a/agent/src/main.cpp
+++ b/agent/src/main.cpp
@@ -216,8 +216,8 @@ int main(int argc, char **argv)
         std::shared_ptr<Config> config = std::make_shared<Config>(std::move(configSources),
                                                                   ConfigDefinition::mandatory(),
                                                                   ConfigDependencies());
-        auto factory = std::make_shared<SoftwareContainerFactory> ();
-        auto utility = std::make_shared<ContainerUtilityInterface> (config);
+        auto factory = std::make_shared<SoftwareContainerFactory>();
+        auto utility = std::make_shared<ContainerUtilityInterface>(config);
 
         // Create the actual agent
         ::softwarecontainer::SoftwareContainerAgent agent(mainContext, config, factory, utility);
--- a/agent/src/main.cpp
+++ b/agent/src/main.cpp
@@ -216,10 +216,8 @@ int main(int argc, char **argv)
         std::shared_ptr<Config> config = std::make_shared<Config>(std::move(configSources),
                                                                   ConfigDefinition::mandatory(),
                                                                   ConfigDependencies());
-        std::shared_ptr<SoftwareContainerFactory> factory =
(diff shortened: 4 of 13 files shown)


frznlogic:
@breakreturn could you make sure you are ok with the changes to the ContainerUtilityInterface. I'm not sure if this is the right way of doing this and thought i'd push it early so you can have some input.

agent/src/containerutilityinterface.h
```diff
@@ -61,15 +65,24 @@ class ContainerUtilityInterfaceError : public SoftwareContainerError
  * This class contains utility functions
  */
 
-class ContainerUtilityInterface {
+class ContainerUtilityInterface :
+    private FileToolkitWithUndo {
```
breakreturn: Put the brace on the next line or put the inheritance stuff on the same line please.

agent/component-test/softwarecontaineragent_componenttest.cpp
```diff
@@ -105,7 +105,7 @@ class SoftwareContainerAgentTest: public ::testing::Test
 
         try {
             std::shared_ptr<SoftwareContainerFactory> factory = std::shared_ptr<SoftwareContainerFactory> (new SoftwareContainerFactory());
-            std::shared_ptr<ContainerUtilityInterface> utility = std::shared_ptr<ContainerUtilityInterface> (new ContainerUtilityInterface());
+            std::shared_ptr<ContainerUtilityInterface> utility = std::shared_ptr<ContainerUtilityInterface> (new ContainerUtilityInterface(config));
```
kursatkobya: to be concise could you use somethin like that instead 
`auto utility = std::make_shared<ContainerUtilityInterface>(config);`

tobsan: I would prefer you only pass the one value you need from the config. If we need more values later on we can change it to take the whole config as an argument.

agent/src/containerutilityinterface.cpp
```diff
@@ -70,4 +78,21 @@ void ContainerUtilityInterface::removeOldContainers()
     delete containerNames;
 }
 
+void ContainerUtilityInterface::checkWorkspace()
+{
+    const std::string rootDir = m_config->getStringValue("SoftwareContainer", "shared-mounts-dir");
+    if (!isDirectory(rootDir)) {
+        log_debug() << "Container root " << rootDir << " does not exist, trying to create";
+        std::unique_ptr<CreateDir> createDirInstance = std::unique_ptr<CreateDir>(new CreateDir());
```
kursatkobya: and maybe initiate it like 
`std::unique_ptr<CreateDir> createDirInstance{new CreateDir()}`

libsoftwarecontainer/src/container.cpp
```diff
@@ -190,9 +190,9 @@ bool Container::create()
     m_rootFSPath = buildPath(s_LXCRoot, containerID, "rootfs");
 
     if (m_enableWriteBuffer) {
-        const std::string rootFSPathLower = m_containerRoot + m_id + "/rootfs-lower";
-        const std::string rootFSPathUpper = m_containerRoot + m_id + "/rootfs-upper";
-        const std::string rootFSPathWork  = m_containerRoot + m_id + "/rootfs-work";
+        const std::string rootFSPathLower = m_containerRoot + "/rootfs-lower";
+        const std::string rootFSPathUpper = m_containerRoot + "/rootfs-upper";
+        const std::string rootFSPathWork  = m_containerRoot + "/rootfs-work";
```
tobsan, line 195: This seems to be unrelated with the checkContainerRoot stuff. Please split it into its own commit.

tobsan (changes requested):

agent/src/main.cpp
```diff
@@ -219,7 +219,7 @@ int main(int argc, char **argv)
         std::shared_ptr<SoftwareContainerFactory> factory =
             std::shared_ptr<SoftwareContainerFactory> (new SoftwareContainerFactory());
         std::shared_ptr<ContainerUtilityInterface> utility =
-            std::shared_ptr<ContainerUtilityInterface> (new ContainerUtilityInterface());
+            std::shared_ptr<ContainerUtilityInterface> (new ContainerUtilityInterface(config));
```
kursatkobya: again using make_shared would make it more concise and readable

libsoftwarecontainer/src/container.cpp
frznlogic: It's related, the path available in SoftwareContainer is now "proper", ie it contains the m_id also.

agent/src/containerutilityinterface.h
frznlogic: Fixed

agent/component-test/softwarecontaineragent_componenttest.cpp
frznlogic: Fixed for both instances.

agent/src/containerutilityinterface.cpp
frznlogic: Not sure I understand?

agent/component-test/softwarecontaineragent_componenttest.cpp
frznlogic: I tend to disagree. Or rather, I'm ambivalent as I can see a need for the data inside the class as I interpret the use of it. Would you be terribly upset at keeping it?

agent/src/containerutilityinterface.cpp
frznlogic: Ok, done!

frznlogic:
I will split the commits apart a bit tomorrow.

agent/component-test/softwarecontaineragent_componenttest.cpp
```diff
@@ -104,8 +104,8 @@ class SoftwareContainerAgentTest: public ::testing::Test
                                                                   ConfigDependencies());
 
         try {
-            std::shared_ptr<SoftwareContainerFactory> factory = std::shared_ptr<SoftwareContainerFactory> (new SoftwareContainerFactory());
-            std::shared_ptr<ContainerUtilityInterface> utility = std::shared_ptr<ContainerUtilityInterface> (new ContainerUtilityInterface());
+            auto factory = std::make_shared<SoftwareContainerFactory> ();
+            auto utility = std::make_shared<ContainerUtilityInterface> (config);
```
tobsan: Remove the spaces before opening parenthesis.

agent/src/containerutilityinterface.cpp
```diff
@@ -70,4 +78,21 @@ void ContainerUtilityInterface::removeOldContainers()
     delete containerNames;
 }
 
+void ContainerUtilityInterface::checkWorkspace()
+{
+    const std::string rootDir = m_config->getStringValue("SoftwareContainer", "shared-mounts-dir");
+    if (!isDirectory(rootDir)) {
+        log_debug() << "Container root " << rootDir << " does not exist, trying to create";
+        std::unique_ptr<CreateDir> createDirInstance{new CreateDir()};
```
tobsan: Why the { syntax instead of ( ?

agent/src/containerutilityinterface.h
```diff
@@ -61,15 +65,24 @@ class ContainerUtilityInterfaceError : public SoftwareContainerError
...
  */
 
-class ContainerUtilityInterface {
+class ContainerUtilityInterface : private FileToolkitWithUndo
+{
     LOG_DECLARE_CLASS_CONTEXT("CUI", "Container Utility Interface");
 public:
-    virtual ~ContainerUtilityInterface() {}
+    ContainerUtilityInterface(std::shared_ptr<Config> config);
+    ~ContainerUtilityInterface() {}
```
tobsan: The destructor should be virtual if we want to be able to mock this interface.

common/filetoolkitwithundo.cpp
```diff
@@ -174,6 +174,35 @@ bool FileToolkitWithUndo::syncOverlayMount(const std::string &lower,
...
+    std::string fstype = "tmpfs";
+    unsigned long flags = 0;
+    std::unique_ptr<CreateDir> createDirInstance = std::unique_ptr<CreateDir>(new CreateDir());
+
+    if (!createDirInstance->createDirectory(dst)) {
+        log_error() << "Failed to create " << dst << " directory for tmpfs mount";
+        return false;
+    }
+
+    std::string mountoptions = logging::StringBuilder() << "size=" << maxSize;
```
tobsan, line 188: This can be done with std::to_string as well I think. No need to change now, just a thought.

libsoftwarecontainer/src/softwarecontainer.cpp
```diff
@@ -61,18 +61,25 @@ SoftwareContainer::SoftwareContainer(const ContainerID id,
     m_config(std::move(config)),
     m_previouslyConfigured(false)
 {
-    checkWorkspace();
+    m_containerRoot = buildPath(m_config->sharedMountsDir(), "SC-" + std::to_string(id));
+    checkContainerRoot(m_containerRoot);
+    if (m_config->enableWriteBuffer()) {
+        tmpfsMount(m_containerRoot, 100485760);
```
tobsan: Magic number.

tobsan (changes requested):

libsoftwarecontainer/src/softwarecontainer.cpp
frznlogic: The next story is about removing the magic number ;).

agent/src/containerutilityinterface.cpp
frznlogic: Copy paste from Kursats comment before, it's just a syntax difference in this case I think. It does look nicer than the original way of writing it however.

agent/src/containerutilityinterface.h
frznlogic: Ah, ok... Changing it back

common/filetoolkitwithundo.cpp
frznlogic: Thanks;)

libsoftwarecontainer/src/softwarecontainer.cpp
tobsan: I'm a bit reluctant on merging things that are super static. Can't we at least have a variable holding the number or something so that it is clear that it shouldn't be static?

frznlogic: Fair enough. Fixed in new updates ;).

tobsan (approved):

kursatkobya (approved):

agent/src/containerutilityinterface.cpp
kursatkobya: it is also safer in small aspect. one can mistakenly use () and create a function instead of running constructor of variable i thought.

agent/src/main.cpp
```diff
@@ -216,10 +216,8 @@ int main(int argc, char **argv)
         std::shared_ptr<Config> config = std::make_shared<Config>(std::move(configSources),
                                                                   ConfigDefinition::mandatory(),
                                                                   ConfigDependencies());
-        std::shared_ptr<SoftwareContainerFactory> factory =
-            std::shared_ptr<SoftwareContainerFactory> (new SoftwareContainerFactory());
-        std::shared_ptr<ContainerUtilityInterface> utility =
-            std::shared_ptr<ContainerUtilityInterface> (new ContainerUtilityInterface());
+        auto factory = std::make_shared<SoftwareContainerFactory> ();
```
breakreturn: There shouldn't be spaces before the parenthesis.

agent/unit-test/softwarecontaineragent_unittest.cpp
```diff
@@ -152,7 +154,7 @@ class SoftwareContainerAgentTest: public ::testing::Test
         Glib::RefPtr<Glib::MainContext> mainContext = Glib::MainContext::get_default();
         testContainerInterface = std::shared_ptr<::testing::NiceMock<TestContainerInterface>> (new ::testing::NiceMock<TestContainerInterface>());
         factory = std::shared_ptr<SoftwareContainerFactory> (new TestFactory(testContainerInterface));
-        containerUtility = std::shared_ptr<::testing::NiceMock<MockUtility>> (new ::testing::NiceMock<MockUtility>());
+        containerUtility = std::shared_ptr<::testing::NiceMock<MockUtility>> (new ::testing::NiceMock<MockUtility>(config));
```
breakreturn: Remove the space before parenthesis.

libsoftwarecontainer/src/softwarecontainer.cpp
```diff
@@ -61,18 +61,26 @@ SoftwareContainer::SoftwareContainer(const ContainerID id,
...
     m_container = std::shared_ptr<ContainerAbstractInterface>(
         new Container("SC-" + std::to_string(id),
                       m_config->containerConfigPath(),
-                      m_config->sharedMountsDir(),
+                      m_containerRoot,
                       m_config->enableWriteBuffer(),
                       m_config->containerShutdownTimeout()));
 
+
+
```
breakreturn: Why all the whitespace here?

servicetest/filesystem/test_filesystem.py
```diff
@@ -202,3 +203,56 @@ def test_bindmount_sockets(self, flag):
 
         if os.path.exists(absoluteTestFile):
             os.remove(absoluteTestFile)
+
+    @pytest.mark.xfail("platform.release() <= \"3.18.0\"")
+    @pytest.mark.parametrize("flag", [True, False])
+    def test_tmpfs_writebuffer_size(self, flag):
+        """ The SoftwareContainer will create a tmpfs in the location of the
+        containers temporary directory location if the write buffer is enabled.
```
breakreturn: Formatting of these comment are not as we usually do it, please have a look through the code for style and compare to the other modules in the service test tree.

servicetest/filesystem/test_filesystem.py
```diff
@@ -202,3 +203,56 @@ def test_bindmount_sockets(self, flag):
...
+
+        Note: The filesystem is hard coded to 100MB at the time this test was
+        written, but will be configurable via the DATA field in a short while
+        """
+
+        absoluteTestFile = os.path.join("/tmp/container/SC-0/rootfs-upper", TESTFILE)
+
+        ca = Container()
+
+        if flag is True:
```
breakreturn: I'm not a fan of using parametrization like this. What's the benefit compared to having two tests? The code is mostly not reused.

breakreturn (changes requested):

libsoftwarecontainer/src/softwarecontainer.cpp
frznlogic: Because I like it :D (Fixed)

breakreturn (approved):

PR #415, opened 2017-04-10
Dbus

breakreturn:


libsoftwarecontainer/src/gateway/dbus/dbusgatewayinstance.cpp
```diff
@@ -55,12 +64,13 @@ DBusGatewayInstance::DBusGatewayInstance(ProxyType type,
     // Send an empty array as config for the irrelevant proxy type
     json_t* emptyArray = json_array();
     json_object_set(m_entireConfig, unusedTypeStr, emptyArray);
-
-
 }
 
 DBusGatewayInstance::~DBusGatewayInstance()
 {
+    // TODO: The "extra" teardown here is likely unnecessary. This should be removed
```
frznlogic, line 71: Not sure i like leaving this style of comments in the code.

libsoftwarecontainer/src/gateway/dbus/dbusgatewayinstance.cpp
```diff
@@ -123,20 +134,26 @@ bool DBusGatewayInstance::activateGateway()
     return testDBusConnection(config);
 }
 
-bool DBusGatewayInstance::testDBusConnection(const std::string &config)
+bool DBusGatewayInstance::testDBusConnection(const std::string &cfg)
```
frznlogic: The name change is not nice, I'd prefer less abbreviations if possible. Don't think we have any active guideline against it however.

libsoftwarecontainer/src/gateway/dbus/dbusgatewayinstance.cpp
```diff
@@ -81,62 +91,75 @@ bool DBusGatewayInstance::readConfigElement(const json_t *element)
...
+    ss << cfg << std::string("\n");
+    std::string config = ss.str();
+
+    unsigned char configLength = config.length();
+    ssize_t count = (sizeof(char) * configLength);
+    log_debug() << "Expected config byte length " << count;
+
+    log_debug() << "Config: " << config;
+
+    ssize_t configWrite = write(m_proxyStdin, config.c_str(), count);
```
kursatkobya: maybe it would be more concise like
std::string config{cfg + std::string("\n")}
ssize_t configWrite = write(m_proxyStdin, config.c_str(), config.length());

libsoftwarecontainer/src/gateway/dbus/dbusgatewayinstance.cpp
```diff
@@ -81,62 +91,75 @@ bool DBusGatewayInstance::readConfigElement(const json_t *element)
...
 
     // writing didn't work at all
-    if (written == -1) {
+    if (configWrite == -1) {
```
kursatkobya: i would prefer 
`-1 == configWrite`

libsoftwarecontainer/src/gateway/dbus/dbusgatewayinstance.cpp
```diff
@@ -81,62 +91,75 @@ bool DBusGatewayInstance::readConfigElement(const json_t *element)
...
+                                               m_type == SessionProxy ? "session" : "system"};
+
+        // Give the dbus-proxy access to the real dbus bus address.
+        std::vector<std::string> envVec;
+
+        bool hasEnvVar = false;
+        std::string envValue = Glib::getenv(variable, hasEnvVar);
+
+        if (!hasEnvVar) {
+            if (m_type == SessionProxy) {
```
kursatkobya: would prefer
`SessionProxy == m_type`

libsoftwarecontainer/src/gateway/dbus/dbusgatewayinstance.cpp
```diff
@@ -81,62 +91,75 @@ bool DBusGatewayInstance::readConfigElement(const json_t *element)
...
 {
-    log_debug() << "Sending config " << config;
-    unsigned int count = sizeof(char) * config.length();
-    ssize_t written = write(m_infp, config.c_str(), count);
+    std::stringstream ss;
+    ss << cfg << std::string("\n");
+    std::string config = ss.str();
+
+    unsigned char configLength = config.length();
+    ssize_t count = (sizeof(char) * configLength);
```
kursatkobya: if we really need count then i would prefer auto variable i.e : 
`auto count = config.length() * sizeof(char)`

libsoftwarecontainer/src/gateway/dbus/dbusgatewayinstance.cpp
```diff
@@ -81,62 +91,75 @@ bool DBusGatewayInstance::readConfigElement(const json_t *element)
...
 
-bool DBusGatewayInstance::testDBusConnection(const std::string &config)
+bool DBusGatewayInstance::testDBusConnection(const std::string &cfg)
 {
-    log_debug() << "Sending config " << config;
-    unsigned int count = sizeof(char) * config.length();
-    ssize_t written = write(m_infp, config.c_str(), count);
+    std::stringstream ss;
+    ss << cfg << std::string("\n");
+    std::string config = ss.str();
```
kursatkobya: it would be more concise using string constructor like : 
`std::string config{cfg + std::string("\n")}`

libsoftwarecontainer/src/gateway/dbus/dbusgatewayinstance.cpp
```diff
@@ -81,62 +91,75 @@ bool DBusGatewayInstance::readConfigElement(const json_t *element)
...
     // writing didn't work at all
-    if (written == -1) {
+    if (configWrite == -1) {
         log_error() << "Failed to write to STDIN of dbus-proxy: " << strerror(errno);
         return false;
-    } else if (written == (ssize_t)count) {
-        // writing has written exact amout of bytes
-        close(m_infp);
-        m_infp = INVALID_FD;
+    } else if (configWrite == (ssize_t)count) {
```
kursatkobya: i would prefer
`count == configWrite`

libsoftwarecontainer/src/gateway/dbus/dbusgatewayinstance.cpp
kursatkobya, line 155: i would prefer
`auto configWrite`

libsoftwarecontainer/src/gateway/dbus/dbusgatewayinstance.cpp
```diff
@@ -221,6 +246,11 @@ bool DBusGatewayInstance::teardownGateway()
         log_debug() << "Socket not accessible, has it been removed already?";
     }
 
+    // Close stdin to proxy as we will no longer send configs to it
+    if (close(m_proxyStdin) == -1) {
```
kursatkobya: `-1 ==`

libsoftwarecontainer/src/gateway/gateway.cpp
```diff
@@ -100,7 +101,12 @@ bool Gateway::activate() {
 }
 
 bool Gateway::teardown() {
-    if (GatewayState::ACTIVATED != m_state) {
+    /* If a gatweway is dynamic it's possible to do teardown if
+       it has been activated at least once before.
+
+       If a gateway is not dynamic it's possible to do teardown if
+       it is in state ACTIVATED */
+    if ((m_isDynamic && !m_activatedOnce) || (!m_isDynamic && GatewayState::ACTIVATED != m_state)) {
```
kursatkobya: can we repharase this like 
`if (GatewayState::ACTIVATED != m_state && !m_activatedOnce)`

kursatkobya (approved):
Looks good. 
just some minor comments

libsoftwarecontainer/src/gateway/dbus/dbusgatewayinstance.cpp
frznlogic: I agree

libsoftwarecontainer/src/gateway/dbus/dbusgatewayinstance.cpp
breakreturn: I'm not super keen on it either. But I wanted the variable that's actually used to have the best name possible in the method. The "cfg" variable is modified and then named "config", which is then what's used in the rest of the code. Do you have an alternative that you thing would work better for this case?

frznlogic: I saw the usage when I wrote the comment and was torn between just deleting the comment and keeping it because I understand your reasoning ;). 

Perhaps configuration could work? or stringConfig?

libsoftwarecontainer/src/gateway/dbus/dbusgatewayinstance.cpp
breakreturn: No it's not nice. So I'd really like to "fix" this so the comment is moot but there's no time to do that thoroughly now. And as it was, i.e. uncommented, it was really difficult to understand the purpose and I had to waste time on finding people that knew about this. So I think keeping this info here adds some value. Not the best of solutions perhaps but at least it might save some time for the next person.

PR #416, opened 2017-04-10
writebuffers: Temp filesystem support

frznlogic:
This commit adds support for a tmpfs with a fixed size to be added on the temporary filesystem created for creating all container write buffers on. This will limit the impact of writes and make them contained to separate filesystems completely.

### DIFF
--- a/agent/src/containeroptions/containeroptionparser.cpp
+++ b/agent/src/containeroptions/containeroptionparser.cpp
@@ -23,6 +23,7 @@
 
 namespace softwarecontainer {
 
+
 ContainerOptionParser::ContainerOptionParser() :
     m_options(new DynamicContainerOptions())
 {
@@ -36,14 +37,37 @@ void ContainerOptionParser::readConfigElement(const json_t *element)
         throw ContainerOptionParseError(errorMessage);
     }
 
-    bool wo = false;
-    if(!JSONParser::read(element, "enableWriteBuffer", wo)) {
+    bool enableWriteBuffer = false;
+    if(!JSONParser::read(element, "enableWriteBuffer", enableWriteBuffer)) {
         std::string errorMessage("Could not parse config due to: 'enableWriteBuffer' not found.");
         log_error() << errorMessage;
         throw ContainerOptionParseError(errorMessage);
     }
 
-    m_options->setEnableWriteBuffer(wo);
+    m_options->setEnableWriteBuffer(enableWriteBuffer);
+
+    if (enableWriteBuffer == true) {
+        bool enableTemporaryFilesystemWriteBuffer = false;
+        if(!JSONParser::read(element,
+                             "enableTemporaryFileSystemWriteBuffer",
+                             enableTemporaryFilesystemWriteBuffer))
+        {
+            log_warn() << "Could not parse config due to: 'enableTemporaryFilesystemWriteBuffer' not found.";
+        }
+        m_options->setEnableTemporaryFileSystemWriteBuffers(enableTemporaryFilesystemWriteBuffer);
+
+        if (enableTemporaryFilesystemWriteBuffer) {
+            int temporaryFileSystemSize = DEFAULT_TEMPORARY_FILESYSTEM_SIZE;
+            if(!JSONParser::read(element,
+                                 "temporaryFileSystemSize",
+                                 temporaryFileSystemSize))
+            {
+                log_error() << "Could not parse config due to: 'temporaryFileSystemSize' not found.";
+            }
+            m_options->setTemporaryFileSystemSize(temporaryFileSystemSize);
+        }
+    }
+
 }
 
 std::unique_ptr<DynamicContainerOptions> ContainerOptionParser::parse(const std::string &config)
--- a/agent/src/containeroptions/dynamiccontaineroptions.cpp
+++ b/agent/src/containeroptions/dynamiccontaineroptions.cpp
@@ -40,4 +40,24 @@ bool DynamicContainerOptions::enableWriteBuffer() const
     return m_enableWriteBuffer;
 }
 
+void DynamicContainerOptions::setEnableTemporaryFileSystemWriteBuffers(bool enabled)
+{
+    m_enableTemporaryFileSystemWriteBuffers = enabled;
+}
+
+bool DynamicContainerOptions::enableTemporaryFileSystemWriteBuffers() const
+{
+    return m_enableTemporaryFileSystemWriteBuffers;
+}
+
+void DynamicContainerOptions::setTemporaryFileSystemSize(unsigned int size)
+{
+    m_temporaryFileSystemSize = size;
+}
+
+unsigned int DynamicContainerOptions::temporaryFileSystemSize() const
+{
+    return m_temporaryFileSystemSize;
+}
+
 } // namespace softwarecontainer
--- a/agent/unit-test/containeroptionparser_unittest.cpp
+++ b/agent/unit-test/containeroptionparser_unittest.cpp
@@ -44,11 +44,17 @@ public:
     }
 };
 
-TEST_F(ContainerOptionParserTest, parseConfigNice) {
+/*
+ * Test a simple configuration and see that it's set to true
+ */
+TEST_F(ContainerOptionParserTest, parseConfigNiceEnabled) {
     ASSERT_NO_THROW(parse("[{\"enableWriteBuffer\": true}]"));
     ASSERT_TRUE(m_options->enableWriteBuffer());
 }
 
+/*
+ * Test a simple configuration and make sure that it works multiple times
+ */
 TEST_F(ContainerOptionParserTest, parseManyTimes) {
     ASSERT_NO_THROW(parse("[{\"enableWriteBuffer\": true}]"));
     ASSERT_TRUE(m_options->enableWriteBuffer());
@@ -58,19 +64,107 @@ TEST_F(ContainerOptionParserTest, parseManyTimes) {
     ASSERT_TRUE(m_options->enableWriteBuffer());
 }
 
-TEST_F(ContainerOptionParserTest, parseConfigNice2) {
+/*
+ * Test that the configuration is disabled if it's set to being disabled
+ */
+TEST_F(ContainerOptionParserTest, parseConfigNiceDisabled) {
     ASSERT_NO_THROW(parse("[{\"enableWriteBuffer\": false}]"));
     ASSERT_FALSE(m_options->enableWriteBuffer());
 }
 
+/*
+ * Parse a "good" configuration with all parts set in it
+ */
+TEST_F(ContainerOptionParserTest, parseConfigNiceWithTmpfsAndSize) {
+    ASSERT_NO_THROW(parse("[{\"enableWriteBuffer\": true, \
+                    \"enableTemporaryFileSystemWriteBuffer\": true, \
+                    \"temporaryFileSystemSize\": 10485760}]"));
+    ASSERT_TRUE(m_options->enableWriteBuffer());
+    ASSERT_TRUE(m_options->enableTemporaryFileSystemWriteBuffers());
+    ASSERT_EQ(m_options->temporaryFileSystemSize(), 10485760);
+}
+
+/*
+ * Parse a "good" configuration with the enableTemporary.. disabled. Size
+ * should not be parsed in that case.
+ */
+TEST_F(ContainerOptionParserTest, parseConfigNiceWithTmpfsDisabled) {
+    ASSERT_NO_THROW(parse("[{\"enableWriteBuffer\": true, \
+                    \"enableTemporaryFileSystemWriteBuffer\": false, \
+                    \"temporaryFileSystemSize\": 10485760}]"));
+    ASSERT_TRUE(m_options->enableWriteBuffer());
+    ASSERT_FALSE(m_options->enableTemporaryFileSystemWriteBuffers());
+}
+
+/*
+ * Parse a "good" configuration with a smaller size.
+ */
+TEST_F(ContainerOptionParserTest, parseConfigNiceTmpfsEnabledSmall) {
+    ASSERT_NO_THROW(parse("[{\"enableWriteBuffer\": true, \
+                    \"enableTemporaryFileSystemWriteBuffer\": true, \
+                    \"temporaryFileSystemSize\": 100}]"));
+    ASSERT_TRUE(m_options->enableWriteBuffer());
+    ASSERT_TRUE(m_options->enableTemporaryFileSystemWriteBuffers());
+    ASSERT_EQ(m_options->temporaryFileSystemSize(), 100);
+}
+
+/*
+ * Parse a "good" configuration where everything is disabled.
+ */
+TEST_F(ContainerOptionParserTest, parseConfigNiceDisabledWithExtraConfig) {
+    ASSERT_NO_THROW(parse("[{\"enableWriteBuffer\": false, \
+                    \"enableTemporaryFileSystemWriteBuffer\": true, \
+                    \"temporaryFileSystemSize\": 100}]"));
+    ASSERT_FALSE(m_options->enableWriteBuffer());
+    ASSERT_FALSE(m_options->enableTemporaryFileSystemWriteBuffers());
+}
+
+/*
+ * Parse a config with no config
+ */
 TEST_F(ContainerOptionParserTest, parseConfigNoConfig) {
     ASSERT_THROW(parse(""), ContainerOptionParseError);
 }
 
+/*
+ *  Parse a badly formatted config
+ */
 TEST_F(ContainerOptionParserTest, parseConfigBadConfig) {
     ASSERT_THROW(parse("gobfmsrfe"), ContainerOptionParseError);
 }
 
+/*
+ * Parse a config with a missing Size parameter and make sure it still
+ * parses.
+ */
+TEST_F(ContainerOptionParserTest, parseConfigNiceConfigMissingSize) {
+    ASSERT_NO_THROW(parse("[{\"enableWriteBuffer\": true, \
+                    \"enableTemporaryFileSystemWriteBuffer\": true}]"));
+    ASSERT_TRUE(m_options->enableWriteBuffer());
+    ASSERT_TRUE(m_options->enableTemporaryFileSystemWriteBuffers());
+}
+
+/*
+ * Parse a "good" config where the tmpfs is disabled and missing size.
+ */
+TEST_F(ContainerOptionParserTest, parseConfigNiceConfigTmpfsDisabledMissingSize) {
+    ASSERT_NO_THROW(parse("[{\"enableWriteBuffer\": true, \
+                    \"enableTemporaryFileSystemWriteBuffer\": false}]"));
+    ASSERT_TRUE(m_options->enableWriteBuffer());
+    ASSERT_FALSE(m_options->enableTemporaryFileSystemWriteBuffers());
+}
+
+/*
+ * Parse a "good" config where enableTemporary... and size is missing
+ */
+TEST_F(ContainerOptionParserTest, parseConfigNiceConfigMissingTmpfSConfig) {
+    ASSERT_NO_THROW(parse("[{\"enableWriteBuffer\": true}]"));
+    ASSERT_TRUE(m_options->enableWriteBuffer());
+}
+
+/*
+ * Parse a "bad" config where there is a bad parameter name.
+ */
 TEST_F(ContainerOptionParserTest, parseConfigEvilConfig) {
     ASSERT_THROW(parse("[{\"WRONG_PARAM_NAME\": true}]"), ContainerOptionParseError);
     ASSERT_FALSE(m_options->enableWriteBuffer());
--- a/doc/chapters/filesystem/index.rst
+++ b/doc/chapters/filesystem/index.rst
@@ -136,3 +136,36 @@ changes performed during its runtime to be merged into the lower layers.
 .. Note:: Non-directory types of files can not be mounted using overlayfs.
           These will automatically fall back on using the default behavior of 
           bind mounting the files into the filesystem of the container.
+
+Temporary Filesystem
+====================
+
+The write buffers can be configured to use a separate temporary filesystem
+(``tmpfs``) which can be limited in size. The ``tmpfs`` is mounted on top of
+the containers temporary directory as soon as it's created and will remain
+there until the container is destroyed. All ``overlayfs`` mounts and temporary
+directories should be created inside this ``tmpfs`` mount.
+
+The ``size`` of the ``tmpfs`` can also be limited using an extra configuration
+option in the ``Create`` DBus call to the ``SoftwareContainerAgent``.
+An example configuration would look like this::
+
+    [{
+        "enableWriteBuffer": true,
+        "enableTemporaryFileSystemWriteBuffer": true,
+        "temporaryFileSystemSize": 10485760
+    }]
+
+The ``enableTemporaryFileSystemWriteBuffer`` setting enables the ``tmpfs``
+creation as described above, while the ``temporaryFileSystemSize`` variable
+sets the size of the ``tmpfs`` in bytes.
+
+.. Note:: The ``temporaryFileSystemSize`` parameter will not be parsed unless
+          the ``enableTemporaryFileSystemWriteBuffer`` parameter is ``true``.
+          The ``temporaryFileSystemSize`` is not required if the
+          ``enableTemporaryFileSystemWriteBuffer`` is set to ``false`` or not
+          added at all.
+
+.. Note:: The ``tmpfs`` is shared between the upper and work directories in 
+          ``overlayfs`` being mounted to the ``rootfs`` and all the 
+          directories being bindmounted into a single instance of a container.
--- a/servicetest/filesystem/test_filesystem.py
+++ b/servicetest/filesystem/test_filesystem.py
@@ -22,6 +22,7 @@ import time
 import platform
 import socket
 import psutil
+from dbus.exceptions import DBusException
 
 from testframework import Container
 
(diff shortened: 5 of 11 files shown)


agent/src/containeroptions/containeroptionparser.cpp
```diff
@@ -44,6 +44,32 @@ void ContainerOptionParser::readConfigElement(const json_t *element)
...
+                             "enableTemporaryFileSystemWriteBuffer",
+                             enableTemporaryFilesystemWriteBuffer)) {
+            std::string errorMessage("Could not parse config due to: 'enableTemporaryFilesystemWriteBuffer' not found.");
+            log_error() << errorMessage;
+            throw ContainerOptionParseError(errorMessage);
+        }
+        m_options->setEnableTemporaryFileSystemWriteBuffers(enableTemporaryFilesystemWriteBuffer);
+
+        if (enableTemporaryFilesystemWriteBuffer) {
+            bool temporaryFileSystemSize = 100 * 1024 * 1024;
```
kursatkobya: this should not be bool i guess

frznlogic: no it should not ;)

agent/src/containeroptions/containeroptionparser.cpp
```diff
@@ -44,6 +44,31 @@ void ContainerOptionParser::readConfigElement(const json_t *element)
...
+        if(!JSONParser::read(element,
+                             "enableTemporaryFileSystemWriteBuffer",
+                             enableTemporaryFilesystemWriteBuffer)) {
+            std::string errorMessage("Could not parse config due to: 'enableTemporaryFilesystemWriteBuffer' not found.");
+            log_warn() << errorMessage;
+        }
+        m_options->setEnableTemporaryFileSystemWriteBuffers(enableTemporaryFilesystemWriteBuffer);
+
+        if (enableTemporaryFilesystemWriteBuffer) {
+            int temporaryFileSystemSize = 100 * 1024 * 1024;
```
kursatkobya: so basically this is a default temporaryFileSystemSize. 
So i would prefer to have this value as a constexpr in class

frznlogic: It can't be a constexpr as it's set to a value on the line below this comment. The value here is just a "sane default" in case JSONParser fails for some reason.

kursatkobya: But it can be something like 
`int temporaryFileSystemSize = DEFAULT_TEMP_FSIZE;`
right

frznlogic: Ahh, I see, sorry I completely misunderstood you =).

agent/src/containeroptions/containeroptionparser.cpp
```diff
@@ -44,6 +45,30 @@ void ContainerOptionParser::readConfigElement(const json_t *element)
...
     m_options->setEnableWriteBuffer(wo);
+
+    if (wo == true)
+    {
+        bool enableTemporaryFilesystemWriteBuffer = false;
+        if(!JSONParser::read(element,
+                             "enableTemporaryFileSystemWriteBuffer",
+                             enableTemporaryFilesystemWriteBuffer)) {
+            std::string errorMessage("Could not parse config due to: 'enableTemporaryFilesystemWriteBuffer' not found.");
+            log_warn() << errorMessage;
```
kursatkobya: why don't we directly log_output the message like
log_warn() << "Could not parse config due to: 'enableTemporaryFilesystemWriteBuffer' not found.";

frznlogic: It was used for a DBusException before as well, I'll fix it.

kursatkobya (approved):
looks good

agent/src/containeroptions/containeroptionparser.cpp
```diff
@@ -44,6 +45,28 @@ void ContainerOptionParser::readConfigElement(const json_t *element)
     }
 
     m_options->setEnableWriteBuffer(wo);
+
+    if (wo == true)
```
kursatkobya: it would be nice if you can find better name instead wo

frznlogic: Done.

agent/src/containeroptions/dynamiccontaineroptions.cpp
```diff
@@ -40,4 +40,24 @@ bool DynamicContainerOptions::enableWriteBuffer() const
     return m_enableWriteBuffer;
 }
 
+void DynamicContainerOptions::setEnableTemporaryFileSystemWriteBuffers(bool enabled)
+{
+    m_enableTemporaryFileSystemWriteBuffers = enabled;
+}
+
+bool DynamicContainerOptions::enableTemporaryFileSystemWriteBuffers() const
```
thenor, line 48: Shouldn't this be called something like "isTemporaryFile...Enabled"?

frznlogic: it's a getter for the parameter named in the same manner as the other getters though?

thenor: I don't know if we have any code guidelines for this, but I am used to that getters for bools are "isVariable" and getters for other types are "getVariable". Not sure if we have had or should have that in SC.

agent/src/containeroptions/dynamiccontaineroptions.cpp
```diff
@@ -40,4 +40,24 @@ bool DynamicContainerOptions::enableWriteBuffer() const
...
+{
+    return m_enableTemporaryFileSystemWriteBuffers;
+}
+
+void DynamicContainerOptions::setTemporaryFileSystemSize(unsigned int size)
+{
+    m_temporaryFileSystemSize = size;
+}
+
+unsigned int DynamicContainerOptions::temporaryFileSystemSize() const
```
thenor, line 58: As my previous statement says I would prefer if this is called "getTemporary..."

agent/src/containeroptions/dynamiccontaineroptions.cpp
thenor: It's just that I find this method name confusing since it doesn't enable anything. But I understand that the variables name is "enable...".

agent/unit-test/containeroptionparser_unittest.cpp
```diff
@@ -63,6 +63,40 @@ TEST_F(ContainerOptionParserTest, parseConfigNice2) {
     ASSERT_FALSE(m_options->enableWriteBuffer());
 }
 
+TEST_F(ContainerOptionParserTest, parseConfigNice3) {
```
thenor: I would prefer if the names of the test cases were something a little more descriptive. Also for the first and second of the "Nice" test cases (and "Bad"). Just a preference. It makes it so much easier to understand what's going on in the test case without reading the full log.

doc/chapters/filesystem/index.rst
```diff
@@ -136,3 +136,36 @@ changes performed during its runtime to be merged into the lower layers.
 .. Note:: Non-directory types of files can not be mounted using overlayfs.
           These will automatically fall back on using the default behavior of 
           bind mounting the files into the filesystem of the container.
+
+Temporary Filesystem
+====================
+
+The Write buffers can be configured to use a separate temporary filesystem
```
thenor: Is there a reason for the upper case 'w' in "Write"?

agent/src/containeroptions/containeroptionparser.cpp
```diff
@@ -36,14 +37,36 @@ void ContainerOptionParser::readConfigElement(const json_t *element)
...
         std::string errorMessage("Could not parse config due to: 'enableWriteBuffer' not found.");
         log_error() << errorMessage;
         throw ContainerOptionParseError(errorMessage);
     }
 
-    m_options->setEnableWriteBuffer(wo);
+    m_options->setEnableWriteBuffer(enableWriteBuffer);
+
+    if (enableWriteBuffer == true)
+    {
```
breakreturn: Move this up one line to after the if

agent/src/containeroptions/containeroptionparser.cpp
```diff
@@ -36,14 +37,36 @@ void ContainerOptionParser::readConfigElement(const json_t *element)
...
+        bool enableTemporaryFilesystemWriteBuffer = false;
+        if(!JSONParser::read(element,
+                             "enableTemporaryFileSystemWriteBuffer",
+                             enableTemporaryFilesystemWriteBuffer)) {
```
breakreturn: Move the { to the next line.

agent/src/containeroptions/dynamiccontaineroptions.cpp
frznlogic: I understand your point ;) . We do have a lot of Qt coding style guidelines which I think are reasonable, and this naming does follow those ... or well, they don't mention it really, but their code is the same as this at least ;). Not sure what others feel like? @tobsan @breakreturn ?

doc/chapters/filesystem/index.rst
frznlogic: Nope. fixed.

agent/src/containeroptions/dynamiccontaineroptions.cpp
breakreturn: I don't think there are any guidelines. I'm also guessing we are mixing a bit, but the Qt style I believe is to have setX(), and just x() to get a property. A true/false check I do think is common to have isX() style though. I'd propose that @frznlogic simply do as how it looks in the code around here, and if there are no good references, check the rest of the code base.

doc/chapters/filesystem/index.rst
```diff
@@ -136,3 +136,36 @@ changes performed during its runtime to be merged into the lower layers.
...
+
+The ``enableTemporaryFileSystemWriteBuffer`` setting enables the ``tmpfs``
+creation as described above, while the ``temporaryFileSystemSize`` variable
+sets the size of the ``tmpfs`` in bytes.
+
+.. Note:: The ``temporaryFileSystemSize`` parameter will not be parsed unless
+          the ``enableTemporaryFileSystemWriteBuffer`` parameter is ``true``.
+          The ``temporaryFileSystemSize`` is not required if the
+          ``enableTemporaryFileSystemWriteBuffer`` is set to ``false`` or not
+          added at all.
```
thenor, line 167: I can't see that you have a test case for when the flag is set to false and no size is included.

agent/unit-test/containeroptionparser_unittest.cpp
frznlogic: Fair enough. Fixed naming.

breakreturn: On a glance, this looks like a good candidate for parametrization. If not, then I definitely think they should have proper names and comments on each of them what's being tested.

doc/chapters/filesystem/index.rst
```diff
@@ -136,3 +136,36 @@ changes performed during its runtime to be merged into the lower layers.
...
+
+.. Note:: The tmpfs is shared between the upper and work directories in overlayfs'
```
thenor: Is the ' at the end of the line a typo? Is it supposed to mean plural?

doc/chapters/filesystem/index.rst
frznlogic: There was not. I added a test now, thanks!

doc/chapters/filesystem/index.rst
```diff
@@ -136,3 +136,36 @@ changes performed during its runtime to be merged into the lower layers.
...
+====================
+
+The Write buffers can be configured to use a separate temporary filesystem
+(``tmpfs``) which can be limited in size. The ``tmpfs`` is mounted on top of
+the containers temporary directory as soon as it's created and will remain
+there until the container is destroyed. All ``overlayfs`` mounts and temporary
+directories should be created inside this ``tmpfs`` mount.
+
+The ``size`` of the ``tmpfs`` can also be limited using an extra configuration
+option in the ``Create DBus`` call to the ``SoftwareContainerAgent``.
```
breakreturn: Should "Create DBus" really by code highlighted? I guess you only want the method name to be that?

doc/chapters/filesystem/index.rst
frznlogic: Weird, it's supposed to be highlighted... it's meant to be in plural though.

servicetest/filesystem/test_filesystem.py
```diff
@@ -251,15 +251,94 @@ def test_tmpfs_writebuffer_size_enabled(self):
...
         """ The SoftwareContainer will create a tmpfs in the location of the
             containers temporary directory location if the write buffer is
             enabled. Test that the tmpfs mounted in the containers temporary
             directory structure works as expected.
 
-            Note: The filesystem is hard coded to 100MB at the time this test
-            was written, but will be configurable via the DATA field in a
-            short while
+            This test will create a 10 MB tmpfs which will be tested that it
+            behaves as expected.
```
breakreturn: What's expected?

servicetest/filesystem/test_filesystem.py
```diff
@@ -251,15 +251,94 @@ def test_tmpfs_writebuffer_size_enabled(self):
...
+            ca.terminate()
+
+        if os.path.exists(absoluteTestFile):
+            os.remove(absoluteTestFile)
+
+    @pytest.mark.xfail("platform.release() <= \"3.18.0\"")
+    def test_tmpfs_writebuffer_size_bad_config(self):
+        """ This test will test various bad configuration combinations and
+            make sure that SoftwareContainer-Agent behaves as expected when
+            the bad configuration is sent to it.
```
breakreturn: It's "SoftwareContainerAgent". And this comment doesn't really say what's going to be tested either, what's the expected behavior?

servicetest/filesystem/test_filesystem.py
```diff
@@ -251,15 +251,94 @@ def test_tmpfs_writebuffer_size_enabled(self):
...
+        # temporaryFileSystemSize defined
+        DATA[Container.CONFIG] = '[{"enableWriteBuffer": true, \
+                                 "enableTemporaryFileSystemWriteBuffer": true}]'
+
+        try:
+            ca.start(DATA)
+        finally:
+            ca.terminate()
+
+        # This run should work, but no temporary filesystem will be setup as
```
breakreturn: Why is this one test? It looks like two separate tests in one.

servicetest/filesystem/test_filesystem.py
```diff
@@ -251,15 +251,94 @@ def test_tmpfs_writebuffer_size_enabled(self):
...
+        # the enableTemporaryFileSystemWriteBuffer defaults to false and no
+        # parsing will be performed on the temporaryFileSystemSize parameter.
+        DATA[Container.CONFIG] = '[{"enableWriteBuffer": true, \
+                                 "temporaryFileSystemSize": 10485760}]'
```
breakreturn: What's that number? Is the exact number significant? I think it's better to have a variable.

servicetest/filesystem/test_filesystem.py
```diff
@@ -251,15 +251,94 @@ def test_tmpfs_writebuffer_size_enabled(self):
...
+            the bad configuration is sent to it.
+        """
+
+        absoluteTestFile = os.path.join("/tmp/container/SC-0/rootfs-upper",
+                                        TESTFILE)
+
+        ca = Container()
+
+        # This run should raise a DBus exception as we don't have a
+        # temporaryFileSystemSize defined
```
breakreturn: Where is this exception asserted?

breakreturn (changes requested):

agent/src/containeroptions/containeroptionparser.cpp
frznlogic: Fixed.

doc/chapters/filesystem/index.rst
frznlogic: Fixed.

servicetest/filesystem/test_filesystem.py
frznlogic: Files larger than 10MB can not be written and files smaller than 9MB can be written :). I'll add it.

servicetest/filesystem/test_filesystem.py
```diff
@@ -251,15 +251,94 @@ def test_tmpfs_writebuffer_size_enabled(self):
...
+        """
+
+        absoluteTestFile = os.path.join("/tmp/container/SC-0/rootfs-upper",
+                                        TESTFILE)
+
+        ca = Container()
+
+        DATA[Container.CONFIG] = '[{"enableWriteBuffer": true, \
+                                 "enableTemporaryFileSystemWriteBuffer": true, \
+                                 "temporaryFileSystemSize": 10485760}]'
```
breakreturn: What's this number?

servicetest/filesystem/test_filesystem.py
```diff
@@ -251,15 +251,94 @@ def test_tmpfs_writebuffer_size_enabled(self):
...
+            partitions = psutil.disk_partitions(all=True)
+            interesting_partition = False
+            for part in partitions:
+                if part.mountpoint == '/tmp/container/SC-0':
+                    interesting_partition = part
+                    break
+            assert interesting_partition is not False
+
+            with open(absoluteTestFile, "w") as f:
+                f.write("w" * 9 * 1024 * 1024)
```
breakreturn: What are these number and the ones below?

breakreturn (changes requested):

servicetest/filesystem/test_filesystem.py
```diff
@@ -251,15 +251,94 @@ def test_tmpfs_writebuffer_size_enabled(self):
...
+
+        absoluteTestFile = os.path.join("/tmp/container/SC-0/rootfs-upper",
+                                        TESTFILE)
+
+        ca = Container()
+
+        # This run should raise a DBus exception as we don't have a
+        # temporaryFileSystemSize defined
+        DATA[Container.CONFIG] = '[{"enableWriteBuffer": true, \
+                                 "enableTemporaryFileSystemWriteBuffer": true}]'
```
thenor: Where do you check for the DBus exception?

servicetest/filesystem/test_filesystem.py
frznlogic: Added/fixed and renamed the test.

servicetest/filesystem/test_filesystem.py
frznlogic: I changed the behavior during writing the tests and forgot to update the comments. Sorry, it's fixed now.

servicetest/filesystem/test_filesystem.py
frznlogic: No good reason to be honest except trying to shorten down runtime. I'll break it out.

servicetest/filesystem/test_filesystem.py
frznlogic: it's 10 \* 1024 \* 1024 or 10 megabytes.

servicetest/filesystem/test_filesystem.py
frznlogic: it's 10 \* 1024 \* 1024

servicetest/filesystem/test_filesystem.py
frznlogic: It's 9 megabytes. Just a random number =)

frznlogic: Comments have been added.

servicetest/filesystem/test_filesystem.py
```diff
@@ -251,15 +259,119 @@ def test_tmpfs_writebuffer_size_enabled(self):
...
+                                "enableTemporaryFileSystemWriteBuffer": true,\
+                                "temporaryFileSystemSize": ' + str(TenMB) +
+                                '}]'
+        try:
+            ca.start(DATA)
+
+            partitions = psutil.disk_partitions(all=True)
+            interesting_partition = False
+            for part in partitions:
+                if part.mountpoint == '/tmp/container/SC-0':
```
thenor: Is it guaranteed that the ID for the container will be "SC-0"? I think there are other test cases that fetch the ID instead of hardcoding it.

servicetest/filesystem/test_filesystem.py
```diff
@@ -251,15 +259,119 @@ def test_tmpfs_writebuffer_size_enabled(self):
...
+        absoluteTestFile = os.path.join("/tmp/container/SC-0/rootfs-upper",
+                                        TESTFILE)
+
+        ca = Container()
+
+        DATA[Container.CONFIG] = '[{"enableWriteBuffer": true, \
+                               "temporaryFileSystemSize": ' + str(TenMB) + '}]'
+
+        try:
+            ca.start(DATA)
```
thenor: Will this also verify that it was successful? Will this fail if it is not possible?

thenor (changes requested):
My comments on the code are not blocking the merge to master, in my opinion, but the comments on the service tests I would like to get answers for before approval.

servicetest/filesystem/test_filesystem.py
kursatkobya: service tests for cgroups had that i guess

frznlogic: It's a good idea. That said, we have similar problems in a lot of places in the testing code, but that's no reason to make it worse i guess ;). Fixed.

agent/src/containeroptions/containeroptionparser.cpp
```diff
@@ -36,14 +37,35 @@ void ContainerOptionParser::readConfigElement(const json_t *element)
...
     }
 
-    m_options->setEnableWriteBuffer(wo);
+    m_options->setEnableWriteBuffer(enableWriteBuffer);
+
+    if (enableWriteBuffer == true) {
+        bool enableTemporaryFilesystemWriteBuffer = false;
+        if(!JSONParser::read(element,
+                             "enableTemporaryFileSystemWriteBuffer",
+                             enableTemporaryFilesystemWriteBuffer)) {
```
breakreturn: The '{' goes on the next line when the if is broken up like this, as I mentioned in previous review.

agent/src/containeroptions/containeroptionparser.cpp
```diff
@@ -36,14 +37,35 @@ void ContainerOptionParser::readConfigElement(const json_t *element)
...
+            log_warn() << "Could not parse config due to: 'enableTemporaryFilesystemWriteBuffer' not found.";
+        }
+        m_options->setEnableTemporaryFileSystemWriteBuffers(enableTemporaryFilesystemWriteBuffer);
+
+        if (enableTemporaryFilesystemWriteBuffer) {
+            int temporaryFileSystemSize = DEFAULT_TEMPORARY_FILESYSTEM_SIZE;
+            if(!JSONParser::read(element,
+                                 "temporaryFileSystemSize",
+                                 temporaryFileSystemSize)) {
```
breakreturn: Same here with the {

agent/unit-test/containeroptionparser_unittest.cpp
```diff
@@ -58,11 +58,45 @@ TEST_F(ContainerOptionParserTest, parseManyTimes) {
     ASSERT_TRUE(m_options->enableWriteBuffer());
 }
 
-TEST_F(ContainerOptionParserTest, parseConfigNice2) {
+TEST_F(ContainerOptionParserTest, parseConfigNiceDisabled) {
     ASSERT_NO_THROW(parse("[{\"enableWriteBuffer\": false}]"));
     ASSERT_FALSE(m_options->enableWriteBuffer());
 }
 
+TEST_F(ContainerOptionParserTest, parseConfigNiceWithTmpfsAndSize) {
```
breakreturn: I think all tests should have a comment what they test, what the values mean, what the purpose is etc. Please comment the tests. If it's deemed overkill to have a comment on each, unlikely IMO, then at least a comment before the batch of functions that explains the overall strategy.

servicetest/filesystem/test_filesystem.py
```diff
@@ -251,29 +262,159 @@ def test_tmpfs_writebuffer_size_enabled(self):
...
+                                            "/",
+                                            TESTFILE)
+
+            with open(absoluteTestFile, "w") as f:
+                f.write("w" * 95 * 1024 * 1024)
+
+            # Make sure the file is between 94 and 96 megabyte (exact size
+            # may vary depending on platform, blocksize etc)
+            assert os.path.getsize(absoluteTestFile) >= 94 * 1024 * 1024
+            assert os.path.getsize(absoluteTestFile) <= 96 * 1024 * 1024
```
breakreturn, line 387: Again, why not have descriptive names instead of values like this?

breakreturn (changes requested):

agent/src/containeroptions/containeroptionparser.cpp
frznlogic: Fixed. However, these rules are counterintuitive imho. I'd prefer if we made them more intuitive.

servicetest/filesystem/test_filesystem.py
breakreturn: @frznlogic In this diff I actually missed the comment above, sorry! I think this is explanatory enough, the "magic" numbers aren't magic anymore with the above comments. Sorry.

frznlogic: The comments above where added to alleviate the understanding of the values. I hope those are enough?

agent/src/containeroptions/dynamiccontaineroptions.cpp
frznlogic: I'm staying with the Qt style as discussed at todays meetings but will look into renaming the parameters.

servicetest/filesystem/test_filesystem.py
breakreturn: Yes it looks perfectly fine to me.

breakreturn (approved):

thenor (approved):

PR #418, opened 2017-04-13
servicetest: Save the container ID

frznlogic:
The container ID should be saved in the Container class so it can be
reused by other functions.


breakreturn (approved):

kursatkobya (approved):
save the containerID, save the world

frznlogic:
self.__container_id is actually set in __create_container method call. My bad.

PR #420, opened 2017-04-18
Testing robustness of bindMount

thenor:
Add more test cases to verify and cover more cases for bindMount.

### DIFF
--- a/libsoftwarecontainer/component-test/filegateway_componenttest.cpp
+++ b/libsoftwarecontainer/component-test/filegateway_componenttest.cpp
@@ -23,6 +23,7 @@
 #include "functionjob.h"
 #include <unistd.h>
 #include <stdlib.h>
+#include <fstream>
 
 class FileGatewayTest : public SoftwareContainerGatewayTest
 {
@@ -37,27 +38,48 @@ public:
 
         gw = std::unique_ptr<FileGateway>(new FileGateway(m_container));
 
-        // Create file
-        ASSERT_TRUE(writeToFile(FILE_PATH, FILE_CONTENT));
+        // Create files
+        assert(mkdir(DIR_PATH.c_str(), 775) == 0);
+        assert(createFile(FILE_PATH, FILE_CONTENT));
+        assert(createFile(LONG_FILE_PATH, FILE_CONTENT_2));
+        assert(createFile(FILE_PATH_W_SPACES, FILE_CONTENT_2));
     }
 
     void TearDown() override
     {
         // Remove file
         unlink(FILE_PATH.c_str());
+        unlink(LONG_FILE_PATH.c_str());
+        unlink(FILE_PATH_W_SPACES.c_str());
+        rmdir(DIR_PATH.c_str());
 
         SoftwareContainerGatewayTest::TearDown();
     }
 
+    bool createFile(const std::string &path, const std::string &content)
+    {
+        std::ofstream out(path);
+        if (out.is_open()) {
+            out << content;
+            return out.good();
+        }
+        return false;
+    }
+
     const std::string FILE_CONTENT = "ahdkhqweuyreqiwenomndlaskmd";
     const std::string FILE_PATH = "/tmp/filename.txt";
+    const std::string DIR_PATH = "/tmp/testdir/";
+    const std::string LONG_FILE_PATH = DIR_PATH + "filename.txt";
+    const std::string FILE_CONTENT_2 = "bgsajbjkbjherkgbjhsbnjdfnjk";
+    const std::string FILE_W_SPACES = "example file name.txt";
+    const std::string FILE_PATH_W_SPACES = "/tmp/" + FILE_W_SPACES;
     const std::string CONTAINER_PATH = "/filename.txt";
 };
 
 /*
  * Test that a minimal conf is accepted and that activate works
  */
-TEST_F(FileGatewayTest, TestActivateWithMinimalValidConf) {
+TEST_F(FileGatewayTest, ActivateWithMinimalValidConf) {
     const std::string config =
     "["
         "{"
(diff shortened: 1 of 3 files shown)


libsoftwarecontainer/component-test/filegateway_componenttest.cpp
```diff
@@ -69,3 +69,43 @@ TEST_F(FileGatewayTest, TestActivateWithMinimalValidConf) {
     ASSERT_TRUE(gw->setConfig(jsonConfig));
     ASSERT_TRUE(gw->activate());
 }
+
+TEST_F(FileGatewayTest, MalformedPath) {
+// Mount a path which is malformed, e.g. too many "/"
```
breakreturn: I think the comment should either be indented along with the code inside the test, or located above the test if it's a comment for the whole test. Same goes for some other places below as well.

libsoftwarecontainer/component-test/filegateway_componenttest.cpp
```diff
@@ -69,3 +69,43 @@ TEST_F(FileGatewayTest, TestActivateWithMinimalValidConf) {
...
+    const std::string config =
+        "[{"
+        "  \"path-host\": \"//tmp//filename.txt\""
+        ", \"path-container\": \"///filename.txt\""
+        "}]";
+    loadConfig(config);
+
+    // Is this ok? Will the extra "/" will be removed?
```
breakreturn: Is this a "todo" comment, or is it a comment for what's being asserted?

libsoftwarecontainer/component-test/filegateway_componenttest.cpp
```diff
@@ -69,3 +69,43 @@ TEST_F(FileGatewayTest, TestActivateWithMinimalValidConf) {
...
+    ASSERT_TRUE(gw->setConfig(jsonConfig));
+    ASSERT_TRUE(gw->activate());
+}
+
+TEST_F(FileGatewayTest, MountNonExisting) {
+// Mount a non-existing path
```
breakreturn: Is this a comment for what the test does? Put it above the test in that case, and comment on what's being tested, e.g. "Mount a non-existing path and assert that ... "

libsoftwarecontainer/component-test/filegateway_componenttest.cpp
```diff
@@ -69,3 +69,43 @@ TEST_F(FileGatewayTest, TestActivateWithMinimalValidConf) {
...
+            ", \"path-container\": \"/tmp/thisFileDoesNotExist.txt\""
+        "}]";
+    loadConfig(config);
+
+    ASSERT_TRUE(gw->setConfig(jsonConfig));
+    ASSERT_FALSE(gw->activate());
+}
+
+TEST_F(FileGatewayTest, BadMount) {
+// Mount a non-existing path
```
breakreturn: Copy pasted from above? Same comment as for above applies here I think.

breakreturn (changes requested):

libsoftwarecontainer/component-test/filegateway_componenttest.cpp
thenor: Yes this is more or less on the level of a TODO. I saw the printout during the test and couldn't see if this path (with extra slashes) was what was being using for the bindmount command. I wanted to check up if we remove extra slashes (couldn't find it in our code, so perhaps it is the help functions we use). So it was a reminder to myself to check up on it.

libsoftwarecontainer/component-test/filegateway_componenttest.cpp
thenor: Yes, agreed. I wrote it in the beginning of the test case to specify what to test, then forgot to redo it as a test header/description.

libsoftwarecontainer/component-test/filegateway_componenttest.cpp
thenor: Yes, will redo all the comments and move them.

libsoftwarecontainer/component-test/filegateway_componenttest.cpp
```diff
@@ -37,27 +38,54 @@ class FileGatewayTest : public SoftwareContainerGatewayTest
 
         gw = std::unique_ptr<FileGateway>(new FileGateway(m_container));
 
-        // Create file
-        ASSERT_TRUE(writeToFile(FILE_PATH, FILE_CONTENT));
+        // Create files
+        ASSERT_EQ(mkdir(DIR_PATH.c_str(),775),0);
```
breakreturn: There should be spaces after the commas.

libsoftwarecontainer/component-test/filegateway_componenttest.cpp
```diff
@@ -37,27 +38,54 @@ class FileGatewayTest : public SoftwareContainerGatewayTest
...
+        ASSERT_TRUE(createFile(FILE_PATH, FILE_CONTENT));
```
breakreturn: I'm not a fan of testing things in the test setup. This also looks like it's actually testing the test setup code. If the test should break if this doesn't work I think some normal assert should work better.

libsoftwarecontainer/component-test/filegateway_componenttest.cpp
```diff
@@ -69,3 +97,117 @@ TEST_F(FileGatewayTest, TestActivateWithMinimalValidConf) {
...
+ */
+TEST_F(FileGatewayTest, MalformedPath) {
+    const std::string config =
+        "[{"
+        "  \"path-host\": \"//tmp//filename.txt\""
+        ", \"path-container\": \"///filename.txt\""
+        "}]";
+    loadConfig(config);
+
+    // TODO: Verify that this is not a false positive
```
breakreturn: Should this be resolved or is it intentionally left?

libsoftwarecontainer/component-test/filegateway_componenttest.cpp
thenor: Ok. I will change it to use the regular c++ assert.

libsoftwarecontainer/component-test/filegateway_componenttest.cpp
thenor: I forgot to remove it after I had tested this.

breakreturn (approved):

libsoftwarecontainer/component-test/filegateway_componenttest.cpp
```diff
@@ -37,27 +38,54 @@ class FileGatewayTest : public SoftwareContainerGatewayTest
...
         SoftwareContainerGatewayTest::TearDown();
     }
 
+    bool createFile(const std::string &path, const std::string &content)
+    {
+        bool ret = true;
+        std::ofstream out(path);
+        if (out.is_open()) {
+            out << content;
+            if (!out.good()) {
```
kursatkobya: just an option, you can get rid of ret variable just by returning good method to have concise function
```
std::ofstream out(path);
if (out.is_open()) {
    out << content;
    return out.good();
}

return false;
```

libsoftwarecontainer/component-test/filegateway_componenttest.cpp
```diff
@@ -37,27 +38,54 @@ class FileGatewayTest : public SoftwareContainerGatewayTest
...
+        return ret;
+    }
+
     const std::string FILE_CONTENT = "ahdkhqweuyreqiwenomndlaskmd";
     const std::string FILE_PATH = "/tmp/filename.txt";
+    const std::string DIR_PATH = "/tmp/testdir/";
+    const std::string LONG_FILE_PATH = DIR_PATH + "filename.txt";
+    const std::string FILE_CONTENT_2 = "bgsajbjkbjherkgbjhsbnjdfnjk";
+    const std::string FILE_W_SPACES = "example file name.txt";
+    const std::string FILE_PATH_W_SPACES = "/tmp/example file name.txt";
```
kursatkobya: this looks like filename with spaces to me.  Adding space after tmp would be more explanatory.
`/tmp /example file/name.txt`

kursatkobya (approved):
looks nice.
approved with small comments

libsoftwarecontainer/component-test/filegateway_componenttest.cpp
thenor: But that would be a new directory: "tmp\ "
What I wanted was just "/tmp/"+FILE_W_SPACES, whic I could have written.

libsoftwarecontainer/component-test/filegateway_componenttest.cpp
thenor: Oh yes of course. Good catch.

thenor: Hm well... I want to do the out.close also.

kursatkobya: I think ofstream destructor closes the file but sure

thenor: Tbh I copied this code from common and didn't analyze it too much. Maybe you are right.

frznlogic (approved):

libsoftwarecontainer/component-test/filegateway_componenttest.cpp
```diff
@@ -37,27 +38,48 @@ class FileGatewayTest : public SoftwareContainerGatewayTest
...
 
+    bool createFile(const std::string &path, const std::string &content)
+    {
+        std::ofstream out(path);
+        if (out.is_open()) {
+            out << content;
+            return out.good();
+        }
+        return false;
+    }
```
kursatkobya, line 67: nice !!

PR #421, opened 2017-05-04
Add uncrustify config

thenor:
Create a basic uncrustify config which matches the C++ formatting guidelines.
Create a script that executes uncrustify on the cpp, h and c files in the main code dirs.


kursatkobya (approved):

CMakeLists.txt
```diff
@@ -304,6 +307,10 @@ if(ENABLE_DBUSGATEWAY)
     install(PROGRAMS ${CMAKE_BINARY_DIR}/dbus-proxy/bin/dbus-proxy DESTINATION bin)
 endif()
 
+if(ENABLE_UNCRUSTIFY)
+    ## TODO: change this to use a config source
```
breakreturn: Is this left intentionally? What does it mean?

servicetest/timingprofiling/simple.c
```diff
@@ -21,6 +20,7 @@
 
 #include <unistd.h>
 
-int main(int argc, char **argv) {
+int main(int argc, char * *argv)
```
breakreturn: This should not be a part of this commit right?

Also, if this is how the config wants the code to look, could we configure that so it wants it to look like we want it?

breakreturn (changes requested):

servicetest/timingprofiling/simple.c
thenor: It was included by mistake. I'm removing it from the commit.

CMakeLists.txt
thenor: This was left intentionally since I didn't start using this flag. I could remove the change completely since it doesn't do anything at the moment.

breakreturn (approved):

scripts/format_sources.sh
```diff
@@ -0,0 +1,73 @@
...
+# OR ANY DAMAGES WHATSOEVER RESULTING FROM LOSS OF USE, DATA OR PROFITS,
+# WHETHER IN AN ACTION OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION,
+# ARISING OUT OF OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS
+# SOFTWARE.
+#
+# For further information see LICENSE
+#
+
+#
+# This is a helper script for invoking uncrustify on a set of C++ files.
```
frznlogic: These comments are wrong. The output is stored in a prefix

scripts/format_sources.sh
```diff
@@ -0,0 +1,73 @@
...
+
+echo "Formatting with Uncrustify configuration file : $UNCRUSTIFY_CONFIG_FILE"
+echo "Running uncrustify starting at $COMPONENT_FOLDER"
+
+## This will create a dir formatting_ouput with all the uncrustify result files
+UNCRUSTIFY_COMMAND="uncrustify -c $UNCRUSTIFY_CONFIG_FILE --prefix formatting_output 'file'"
+
+read -p "We are going to format the C/C++ (*.cpp, *.c, *.h) source files from the folder
+'$COMPONENT_FOLDER'. We recommend that you first commit your files to your local GIT repo
+before going further, since the formatter will modify your files. You will be able to amend
```
frznlogic: This is also not true... we are not going to overwrite anything as far as I can tell.

scripts/format_sources.sh
```diff
@@ -0,0 +1,73 @@
...
+# This is a helper script for invoking uncrustify on a set of C++ files.
+# Invoking this script on a directory containing C++ files (*.cpp, *.h, *.c)
+# will format these files _in place_, according to an uncrustify config file.
+#
+# The $UNCRUSTIFY_CONFIG_FILE environment variable can be used to specify which
+# uncrustify config file to use.
+#
+
+COMPONENT_FOLDER="" # Set to softwarecontainer base dir
+UNCRUSTIFY_CONFIG_FILE="" # Add option to set this
```
frznlogic: Do we want to add support for reading this from command line?

frznlogic (changes requested):

frznlogic:
The comments and the read -p command are wrong with regards to where data is stored.

scripts/format_sources.sh
thenor: Oops copy paste error. Thanks for catching it.

scripts/format_sources.sh
thenor: I had some idea that it would use the default config file unless specified, but that another config could be used with a flag/argument to "make uncrustify".

scripts/format_sources.sh
thenor: Same as above. I forgot to fix the comment.
Also: when we change to use this with make or in the CI system we probably don't want this prompt, since it waits for an answer.

thenor: Should I remove it completely, perhaps?

scripts/format_sources.sh
frznlogic: Np:)

scripts/format_sources.sh
frznlogic: I think so... it simplifies using the script automatically if nothing else.

scripts/format_sources.sh
```diff
@@ -0,0 +1,63 @@
...
+# Invoking this script will find all C++ files (*.cpp, *.h, *.c) in
+# the 'agent', 'common' and 'libsoftwarecontainer' directories and format the
+# files according to an uncrustify config file.
+# The resulting files will be put in a directory called 'formatting_output'.
+#
+# The $UNCRUSTIFY_CONFIG_FILE environment variable can be used to specify which
+# uncrustify config file to use.
+#
+
+UNCRUSTIFY_LOGFILE="uncrustify_output.log" # Currently not used
```
frznlogic: Could you make the same pattern as the other variables for this one?

frznlogic (changes requested):

scripts/format_sources.sh
thenor: Of course!

frznlogic (approved):

PR #422, opened 2017-07-24
agent: only remove containers created by SC

tobsan:
Adds a simple regex that checks for container names matching SC-* and if
they don't match - don't remove them.

### DIFF
--- a/agent/src/containerutilityinterface.cpp
+++ b/agent/src/containerutilityinterface.cpp
@@ -22,8 +22,8 @@
 #include "softwarecontainer-common.h"
 #include "createdir.h"
 
-
 #include <lxc/lxccontainer.h>
+#include <regex>
 
 namespace softwarecontainer {
 ContainerUtilityInterface::ContainerUtilityInterface(std::shared_ptr<Config> config)
@@ -39,43 +39,54 @@ void ContainerUtilityInterface::removeOldContainers()
     const char *basePath = lxc_get_global_config_item("lxc.lxcpath");
     auto num = list_all_containers(basePath, &containerNames, &containerList);
 
-    if (0 == num) {
-        // there are no container residue. No need to run more.
-        delete containerList;
+    if (-1 == num) {
+        log_error() << "An error occurred while trying to get container list";
+        throw ContainerUtilityInterfaceError("An error occurred while trying to get container list");
+    } else if (0 == num) {
         delete containerNames;
+        delete containerList;
         return;
-    } else if (-1 == num) {
-        log_error() << "An error is occurred while trying to get deprecated container list";
-        throw ContainerUtilityInterfaceError("An error is occurred while trying to get container list");
     }
 
-    log_warning() << num << " unused deprecated containers found";
-    for (auto i = 0; i < num; i++) {
+    // Convert the list/names to a map for convenience
+    std::map<std::string, struct lxc_container *> containerMap;
+    for(int i=0; i<num; i++) {
+        std::string name = std::string(containerNames[i]);
         struct lxc_container *container = containerList[i];
-        log_debug() << "Deprecated container named " << containerNames[i] << " will be deleted";
+
+        containerMap[name] = container;
+        delete containerNames[i]; // We don't need this anymore
+    }
+    delete containerNames; // We don't need this anymore
+
+    for(const auto value : containerMap) {
+        std::string containerName = value.first;
+        struct lxc_container *container = value.second;
+
+        if (!std::regex_match(containerName, std::regex("SC-.*"))) {
+            log_debug() << "Found non-SC container " << containerName;
+            continue;
+        }
+
+        log_debug() << "Deprecated container named " << containerName << " will be deleted";
 
         if (container->is_running(container)) {
             bool success = container->stop(container);
             if (!success) {
-                std::string errorMsg = "Unable to stop deprecated container " +
-                                       std::string(containerNames[i]);
+                std::string errorMsg = "Unable to stop deprecated container " + containerName;
                 throw ContainerUtilityInterfaceError(errorMsg);
             }
         }
 
         bool success = container->destroy(container);
         if (!success) {
-            std::string errorMsg = "Unable to destroy deprecated container " +
-                                   std::string(containerNames[i]);
+            std::string errorMsg = "Unable to destroy deprecated container " + containerName;
             throw ContainerUtilityInterfaceError(errorMsg);
         }
 
-        log_debug() << "Deprecated container " << containerNames[i] << " is successfully destroyed";
-        delete container;
-        delete containerNames[i];
+        log_debug() << "Deprecated container " << containerName << " is successfully destroyed";
+        delete value.second; // Remove the lxc_container struct that we allocated
     }
-    delete containerList;
-    delete containerNames;
 }
 
 void ContainerUtilityInterface::checkWorkspace()
--- a/servicetest/agent/test_agent.py
+++ b/servicetest/agent/test_agent.py
@@ -133,8 +133,11 @@ class TestClearOldContainers(object):
 
         other_name = "agent-servicetest"
         other_counter = 0
-        sc_containers = dict()
+        sc_containers = 0
         other_containers = []
+
+        # Amount of containers prior to test start
+        initial_container_count = len(lxc.list_containers())
 
         try:
             for contype in containers:
@@ -143,7 +146,7 @@ class TestClearOldContainers(object):
                     sc = Container()
                     container_id = sc.start(DATA)
                     container_name = "SC-" + str(container_id)
-                    sc_containers[container_name] = sc
+                    sc_containers += 1
                 # Create just any LXC container
                 else:
                     skipped_container = self.__create_lxc_container(other_name + str(other_counter))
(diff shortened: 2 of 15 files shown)


agent/src/containerutilityinterface.cpp
```diff
@@ -52,25 +52,29 @@ void ContainerUtilityInterface::removeOldContainers()
...
-            std::string errorMsg = "Unable to destroy deprecated container " +
-                                   std::string(containerNames[i]);
+            std::string errorMsg = "Unable to destroy deprecated container " + containerName;
             throw ContainerUtilityInterfaceError(errorMsg);
         }
 
-        log_debug() << "Deprecated container " << containerNames[i] << " is successfully destroyed";
+        log_debug() << "Deprecated container " << containerName << " is successfully destroyed";
         delete container;
         delete containerNames[i];
```
DunderRoffe: Why are we using two lists instead of a map?

servicetest/agent/test_agent.py
```diff
@@ -93,23 +94,41 @@ def test_old_containers_cleanup(self):
...
+            existing_containers.append(line.strip())
+        assert proc.wait() == 0
+
+        assert existing_containers.count(container_name) == 0
+        assert existing_containers.count(skipped_container_name) == 1
+
+        # Exit and cleanup
         agent_handler.terminate()
-        os.remove("output")
+        os.system("lxc-destroy -n " + skipped_container_name)
```
DunderRoffe: This cleanup will be skipped it there is an exception (or assertion error) in the code above.

servicetest/agent/test_agent.py
```diff
@@ -93,23 +94,41 @@ def test_old_containers_cleanup(self):
...
 
-        os.system("lxc-ls >> output")
-        with open("output", "r") as fh:
-            file_content = fh.readline()
-        file_content = file_content.strip()
-        assert file_content == ("SC-" + str(container_id))
-        os.remove("output")
+        skipped_container_name = "agent-servicetest"
+        # Create one that SC shouldn't touch
+        os.system("lxc-create -t busybox -n " + skipped_container_name)
```
DunderRoffe: It might be neater to use the lxc python module for this

DunderRoffe (changes requested):
Looks legit. However, I would like to see the following tests as well:

* Starting more than 1 SC container and 1 non-SC container
* Starting them in different orders
* Starting no SC containers

agent/src/containerutilityinterface.cpp
tobsan: Because the lists are populated through a call to the LXC API.

DunderRoffe: sure but that does not mean that we need to keep the data stored that way right?

tobsan: No but should we iterate over the data and store it in a different way just because it is convenient?

servicetest/agent/test_agent.py
tobsan: Working on it. The LXC bindings for Python are Python3 only, which means some refactoring and updating of the service test code.

servicetest/agent/test_agent.py
tobsan: The whole test should have a try/finally

agent/src/containerutilityinterface.cpp
DunderRoffe: Convenience sounds like a good reason yes :)
Also it would let you use c++11 style for loops

servicetest/dbus/dbusapp.py
```diff
@@ -60,7 +60,7 @@ def Bounce(self, message):
 
     @dbus.service.method(IFACE, in_signature="s", out_signature="s")
     def Ping(self, message):
-        print "Got a Ping: ", message
+        print(("Got a Ping: ", message))
```
DunderRoffe: I prefer `print("Got a Ping {}".format(message))` as it explicitly shows where the message will be inserted

servicetest/dbus/dbusapp.py
```diff
@@ -153,7 +154,7 @@ def call_ping(self):
         r = Server(args.outdir)
         r.start()
     elif args.mode == "client":
-        print "Using mode 'client'"
-        print "method: ", args.method
+        print("Using mode 'client'")
+        print(("method: ", args.method))
```
DunderRoffe: I prefer `print("method: {}".format(args.method))`

servicetest/dbus/test_dbus.py
```diff
@@ -274,9 +275,9 @@ def test_spam_out(self):
             assert serv.wait_until_requests(multiplier=clients) is True
             t2 = time.time()
             print("\n")
-            print("Clients started:              {0:.4f} seconds\n".format(t1 - t0))
-            print("Server received all messages: {0:.4f} seconds\n".format(t2 - t1))
-            print("Total time:                   {0:.4f} seconds\n".format(t2 - t0))
+            print(("Clients started:              {0:.4f} seconds\n".format(t1 - t0)))
+            print(("Server received all messages: {0:.4f} seconds\n".format(t2 - t1)))
+            print(("Total time:                   {0:.4f} seconds\n".format(t2 - t0)))
```
DunderRoffe: Redundant parentheses

servicetest/filesystem/test_filesystem_strace.py
```diff
@@ -178,7 +178,7 @@ def test_bindmount_sockets_strace(self):
 
         # Remove the fileset files from the files list
         for f in fileset:
-            files = filter(lambda a: a != f, files)
+            files = [a for a in files if a != f]
```
DunderRoffe, line 181: A bit weird to mix standard for loop notation with a list comprehension

servicetest/timingprofiling/test_profiling.py
```diff
@@ -93,7 +93,7 @@ def run_test(num_starts=3):
         for app in range(0, num_starts):
             """ Start numStarts apps in softwarecontainer.
             """
-            print "Start app " + str(app)
+            print("Start app " + str(app))
```
DunderRoffe: Use .format instead of +
https://softwareengineering.stackexchange.com/questions/304445/why-is-s-better-than-for-concatenation

servicetest/timingprofiling/test_profiling.py
```diff
@@ -134,7 +134,7 @@ def get_python_point(log_file, point_name, match_number=1):
             match = match + 1
             if match == match_number:
                 return remove_ansi(string.split(line)[2])
-    print "Nothing found! " + point_name + " " + str(match_number)
+    print("Nothing found! " + point_name + " " + str(match_number))
```
DunderRoffe: Use .format instead of +
https://softwareengineering.stackexchange.com/questions/304445/why-is-s-better-than-for-concatenation

servicetest/timingprofiling/test_profiling.py
```diff
@@ -145,9 +145,9 @@ def get_function_log(log_file, function_name, match_number=1):
         if re.search(function_name + " end", line):
             match = match + 1
             if match == match_number:
-                print line
+                print(line)
                 return remove_ansi(string.split(line)[6])
-    print "Nothing found! " + function_name + " " + str(match_number)
+    print("Nothing found! " + function_name + " " + str(match_number))
```
DunderRoffe: Use .format instead of +
https://softwareengineering.stackexchange.com/questions/304445/why-is-s-better-than-for-concatenation

servicetest/timingprofiling/test_profiling.py
```diff
@@ -165,7 +165,7 @@ def get_log_point(log_file, point_name, match_number=1):
             match = match + 1
             if match == match_number:
                 return remove_ansi(string.split(line)[4])
-    print "Nothing found! " + point_name + " " + str(match_number)
+    print("Nothing found! " + point_name + " " + str(match_number))
```
DunderRoffe: Use .format instead of +
https://softwareengineering.stackexchange.com/questions/304445/why-is-s-better-than-for-concatenation

DunderRoffe (changes requested):

servicetest/filesystem/test_filesystem_strace.py
tobsan: This was suggested by 2to3, actually.

DunderRoffe (approved):

servicetest/testframework/lib.py
```diff
@@ -441,7 +441,8 @@ def is_alive(self):
         return self.__agent.poll() is None
 
     def terminate(self, kill=False):
-        if self.is_alive():
+        while self.is_alive():
+#        if self.is_alive():
```
DunderRoffe: Please clean up comment

servicetest/agent/test_agent.py
```diff
@@ -133,17 +133,20 @@ def test_old_containers_cleanup(self, containers):
...
+
         try:
             for contype in containers:
                 # Create an SC container
                 if contype == self.CON_TYPE.SC:
                     sc = Container()
                     container_id = sc.start(DATA)
                     container_name = "SC-" + str(container_id)
-                    sc_containers[container_name] = sc
+                    sc_containers += 1
```
DunderRoffe: Make sure that sc is cleaned up if there is a failure

DunderRoffe (changes requested):

DunderRoffe (approved):

PR #428, opened 2019-02-05
update lxc config element names

TSAnsari:
Several of the lxc config element names have been
renamed in lxc 3 version. This results in failing
to load lxc configuration after upgrading the lxc
version. This fixes the issue by renaming the
config element names to the newer names.

Signed-off-by: Tariq Ansari <tansari@luxoft.com>


thenor (approved):

thenor:
Seems the build doesn't work in Jenkins. Master isn't building successfully either.

sashko:
lxc: update...

thenor:
If I'm going to be picky I would prefer that all mentions of LXC which are not code should be uppercase, since it's an abbreviation.

PR #430, opened 2019-04-03
lxc: update softwarecontainer to use lxc 3.1.0

TSAnsari:
Signed-off-by: Tariq Ansari <tansari@luxoft.com>

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -35,7 +35,7 @@ include(Configs)
 # Set software version
 #
 set(${PROJECT_NAME}_MAJOR_VERSION 0)
-set(${PROJECT_NAME}_MINOR_VERSION 18)
+set(${PROJECT_NAME}_MINOR_VERSION 19)
 set(${PROJECT_NAME}_PATCH_LEVEL 0)
 set(VERSION ${${PROJECT_NAME}_MAJOR_VERSION}.${${PROJECT_NAME}_MINOR_VERSION}.${${PROJECT_NAME}_PATCH_LEVEL})
 add_definitions(-DPACKAGE_VERSION="${VERSION}")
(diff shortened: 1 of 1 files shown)


fhajredini (approved):

fhajredini:
The version is mentioned here as well. https://github.com/Pelagicore/vagrant-cookbook/blob/master/deps/softwarecontainer-dependencies.sh, does this need to point to master now or is the hard coded version ok?

TSAnsari:
> The version is mentioned here as well. https://github.com/Pelagicore/vagrant-cookbook/blob/master/deps/softwarecontainer-dependencies.sh, does this need to point to master now or is the hard coded version ok?

In the above cookbook link it is fetching and installing from lxc 3.1.0 branch/tag to be used by softwarecontainer. It should point to 3.1.0 tag.

In this commit it only updates CMake so that it checks for the 3.1.0 installed version. The reason being that softwarecontainer was updated to work with 3.1.0 or higher versions.

sashko:
> > The version is mentioned here as well. https://github.com/Pelagicore/vagrant-cookbook/blob/master/deps/softwarecontainer-dependencies.sh, does this need to point to master now or is the hard coded version ok?
> 
> In the above cookbook link it is fetching and installing from lxc 3.1.0 branch/tag to be used by softwarecontainer. It should point to 3.1.0 tag.
> 
> In this commit it only updates CMake so that it checks for the 3.1.0 installed version. The reason being that softwarecontainer was updated to work with 3.1.0 or higher versions.

Would be nice if it was explained in the commit message.

martin-ejdestig:
> In this commit it only updates CMake

The submodule is updated as well... ? I meant to ask if this was by accident or not since it is not mentioned in commit message. Did not though... but this makes it sound like it was by accident...?

martin-ejdestig:
If the submodule was not update by accident, it would have been nice with a blurb about it in the commit message. :)

TSAnsari:
> > > The version is mentioned here as well. https://github.com/Pelagicore/vagrant-cookbook/blob/master/deps/softwarecontainer-dependencies.sh, does this need to point to master now or is the hard coded version ok?
> > 
> > 
> > In the above cookbook link it is fetching and installing from lxc 3.1.0 branch/tag to be used by softwarecontainer. It should point to 3.1.0 tag.
> > In this commit it only updates CMake so that it checks for the 3.1.0 installed version. The reason being that softwarecontainer was updated to work with 3.1.0 or higher versions.
> 
> Would be nice if it was explained in the commit message.

Sure will add more comment

TSAnsari:
> > In this commit it only updates CMake
> 
> The submodule is updated as well... ? I meant to ask if this was by accident or not since it is not mentioned in commit message. Did not though... but this makes it sound like it was by accident...?

Not an accident :) Will add more comment

TSAnsari:
Updated the commit message

